PR #322, opened 2013-05-10
add -d flag for clone to owner/name directory

AndersDJohnson:
Adds a -d flag to the clone command for automatically cloning to a nested directory "owner/name" matching the GitHub username and project name.

e.g.

``` sh
$ hub clone -d defunkt/hub
```

which clones to "./defunkt/hub" instead of the default "./hub".

I use this pattern myself anyway for all my GitHub clones; I wanted to make it convenient.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,12 @@ superpowers:
     $ git clone resque
     > git clone git@github.com/YOUR_USER/resque.git
 
+    $ git clone -d schacon/ticgit
+    > git clone git://github.com/schacon/ticgit.git schacon/ticgit
+
+    $ git clone -d resque
+    > git clone git@github.com:YOUR_USER/resque.git YOUR_USER/resque
+
 ### git remote add
 
     $ git remote add rtomayko


dhirschfeld:
I don't use `hub` but I came here to see if `hub` provided this exact functionality. If it did, the convenience alone would be enough for me to consider using it. I think it's a pretty popular way for users to organise repos.

As I see it the point of using `hub` is convenience and this feature would help in that regard for a number of users. Since it's a very minimal change which is opt-in I think consideration should be given to reopening this.

mislav:
@dhirschfeld I don't think this would see wide usage, but you can easily make a git alias for it:

```console
$ git config --global alias.cl '!bash -c "echo hub clone \\$1 \\$1"'
$ git cl owner/repo
```

dhirschfeld:
Sure, but the same could said for most of hubs functionality no?

For me it's easier to deploy a tool that has this functionality builtin than it is to pre-configure users git installs. I can certainly do that... but then there's not much point in introducing another dependency - just pre-configure a bunch of git aliases and call it a day

PR #350, opened 2013-07-09
Add correct hub alias for fish shell (fixes #248)

### LINKED ISSUE #248: hub alias for fish shell is wrong. twice.
For fish shell, the output of `hub alias` and `hub alias -s` are both wrong.
The output for `hub alias fish` must be
`eval (hub alias -s)`

And the output for `hub alias -s fish` must be
`alias git hub`
(I know the doc of alias say it support the current output, it doesn't work)

jm3:
Added proper alias and path to fish dotfile `~/.config/fish/config.fish`. Commit adds two new tests; all scenarios + steps pass. @dideler + @KangOl, please let me know if this works for you. :fish:!

### DIFF
--- a/features/alias.feature
+++ b/features/alias.feature
@@ -10,6 +10,16 @@ Feature: hub alias
       eval "$(hub alias -s)"\n
       """
 
+  Scenario: fish instructions
+    Given $SHELL is "/usr/local/bin/fish"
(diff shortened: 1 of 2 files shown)


corysimmons:
Care to update README with this?

PR #442, opened 2013-12-04
[WIP] commands: make fork() set remote.pushdefault

artagnon:
Hi,

Unfortunately, my patch doesn't work as expected, and I can't figure out why. Any clues?

Thanks.

### DIFF
--- a/features/fork.feature
+++ b/features/fork.feature
@@ -14,6 +14,7 @@ Feature: hub fork
     When I successfully run `hub fork`
     Then the output should contain exactly "new remote: mislav\n"
     And "git remote add -f mislav git@github.com:mislav/dotfiles.git" should be run
+    And "git config remote.pushdefault mislav" should be run
     And the url for "mislav" should be "git@github.com:mislav/dotfiles.git"
 
   Scenario: --no-remote
--- a/lib/hub/commands.rb
+++ b/lib/hub/commands.rb
@@ -545,6 +545,7 @@ module Hub
       else
         url = forked_project.git_url(:private => true, :https => https_protocol?)
         args.replace %W"remote add -f #{forked_project.owner} #{url}"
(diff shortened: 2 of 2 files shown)


sambrightman:
I was about to submit a PR for exactly this feature. Why would you not want to do this? Automatic tasks that are annoying but trivial seems exactly what hub should do to me. Is there some sort of multi-remote setup where you would fork a GitHub repo and not want to push to your fork by default? It seems unlikely that you have write access to origin if you do this. It also seems unlikely that you can cause much damage by pushing to your fork.

mislav:
@sambrightman The configuration setting looks innocent in theory, but in the long run the existence of the setting may cause headaches for some in a multi-git remote scenario when they don't understand that there was a `pushdefault` added for them. I'm trying to avoid hub doing anything potentially surprising.

Yes I agree that it makes the most sense that you would probably want to only push to your fork. I just don't find that it's too difficult to say `git push -u mislav HEAD` when publishing a branch to my fork. After that, for the same branch I can use `git pull/push` without arguments.

sambrightman:
It still seems very unlikely to me that someone runs a complicated multi-remote setup, uses hub with it, and uses implicit push. If you run a complicated multi-remote setup, surely you use and explicit remote and this feature would not affect you. If you did accidentally run implicitly then pushing to your personal fork seems safer than pushing to random remote R.

How about setting this only if there is a single remote called `origin` setup, `remote.pushdefault` is unset, and all branches have `branch.<name>.remote` set to `origin` when `hub fork` is run? Something like that represents the canonical GitHub workflow.

mislav:
Hub will not set any configuration options for you that you didn't ask for explicitly.

`hub fork` is just an alternative to forking a repo on GitHub web interface. It won't do anything other than that.

PR #481, opened 2014-02-10
gh: release command improvements

### LINKED ISSUE #477: gh: Improve creating of Releases
Right now, it's not possible to `hub release create` without the assets flag or a `releases/TAG` directory.

For projects that just want to publish release notes, it would be great if we lifted this restriction.

Also, I feel that the `releases/TAG` default behavior is too arbitrary. We don't know how people will be packaging their software, so I don't think it's useful to have any kind of hardcoded default. I'd rather have separate flags to allow explicit upload of a single file vs. contents of an entire directory.

Note that GitHub web UI doesn't use the term "assets", but refers to them as "binares" which you can "attach". So if we want to be explicit:

```
hub release create --attach pkg/my-v1.0.zip
hub release create --attach-dir pkg/v1.0/
```

/cc @jingweno

calavera:
Fixes #477

I've removed the defaults for the release command and now it will only upload assets explicitly. It detects whether you want to upload a file of every file inside a directory using the flag --attach or -a.

This change made all the tests that we had for this command obsolete. I'm trying to find a way we can add some coverage to the new changes.

### DIFF
--- a/commands/release.go
+++ b/commands/release.go
@@ -30,8 +30,7 @@ var (
 		Long: `Creates a new release in GitHub for the project that the "origin" remote points to.
 It requires the name of the tag to release as a first argument.
 
-Specify the assets to include in the release from a directory via "-a". Without
-"-a", it finds assets from "releases/TAG" of the current directory.
+Specify the assets to include in the release from a directory via "-a".
 
 Without <MESSAGE> or <FILE>, a text editor will open in which title and body
 of the release can be entered in the same manner as git commit message.
@@ -44,7 +43,7 @@ If "-p" is given, it creates a pre-release.
 	flagReleaseDraft,
 	flagReleasePrerelease bool
 
-	flagReleaseAssetsDir,
+	flagReleaseAssets,
 	flagReleaseMessage,
 	flagReleaseFile string
 )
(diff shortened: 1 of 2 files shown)


KHOENIX:
> Great, thanks for working on this. Two things:
> 
> 1. What do you think about dropping the shorthand flag `-a` and just supporting `--attach`? I only support shorthand flags for often-typed commands, but I don't think the `release create` command will be typed often. I see it as something that will embedded as a part of a larger release script (since you need to build binaries first anyway).
> 2. Our test suite on the "gh" branch is broken anyway (mostly Cucumbers) so you didn't need to comment out the tests. It's a bit unfortunate that a smaller change breaks all existing unit tests, though.

KHOENIX (commented):
Leave A comment

KHOENIX (commented):
Leave of comment

KHOENIX:
> Fixes #477
> 
> I've removed the defaults for the release command and now it will only upload assets explicitly. It detects whether you want to upload a file of every file inside a directory using the flag --attach or -a.
> 
> This change made all the tests that we had for this command obsolete. I'm trying to find a way we can add some coverage to the new changes.

KHOENIX (commented):
Leave of comment

PR #503, opened 2014-02-28
Enable API debug output by setting HUB_VERBOSE

mislav:
This will print information about every HTTP request and response to stderr. Authentication info is redacted.

```
$ HUB_VERBOSE=1 hub ci-status
> GET https://api.github.com/...
> ...
```

This option for debugging is crucial for some of our users that submit reports about API failures, but their console output is not enough to debug their issue.

@jingweno: Would it be possible to get the same feature in Octokit?

### DIFF
--- a/features/create.feature
+++ b/features/create.feature
@@ -141,15 +141,19 @@ Feature: hub create
       """
     And $HUB_VERBOSE is "on"
     When I successfully run `hub create`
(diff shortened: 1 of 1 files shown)


ElijahLynn:
For Fish users `set -x HUB_VERBOSE 1`.

PR #692, opened 2014-11-19
Initial support for checkout a tree url in addition to pull-request urls.

maxandersen:
Getting used to use git co <pullrequest-url> and then see it fail when using a tree url (ttps://github.com/jbosstools/jbosstools-forge/tree/jbosstools-4.2.x) 
always annoyed me.

Here is the basic initial implementation of this.

I would like to write tests for it too, but the current one fails so I'm not sure how to
add tests for it knowing if I broke something. 

Any pointers would be appreciated on what to do about that.

### DIFF
--- a/commands/checkout.go
+++ b/commands/checkout.go
@@ -11,11 +11,11 @@ import (
 var cmdCheckout = &Command{
 	Run:          checkout,
 	GitExtension: true,
-	Usage:        "checkout PULLREQ-URL [BRANCH]",
+	Usage:        "checkout PULLREQ-URL|BRANCH-URL [BRANCH]",
 	Short:        "Switch the active branch to another branch",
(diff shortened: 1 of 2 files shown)


mislav:
@maxandersen Do you still want this feature? We've been waiting for an update for a long time :wink: 

I don't think this feature would get a lot of use. If you don't need it, I suggest you close the PR at your convenience. Thanks!

maxandersen:
I still use it but I couldn't get the tests to reliably pass/fail even before adding my tests so I didn't do more yet ;/

the feature is very useful when PR's aren't opened yet. I'll try look at again but i'll close this and reopen when I have something to not "pollute" the queue.

mislav:
Thanks for the update!

PR #748, opened 2014-12-25
Improve package script

owenthereal:
- Friendly OS name for packaged zips
- Options to skip test and toolchain building
- Fix problems running on Windows

Unfortunately we can't build 64bit and 32 bit Windows binary on a 64bit machine. Seeing issues like [this](https://github.com/mitchellh/gox/issues/19#issuecomment-68117016). A solution is to build 64bit hub on a 64bit machine and 32bit hub on a 32bit machine. It's annoying but get us unblock for now.

### DIFF
--- a/script/package
+++ b/script/package
@@ -20,8 +20,24 @@ module OS
       end
     end
 
+    def friendly_name
+      if darwin?
+        "mac"
+      elsif linux?
+        "linux"
+      elsif windows?
+        "windows"
+      else
+        raise "Unknown OS type #{RUBY_PLATFORM}"
+      end
+    end
+
     def windows?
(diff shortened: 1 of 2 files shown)


mislav:
Nice! How do the conventions by which we name binaries affect the auto-updater functionality in the long run?

jingweno:
@mislav The name changes affect the name of the directory containing the binary and the tar file name while the binary name is still `hub`. However, auto-update is currently [broken due to the tar file name changes](https://github.com/github/hub/blob/master/commands/updater.go#L120). I'll mark it down in [Trello](https://trello.com/c/AdNx7oxT/13-auto-update) as TODO when we want to turn this feature on.

PR #751, opened 2015-01-02
Skip completion scripts for Windows

jingweno:


mislav:
This is for the packaging task right? There's no harm in packaging completion script in the archive. People with win-bash might even figure out how to make use of them.

jingweno:
>  People with win-bash might even figure out how to make use of them.

Alright, I'm leaving them alone them.

PR #760, opened 2015-01-07
Add flag to sign pull-requests.

calavera:
Pull requests are already signed if you use `commit -s` and there is only one
commit when you create them. This gives authors the possibility of
signing pull requests explicitly without signing every commit.

TODO:

- [x] Add config var to always sign pull requests by default
- [x] Add tests/specs
- [x] Add documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,9 @@ superpowers:
     # explicit title, pull base & head:
     $ git pull-request -m "Implemented feature X" -b defunkt:master -h mislav:feature
 
+    # explicit pull-request sign off:
+    $ git pull-request -s
+
 ### git checkout
 
     $ git checkout https://github.com/defunkt/hub/pull/73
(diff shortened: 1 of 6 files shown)


jingweno:
:+1: to this feature!

calavera:
@jingweno, @mislav I've added tests and documentation. I'd really appreciate your feedback on this.

jingweno:
@calavera It looks like the sign-off person is the committer. After a second thought, what would be a use case to sign off my own pull request? I've seen workflow that repo maintainers sign off pull requests from contributors. Maybe the sign-off feature should be in `git-merge`?

mislav:
Sorry for my late reply. Do we still want this feauture? Isn't a pull request implicitly signed-off by you by the act of opening it in the first place?

mislav:
Thanks for the PR, but I'm going to axe this feature idea. We're having trouble keeping up with hub feature requests and maintenance anyway, and this wasn't an often requested feature. A more often requested feature is adding PR templates, which we want to do in the future, and which would also allow adding a signed-off line if there's a need for explicit signoff.

Closing in favor of #937 idea.

PR #765, opened 2015-01-18
Clarify setup when already developing on Go

soniah:
It is inclear from the setup instructions whether hub
requires special compilation (like Terraform), or is just
another Golang project.

This tweak to the readme adds the standard ```go get``` to
the install instructions.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -60,7 +60,7 @@
 		},
 		{
 			"ImportPath": "github.com/mattn/go-isatty",
-			"Rev": "fdbe02a1b44e75977b2690062b83cf507d70c013"
+			"Rev": "6152ce208cfa13d58f065348a3312b4160fb98d1"
 		},
 		{
 			"ImportPath": "github.com/octokit/go-octokit/octokit",
--- a/Godeps/_workspace/src/github.com/mattn/go-isatty/isatty_bsd.go
+++ b/Godeps/_workspace/src/github.com/mattn/go-isatty/isatty_bsd.go
@@ -1,4 +1,4 @@
-// +build darwin freebsd
+// +build darwin freebsd openbsd
 
 package isatty
 


mislav:
Hi, thanks for the recommendation. hub is a pretty standard Go project. The `go get` command could be a good way for people to obtain hub if they're already set up with Go.

README.md
```diff
@@ -54,6 +54,11 @@ $ ./script/build
 $ cp hub YOUR_BIN_PATH
 ~~~
 
+# Or, if your go development bin directory is already in your $PATH:
```
mislav: What do you mean by "go development bin directory" exactly?

soniah: The directory where go puts it's binaries. For example, on my machines:
```
% echo $GOPATH
/Users/sonia/go
```
So compiled binaries (eg golint, go-bindata) get put in ```/Users/sonia/go/bin```. Which is also in my $PATH:
```
% echo $PATH
/usr/local/bin:/Users/sonia/bin:/Users/sonia/bin/convert:/Users/sonia/go/bin:
                                                                                               ^^^^^^^^^^^^^^^
/usr/local/go/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/usr/texbin:/usr/local/ec2/ec2-api-tools-1.7.3.0/bin
```

mislav: Ah, gotcha; so it's `$GOPATH/bin`. Maybe we should say that here to avoid ambiguity?

soniah: Ah yes that makes sense. Sorry to have caused so much work! :-)

mislav: Could you then reword this line to say something like

> Or, if you've done Go development before and your `$GOPATH/bin` directory is already in your PATH:

soniah: OK. I've squashed those 2 commits and repushed. Now ready for you to pull.

mislav:
Thanks. Two things we need to solve right now:

1. The sentence starts with `#` character which Markdown turns into an `<h1>` element which is what we don't want. We should remove that character, and maybe add a blank line between that sentence and the next code block (for readability).
2. The rebase seems to have introduced duplicate commits into this PR. For example, commit f1f50c59fe5ec1b7ff0dc22ddc0a1c9249dc384a that's part of the master branch in hub is duplicated in your PR as https://github.com/soniah/hub/commit/581e841506251c489b7587de6fb210b6438df73f.

  You should `git fetch origin` again, assuming that "origin" points to `github/hub`, and `git rebase -i origin/master` and then remove the commits except your own. Then force push to this branch again. Thanks!

PR #767, opened 2015-01-18
Add brew cask installation instructions

pcasaretto:
I have [just added](https://github.com/caskroom/homebrew-cask/commit/354168b4b17f1f67d5ecc408329f95f328ab6b2d
) hub to homebrew cask.
I thought it would be worth mentioning it in the README :) .

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,12 @@ $ brew install --HEAD hub
 [compiled binaries](https://github.com/github/hub/releases) and put it anywhere
 in your executable path.
 
+Precompiled binaries are also available on [Homebrew cask](https://github.com/caskroom/homebrew-cask).
+
+~~~ sh
+$ brew cask install hub
+~~~
+
 #### Source
 
 To install `hub` 2.x from source, you need to have a [Go development environment](http://golang.org/doc/install),


hkdobrev:
@pcasaretto It is already available in Homebrew: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/hub.rb

Why would you add it to Cask?

pcasaretto:
For the simplicity of it. It's even easier (assuming you already use cask) than downloading the binaries.

callumacrae:
@pcasaretto How is `brew cask install hub` simpler than `brew install hub`?

pcasaretto:
Given you already use cask, there is no need to install Go (or compile anything).
It is, therefore, simpler.

mislav:
@pcasaretto I get how it's simpler and faster install process that right now `brew install hub --HEAD` is. However, when we release 2.2 final (just around the corner!), the main Homebrew task will install the precompiled binary. Then, the cask won't be needed, as Go won't be required unless you use `--HEAD`.

jingweno:
I'm not a cask user. Is it similar to [Homebrew Bottles](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Bottles.md)?

pcasaretto:
@mislav that's awesome. I was under the impression that homebrew did not like precompiled binaries.
I guess we can go ahead and close this out.

Issue #769, opened 2015-01-19
Fork command on GitHub Enterprise prompts with incorrect hostname


mislav:
While in a repo hosted on `ghe.io` (GitHub Enterprise instance):

```
$ git fork
github.com username: mislav
ghe.io password for mislav (never stored):
```

Although the 2nd prompt was correct, the first prompts for `github.com` username which is definitely wrong since no git remotes in my local repo point to to `github.com`.

jingweno:
Good find. Fixed in https://github.com/github/hub/pull/770

PR #770, opened 2015-01-20
Fix prompt for user host for GHE

jingweno:
Reference: https://github.com/github/hub/issues/769


mislav:
Nice and quick thanks :)

PR #774, opened 2015-01-21
Fix package's file extensions

nbraud:
Tarballs are traditionally .tar.gz, not .gz.tar
This can be seen in the tarballs for 2.2.0-rc1 provided on https://github.com/github/hub/releases


jingweno:
:+1: Good catch! Thanks!

cebe:
was about to report the same :) :+1:

PR #776, opened 2015-01-21
Fix `hub checkout <PULLREQ-URL>` for renamed fork

### LINKED ISSUE #768: The "hub co" for pull request on renamed repo fails
There 2 repos:
* https://github.com/Behat/docs
* https://github.com/minkphp/docs

If you try to fork both of them into same account you run into naming conflict and are forced to fork either of repos under different name.

For example try:

1. clone https://github.com/minkphp/docs repo
2. do `hub co https://github.com/minkphp/docs/pull/45`

It fails with this error:
```
fatal: Couldn't find remote ref refs/heads/session
error: Could not fetch stof
```
This is because attempt is made to to read `docs/session` ref from `Behat/docs` fork.

I think that we shouldn't assume that every forked repo is named the same and use GitHub API for detecting actual fork name for every user.

mislav:
Start reading the `Head.Repo.Name` property from API payload to determine the name of the forked repo which might not match the name of the base repo.

Fixes #768

### DIFF
--- a/commands/checkout.go
+++ b/commands/checkout.go
@@ -93,7 +93,7 @@ func transformCheckoutArgs(args *Args) error {
 		remoteURL := fmt.Sprintf("+refs/heads/%s:refs/remotes/%s/%s", branch, user, branch)
 		args.Before("git", "fetch", user, remoteURL)
 	} else {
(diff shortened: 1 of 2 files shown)


jingweno:
:+1:

jingweno:
You beat me to it :)

PR #778, opened 2015-01-22
don't exit when "git rev-parse" fails

Mischi:
so that the version number is always echoed on stdout.

With this change, the version script will print a version
number even if git or the ".git" directory are not available
like in the "release-source.tar.gz" archive.


jingweno:
LGTM :+1:

PR #779, opened 2015-01-22
update mattn/go-isatty dep to support OpenBSD

Mischi:
With this change i'm able to run hub on OpenBSD.

Regards,
Fabian


jingweno:
:+1:

jingweno:
Updated dependency PR: https://github.com/mattn/go-isatty/pull/1

PR #784, opened 2015-01-27
Support clone from owner/repo for Enterprise

msabramo:
Without this change, it seems to try to clone from github.com and it fails.

    $ ./hub clone my_org/my_repo
    Cloning into 'my_repo'...
    fatal: remote error:
      Repository not found.

With this, I get prompted to pick the host:

    $ ./hub clone my_org/my_repo
    Select host:
     1. github.com
     2. github-enterprise.mycompany.com
    > 2
    Cloning into 'my_repo'...
    remote: Counting objects: 9892, done.
    remote: Compressing objects: 100% (2838/2838), done.
    remote: Total 9892 (delta 7130), reused 9729 (delta 7004)
    Receiving objects: 100% (9892/9892), 3.45 MiB | 2.16 MiB/s, done.
    Resolving deltas: 100% (7130/7130), done.

### DIFF
--- a/commands/clone.go
+++ b/commands/clone.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"os"
 	"regexp"
 	"strings"
 
@@ -58,7 +59,7 @@ func transformCloneArgs(args *Args) {
 			if nameWithOwnerRegexp.MatchString(a) && !isDir(a) {
 				name, owner := parseCloneNameAndOwner(a)
 				var host *github.Host
-				if owner == "" {
+				if owner == "" || os.Getenv("HUB_CHOOSE_HOST") != "" {
 					config := github.CurrentConfig()
 					h, err := config.DefaultHost()
 					if err != nil {


msabramo:
```
$ go test ./...
?   	github.com/github/hub	[no test files]
ok  	github.com/github/hub/cmd	0.005s
ok  	github.com/github/hub/commands	0.347s
?   	github.com/github/hub/fixtures	[no test files]
ok  	github.com/github/hub/git	0.315s
ok  	github.com/github/hub/github	0.332s
ok  	github.com/github/hub/utils	0.005s
```

jingweno:
Your changes have broken [some cukes](https://travis-ci.org/github/hub/builds/48532683). The failures are prob. due to the prompt that you added.

msabramo:
Updated so that it only does the prompt when `HUB_CHOOSE_HOST` is set. Easiest way to not break existing behavior and thus tests.

jingweno:
My bad I didn't give more info on what would be a good user experience. I actually think prompting the user to select host for `hub-clone` is fine. We'd just need to look into why the other cukes are failing. Let's see what @mislav think.

msabramo:
I kinda wanted a `--host` option so I could explicitly specify the host I want to use, but I didn't want to go that route until there was some discussion like this.

msabramo:
AFAICT, the submodule cukes are failing, because the submodule code also uses this function for clone (somewhat unexpected to me) and they are relying on the old behavior of assuming you always mean github.com.

mislav:
I've just discovered the prompt recently, not knowing before that that this is a feature of Go hub. It was an interesting touch, and worked well for my use-case, but ultimately I think we should refrain from doing prompts. Git itself never requires you to go through an interactive step unless you ask it to (e.g. `git rebase -i`).

This is the relevant section for this issue in the hub man page:

```
The default host for commands like init and clone is still github.com, but
this can be affected with the GITHUB_HOST environment variable:

    $ GITHUB_HOST=my.git.org git clone myproject
```

Yeah, setting an environment variable is awkward, but I don't know how to best handle the case where we don't know which of your GitHub instances do you want to communicate with. Suggestions welcome, as long as they follow the unix philosophy of building well-behaved command-line tools.

mislav:
@msabramo I think the `--host` option to hub would be the way to go. I'll see if I can find the time to work on this so it's supported globally on every hub command where that makes sense. In the meantime, I'll close this in the favor of the GITHUB_HOST environment variable that we already support.

jasonkarns:
I just wanted to throw my vote in favor of restoring the prompt behavior. I am rarely (if ever) in situation where I'm interacting with only a single github host, so setting an environment variable doesn't really meet my needs at all.

To be quite honest, without the prompt behavior (which I was apparently relying on for a quite a while unknowingly until I updated my hub client), setting the host via env var doesn't really help. In order to use one's non-default host would require typing the full hostname as part of the command anyway. So it rather defeats the point of using hub as it is then faster to just type the full repo url.

```
GITHUB_HOST=github.example.com hub clone foo/bar
git clone https://github.example.com/foo/bar
```

I understand the desire to avoid any interaction by default. Perhaps the interactive behavior could be opt-in with a config setting? (Like the `HUB_CHOOSE_HOST` env var but set via gitconfig?)

mislav:
How about on `hub clone foo/bar`, hub iterates over the list of known
GitHub hosts and downloads the first matching repo?

jasonkarns:
I could see that. I would hope github.com is last in that list, though. I think manually added hosts should take priority. (And should clearly state in the output which host was used.)

Alternatively, check for the user/repo on all known hosts and only clone if it exists on just a single host. That way there is no ambiguity around where the clone came from and it forces one to be explicit if there are multiple matching repos.

nilbus:
`hub clone foo/bar` iterating over hosts, github.com last, is exactly the intuitive behavior I hoped hub might implement when I tried to clone an enterprise repo the first time.

jasonkarns:
Since this has some added activity, I thought I'd just share how I've been working around this lately. I've set up aliases for `hub` (like `hubs` and `huba`) that prepend the `GITHUB_HOST` env var accordingly. It's not as clean as a this feature proposal, but at least gets back to short CLI commands.

PR #785, opened 2015-01-28
Refine package scripts

jingweno:
This is to improve package script in two ways:

1 Add man pages to build package
2 Rename package name as `hub_OS_CPU_vVERSION`. Ending package name with `vVERSION` allows Homebrew to recognize the build version in `brew install`. See https://github.com/boxen/puppet-hub/pull/10#issuecomment-71783691


mislav:
Nice, thanks!

PR #791, opened 2015-02-03
Support `hub init -g DIR`

### LINKED ISSUE #790: Does not support syntaxes like `hub init -g DIRECTORY`
Git itself supports syntax like `git init DIRECTORY`, which initializes the repo in the specified directory (possibly after creating it).
But as for hub, `hub init -g` always takes cwd as the path of newly initialized repo. So, in cases like `hub init -g DIRECTORY`, the hub command, which is supposed to function in the specified directory, still stays in the current working directory, and is likely to raise an error like this:
```fatal: Not a git repository (or any of the parent directories): .git```

### PR DESCRIPTION
**Pull request by jingweno**
This fixes #790.

### DIFF
--- a/commands/init.go
+++ b/commands/init.go
@@ -1,6 +1,8 @@
 package commands
 
 import (
+	"path/filepath"
+
 	"github.com/github/hub/github"
 	"github.com/github/hub/utils"
 )
@@ -37,14 +39,32 @@ func transformInitArgs(args *Args) error {
 		return nil
 	}
 
-	name, err := utils.DirName()
-	if err != nil {
-		return err
+	var (
+		name   string
+		newDir bool
+		err    error
+	)
+
+	if args.IsParamsEmpty() {
+		name, err = utils.DirName()
+		if err != nil {
+			return err
+		}
+	} else {
+		name = args.LastParam()
+		newDir = true
 	}
 
 	project := github.NewProject("", name, "")
 	url := project.GitURL("", "", true)
-	args.After("git", "remote", "add", "origin", url)
+
+	cmds := []string{"git"}
+	if newDir {
+		cmds = append(cmds, "--git-dir", filepath.Join(name, ".git"))
+	}
+
+	cmds = append(cmds, "remote", "add", "origin", url)
+	args.After(cmds...)
 
 	return nil
 }
--- a/features/init.feature
+++ b/features/init.feature
@@ -8,6 +8,11 @@ Feature: hub init
     When I successfully run `hub init -g`
     Then the url for "origin" should be "git@github.com:mislav/dotfiles.git"
 
+  Scenario: Initializes a git repo in a new directory with remote
+    When I successfully run `hub init -g new_dir`
+    And I cd to "new_dir"
+    Then the url for "origin" should be "git@github.com:mislav/new_dir.git"
+
   Scenario: Enterprise host
     Given $GITHUB_HOST is "git.my.org"
     And I am "mislav" on git.my.org with OAuth token "FITOKEN"

### CODE REVIEW

**Comment by jingweno** [comment]
Shamelessly self merged. I'm confident enough this should fix it.

/cc @mislav

File: commands/init.go [hunk 1]
```diff
@@ -37,14 +39,32 @@ func transformInitArgs(args *Args) error {
...
+		err    error
+	)
+
+	if args.IsParamsEmpty() {
+		name, err = utils.DirName()
+		if err != nil {
+			return err
+		}
+	} else {
+		name = args.LastParam()
```
mislav: Could `LastParam` possibly be an unrelated flag or value for `--template` or similar?

commands/init.go
```diff
@@ -37,14 +39,32 @@ func transformInitArgs(args *Args) error {
...
+		newDir = true
 	}
 
 	project := github.NewProject("", name, "")
 	url := project.GitURL("", "", true)
-	args.After("git", "remote", "add", "origin", url)
+
+	cmds := []string{"git"}
+	if newDir {
+		cmds = append(cmds, "--git-dir", filepath.Join(name, ".git"))
```
mislav: This fails to respect `--separate-git-dir` which might be optionally given

features/init.feature
```diff
@@ -8,6 +8,11 @@ Feature: hub init
     When I successfully run `hub init -g`
     Then the url for "origin" should be "git@github.com:mislav/dotfiles.git"
 
+  Scenario: Initializes a git repo in a new directory with remote
+    When I successfully run `hub init -g new_dir`
+    And I cd to "new_dir"
+    Then the url for "origin" should be "git@github.com:mislav/new_dir.git"
```
mislav: This step would pass even if the remote configuration erroneously happened in the starting directory before you cd'd into `new_dir`. So before changing the directory, maybe we should assert that no `.git` directory was created in the current location as the result of `hub init` operation (exactly the bug that we're fixing).

commands/init.go
jingweno: It's possible. My bad, will fix it in another PR

commands/init.go
jingweno: `--separate-git-dir` is a flag for `git-init`. But the line is to run `git --git-dir DIR remote add origin URL` for `git-remote`

features/init.feature
mislav: I wouldn't pass because of the directory name, correct. But, we're not sufficiently asserting that the `.git/` directory is created in the exact location. I'm a little nitpicky, I know.

commands/init.go
mislav: If `--separate-git-dir` is given, the path to the git directory will be different than inferring it as `filepath.Join(name, ".git")`.

jingweno: `git init --separate-git-dir` creates a filesytem-agnostic symlink to current folder or a folder passed in. For both cases, we need to run `git-remote` in the current folder or the folder passed in, instead of the folder as part of the `--separate-git-dir` flag which contains the bare repo:

```
$ git init foo --separate-git-dir=../bar
$ cd foo
$ cat .git
gitdir: /Users/owenou/src/bar
```

features/init.feature
jingweno: > I'm a little nitpicky, I know.

Why wouldn't @mislav let go :smiley_cat: ? I added the assertion in https://github.com/github/hub/pull/793.

commands/init.go
jingweno: Fixed in https://github.com/github/hub/pull/793

PR #793, opened 2015-02-05
Properly filter `git-init` optional folder

jingweno:
This fixes https://github.com/github/hub/pull/791#discussion_r24013319 and https://github.com/github/hub/pull/791#discussion_r24039843.

### DIFF
--- a/commands/init.go
+++ b/commands/init.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"path/filepath"
 
+	flag "github.com/github/hub/Godeps/_workspace/src/github.com/ogier/pflag"
 	"github.com/github/hub/github"
 	"github.com/github/hub/utils"
 )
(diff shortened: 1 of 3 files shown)


features/steps.rb
```diff
@@ -257,3 +257,9 @@
     File.open(file, "w") { |f| f.write(contents) }
   end
 end
+
+Then(/^the folder "(.*?)" doesn't exist$/) do |folder|
```
mislav: Aruba provides us with this step implementation:

```rb
Then /^a directory named "([^"]*)" should not exist$/ do |directory|
  check_directory_presence([directory], false)
end
```

mislav:
We should add a test for `--separate-git-dir` as well. What it will do is create a file (not directory) named `.git` in the current directory that as its contents has the path to the actual git repo on disk. I think our current implementation should work with it, but just in case.

commands/init.go
```diff
@@ -45,13 +45,14 @@ func transformInitArgs(args *Args) error {
 		err    error
 	)
 
-	if args.IsParamsEmpty() {
+	words := args.Words()
```
mislav: How can we distinguish between a "word" that stands on its own and words the are values for options such as `--template`, `--separate-git-dir`, `--shared`?

jingweno: [`args.Words()'](https://github.com/github/hub/blob/master/commands/args.go#L23-L32) returns args not starting with "-"

features/steps.rb
jingweno: :+1: Fixed with `--amend`

commands/init.go
mislav: Yes, in the invocation `git init --template foo`, `foo` is a "word" since it doesn't start with "-" but is not the directory argument.

jingweno: I missed this case. I've fixed it by parsing all flags and arguments for `git-init` to get the optional directory argument.

jingweno:
@mislav I added a cuke to test `--separate-git-dir` as well. :shipit:?

mislav:
Cuke looks alright, but I didn't like the fact that we're pulling in a commandline flag processing dependency in here, as well as hardcoding all of `git init` core flags. It would be enough to just hardcode the 3 parameters that can hold values (`template`, `shared`, `separate-git-dir`) and iterate through arguments with a simple loop. I opened another PR with an implementation of what I have in mind. I also wrote more tests in Go instead of cukes since they're faster.

PR #796, opened 2015-02-07
merge: do not insert "--no-ff" for "--squash"

blueyed:
Otherwise `git` aborts with an error:

> You cannot combine --squash with --no-ff

There should probably get a test added for it, but I wasn't sure what it should look like.. like the `--ff-only` one, but adjusted to different output, or just test that it won't fail?


mislav:
Fix looks great; thanks! How about adding tests? Either Go tests or an extra cuke would be fine.

blueyed:
> How about adding tests?

Please feel free to do so.. :)

I would have copied / adopted the `--ff-only` one (Go test IIRC), but then the suite did not run properly for me (IIRC it was about the git editor setting), and I couldn't say if the expected output would have been correct.

jingweno:
@blueyed Thanks for the fix. I accommodated your commit and added some tests in https://github.com/github/hub/pull/807. Closing.

Issue #762, opened 2015-01-09
add option to `hub fork` to set the name of the remote that is created


ryan-williams:
it creates a remote named `$GITHUB_USER` but I frequently want to use something shorter ("ryan" vs. "ryan-williams")

jingweno:
Duplicated of https://github.com/github/hub/issues/754

ryan-williams:
It seems like #697 wants the fork on github to have a different name; I just want the created remote (that points to the fork) in my local git repo to have a name other than my `$GITHUB_USER`.

The former seems a good deal trickier than the latter, which is what I was asking for here, so I'm not sure I agree that this is a duplicate of that, fwiw.

ryan-williams:
Let me know if I'm misreading or misunderstanding.

jingweno:
Sorry, maybe not completely duplicate. But could be solved if https://github.com/github/hub/issues/697 is implemented - just provide your own fork origin name.

ryan-williams:
maybe, but actually the name of the fork and the name of my local git-remote that points at it are two distinct things.

I personally always like my forks to have the same name as the upstream, so I have no interest in #697 being solved (if I'm correctly understanding what's being asked for there).

I just don't like to use my github username to name my git-remotes that point at my forks.

ryan-williams:
sorry to be pedantic, but it seems like this could also be solved much more easily than #697, and I don't think their respective solutions would have much to do with one another

ryan-williams:
(having not actually looked at the implementation of `hub fork`, just surmising from these issues and the way I would guess that it all works)

jingweno:
> maybe, but actually the name of the fork and the name of my local git-remote that points at it are two distinct things.

The remote name of your fork repository is what I meant. I'm thinking something like this:

```
$ git clone github/hub hub-custom
$ cd hub-custom
$ hub fork ryan-williams
```

What do you think?

jingweno:
Ah...sorry, I misunderstood what https://github.com/github/hub/issues/697 does. I thought it only means the remote name...reopening

jingweno:
https://github.com/github/hub/issues/754 is a closer issue

ryan-williams:
[...]

Issue #754, opened 2015-01-04
FR: customizable remote names when hub fork a repo


yegle:
`hub fork` works a little bit counter-intuitive to me.

The following is the manual way to fork a repo before I know this great tool `hub`

1. Go to https://github.com/github/hub, fork to my github
2. On my workstation, `git clone git@github.com:yegle/hub`
3. Optionally I'll create an `upstream` referring to `https://github.com/github/hub`

But `hub fork` will use `origin` to refer to the ORIGINAL repo while manual forking will have `origin` refer to your forked repo. I'd suggest use `upstream` or any customizable name to refer to the ORIGINAL repo, while still use `origin` to refer to your forked repo.

Possibly related issue: #697 and #720

brock:
:+1: 
Similar to the discussion here https://github.com/isaacs/github/issues/280

I'd love it if `hub fork` cloned my fork, set my forks remote as "origin" and set the original project as "upstream" as is outlined in the Github documentation here: https://help.github.com/articles/configuring-a-remote-for-a-fork/

kbrock:
:+1: this matches my workflow to a T.

mislav:
Closing in favor of the longer discussion in #762. We'll try to make this configurable in the future.

Issue #348, opened 2013-07-08
Fork and clone in one command


krlmlr:
Examples:

```
hub clone -f user/repo
hub clone --fork user/repo
```

> Forks and then clones, adding both the upstream and own repos as reference. I prefer my own repo to be called `origin` (since this is usually the only one I'm allowed to push to), and the remote repo to be called by the other user name or `upstream`.

```
hub fork user/repo
```

> Should work outside a repository. Perhaps easier than enhancing `clone`.

vermiculus:
@robdodson See also #720

olivierlacan:
@mislav Any new thoughts on this. Reached for `git fork org/repo` again today and was completely confused (it's been a while) that I had to do this:

```
$ git clone evanphx/benchmark-ips
Cloning into 'benchmark-ips'...
...
$ cd benchmark-ips/
$ git fork
Updating olivierlacan
From https://github.com/evanphx/benchmark-ips
 * [new branch]      clearer-compare -> olivierlacan/clearer-compare
 * [new branch]      master          -> olivierlacan/master
new remote: olivierlacan
$ git remote rename origin upstream
$ git remote rename olivierlacan origin
$ git remote -v
origin	git@github.com:olivierlacan/benchmark-ips.git (fetch)
origin	git@github.com:olivierlacan/benchmark-ips.git (push)
upstream	https://github.com/evanphx/benchmark-ips.git (fetch)
upstream	https://github.com/evanphx/benchmark-ips.git (push)
```

mislav:
@olivierlacan It makes absolute sense that `git fork org/repo` should ideally clone `org/repo` as `upstream` remote, fork it, and add an extra `origin` remote for `YOURNAME/repo`.

PRs welcome!

apjanke:
> It makes absolute sense that git fork org/repo should ideally clone org/repo as upstream remote, fork it, and add an extra origin remote for YOURNAME/repo.

What do you think about similar functionality for `hub clone` for when you're cloning a repo of yours that is an existing fork of a canonical repo?

For example, I work on several machines, and when I'm setting up a new machine, I have to clone my forked repos, and then go in and add `upstream` remotes in each of them.

How about, when doing a `hub clone <repo>` or `hub clone <user>/<repo>`, it still creates the specified repo as the `origin` remote, and then checks to see if that repo is a fork, and if so, adds an `upstream` remote pointing to the root of the fork tree?

[...]

PR #798, opened 2015-02-08
Renamed origin as upstream in git fork

RichardLitt:
This pertains to #348, #754, #762 and https://github.com/isaacs/github/issues/280

### DIFF
--- a/commands/fork.go
+++ b/commands/fork.go
@@ -13,7 +13,7 @@ var cmdFork = &Command{
 	Usage: "fork [--no-remote]",
 	Short: "Make a fork of a remote repository on GitHub and add as remote",
 	Long: `Forks the original project (referenced by "origin" remote) on GitHub and
-adds a new remote for it under your username.
+adds a new remote for it as origin, renaming the origin as upstream.
 `,
 }
 
@@ -28,7 +28,8 @@ func init() {
 /*
   $ gh fork
   [ repo forked on GitHub ]
-  > git remote add -f YOUR_USER git@github.com:YOUR_USER/CURRENT_REPO.git
+  > git remote rename origin upstream
+  > git remote add -f origin git@github.com:YOUR_USER/CURRENT_REPO.git
 
   $ gh fork --no-remote
   [ repo forked on GitHub ]
@@ -73,9 +74,17 @@ func fork(cmd *Command, args *Args) {
 	} else {
 		originRemote, _ := localRepo.OriginRemote()
 		originURL := originRemote.URL.String()
+		upstreamRemote, _ := localRepo.RemoteByName("upstream")
 		url := forkProject.GitURL("", "", true)
 		args.Replace("git", "remote", "add", "-f", forkProject.Owner, originURL)
 		args.After("git", "remote", "set-url", forkProject.Owner, url)
-		args.After("echo", fmt.Sprintf("new remote: %s", forkProject.Owner))
+		if upstreamRemote == nil {
+	 		args.After("git", "remote", "rename", "origin", "upstream")
+	 		args.After("echo", fmt.Sprintf("remote renamed: origin is now upstream"))
+			args.After("git", "remote", "rename", forkProject.Owner, "origin")
+			args.After("echo", fmt.Sprintf("new remote: %s", "origin"))
+		} else {
+			args.After("echo", fmt.Sprintf("new remote: %s", forkProject.Owner))
+		}
 	}
 }
--- a/features/authentication.feature
+++ b/features/authentication.feature
@@ -262,4 +262,5 @@ Feature: OAuth authentication
     And the file "../home/.config/hub" should contain "git.my.org"
     And the file "../home/.config/hub" should contain "user: mislav"
     And the file "../home/.config/hub" should contain "oauth_token: OTOKEN"
-    And the url for "mislav" should be "git@git.my.org:mislav/dotfiles.git"
+    And the url for "upstream" should be "git@git.my.org:evilchelu/dotfiles.git"
+    And the url for "origin" should be "git@git.my.org:mislav/dotfiles.git"
--- a/features/fork.feature
+++ b/features/fork.feature
@@ -15,10 +15,13 @@ Feature: hub fork
       post('/repos/evilchelu/dotfiles/forks', :host_name => 'api.github.com') { '' }
       """
     When I successfully run `hub fork`
-    Then the output should contain exactly "new remote: mislav\n"
+    Then the output should contain exactly "remote renamed: origin is now upstream\nnew remote: origin\n"
     And "git remote add -f mislav git://github.com/evilchelu/dotfiles.git" should be run
     And "git remote set-url mislav git@github.com:mislav/dotfiles.git" should be run
-    And the url for "mislav" should be "git@github.com:mislav/dotfiles.git"
+    And "git remote rename origin upstream" should be run
+    And "git remote rename mislav origin" should be run
+    And the url for "origin" should be "git@github.com:mislav/dotfiles.git"
+    And the url for "upstream" should be "git://github.com/evilchelu/dotfiles.git"
 
   Scenario: Fork the repository when origin URL is private
     Given the "origin" remote has url "git@github.com:evilchelu/dotfiles.git"
@@ -29,10 +32,13 @@ Feature: hub fork
       post('/repos/evilchelu/dotfiles/forks', :host_name => 'api.github.com') { '' }
       """
     When I successfully run `hub fork`
-    Then the output should contain exactly "new remote: mislav\n"
+    Then the output should contain exactly "remote renamed: origin is now upstream\nnew remote: origin\n"
     And "git remote add -f mislav ssh://git@github.com/evilchelu/dotfiles.git" should be run
     And "git remote set-url mislav git@github.com:mislav/dotfiles.git" should be run
-    And the url for "mislav" should be "git@github.com:mislav/dotfiles.git"
+    And "git remote rename origin upstream" should be run
+    And "git remote rename mislav origin" should be run
+    And the url for "origin" should be "git@github.com:mislav/dotfiles.git"
+    And the url for "upstream" should be "git@github.com:evilchelu/dotfiles.git"
 
   Scenario: --no-remote
     Given the GitHub API server:
@@ -41,7 +47,6 @@ Feature: hub fork
       """
     When I successfully run `hub fork --no-remote`
     Then there should be no output
-    And there should be no "mislav" remote
 
   Scenario: Fork failed
     Given the GitHub API server:
@@ -54,7 +59,6 @@ Feature: hub fork
       """
       Error creating fork: Internal Server Error (HTTP 500)\n
       """
-    And there should be no "mislav" remote
 
   Scenario: Unrelated fork already exists
     Given the GitHub API server:
@@ -70,9 +74,8 @@ Feature: hub fork
       """
       Error creating fork: mislav/dotfiles already exists on github.com\n
       """
-    And there should be no "mislav" remote
 
-Scenario: Related fork already exists
+  Scenario: Related fork already exists
     Given the GitHub API server:
       """
       get('/repos/mislav/dotfiles') {
@@ -81,7 +84,7 @@ Scenario: Related fork already exists
       """
     When I run `hub fork`
     Then the exit status should be 0
-    And the url for "mislav" should be "git@github.com:mislav/dotfiles.git"
+    And the url for "origin" should be "git@github.com:mislav/dotfiles.git"
 
   Scenario: Invalid OAuth token
     Given the GitHub API server:
@@ -103,8 +106,8 @@ Scenario: Related fork already exists
       """
     And HTTPS is preferred
     When I successfully run `hub fork`
-    Then the output should contain exactly "new remote: mislav\n"
-    And the url for "mislav" should be "https://github.com/mislav/dotfiles.git"
+    Then the output should contain exactly "remote renamed: origin is now upstream\nnew remote: origin\n"
+    And the url for "origin" should be "https://github.com/mislav/dotfiles.git"
 
   Scenario: Not in repo
     Given the current dir is not a repo
@@ -134,7 +137,8 @@ Scenario: Related fork already exists
     And I am "mislav" on git.my.org with OAuth token "FITOKEN"
     And "git.my.org" is a whitelisted Enterprise host
     When I successfully run `hub fork`
-    Then the url for "mislav" should be "git@git.my.org:mislav/dotfiles.git"
+    Then the url for "origin" should be "git@git.my.org:mislav/dotfiles.git"
+    And the "upstream" remote has url "git@git.my.org:evilchelu/dotfiles.git"
 
   Scenario: Enterprise fork using regular HTTP
     Given the GitHub API server:
@@ -150,4 +154,7 @@ Scenario: Related fork already exists
     And I am "mislav" on http://git.my.org with OAuth token "FITOKEN"
     And "git.my.org" is a whitelisted Enterprise host
     When I successfully run `hub fork`
-    Then the url for "mislav" should be "git@git.my.org:mislav/dotfiles.git"
+    Then "git remote rename origin upstream" should be run
+    And "git remote rename mislav origin" should be run
+    Then the url for "origin" should be "git@git.my.org:mislav/dotfiles.git"
+    And the "upstream" remote has url "git@git.my.org:evilchelu/dotfiles.git"


RichardLitt:
The current `origin` remote would be renamed as `upstream`, and instead of using the username as the name of the new remote, the new remote would be called `origin`.

mislav:
It's very hard to convince us to pull a breaking change to our library when you haven't even invested any words in trying to convince us.

RichardLitt:
Hey @mislav, sorry about that. You're totally right. 

`hub fork` currently works by forking a project and adding a new remote to the current dir. This remote is named after your username - so, in my case it would be `RichardLitt`. The thing is, I haven't seen this method of naming remotes anywhere else. Normally, on GitHub and in git in general (as far as I am aware! I may be wrong), the pattern is that origin refers to your own fork, and whatever the root, original, or upstream version is is named the `upstream` remote. 

My changes basically follow this naming convention. When you fork, instead of naming the new remote `RichardLitt`, it would rename `origin`, which would by default be the original branch, to `upstream`, and then it would rename `RichardLitt` to `origin`. This may be confusing - with one command, the remote that used to be `origin` now points somewhere else. However, this matches the other naming conventions I am more familiar with, and I think is something that would help users in general. 

What do you think?

RichardLitt:
There's two main detractions to doing this according to the way specified in my PR: the sudden shift in where `origin` points to, and if there already is an `upstream` branch or the repository you're forking off isn't actually the original repository and there already is an `upstream branch`. In the second case, I use your pre-existing method, of naming the remote `$username`. In the former, I think the cost is negligible compared to having git directories with different naming conventions for origin - which I would expect to be able to push to normally. Having to continually type `git remote -v` to see whether or not my fork was `origin` or `RichardLitt` gets increasingly tedious. 

Again, sorry for the lack of justification - forgot the most important part of code is communication. :)

mislav:
Thanks for explaining. We'll need to think about this. I'm not a giant fan of naming the remote by your GitHub username anymore since in that case it's hard to write automated scripts that interact with that remote (the script needs to know your username then). But I'm also not a big fan of breaking compatibility or doing other surprising things. For instance, I would be in favor of this naming scheme much more if it didn't entail us renaming your "origin" repo to "upstream" behind the scenes.

Any big change to a project should be argued for and thorougly documented. [This should give you a glimpse](https://twitter.com/mislav/status/542158292168884224) of what kind of a person I am and how to reason with me.

RichardLitt:
I'm not a big fan of renaming origin to upstream behind the scenes, either. It might be smart for me to include a notice in here that says this is done - at the moment, that doesn't happen. The thing is, I can't think of a better solution than the one I've implemented here. 

Normally I write a lot, too - last night I was coding late for the first time in a while, and my guard was down regarding the danger of me being terse. My bad. :)

zbeekman:
FWIW, I too dislike the current behavior and think that @RichardLitt’s approach is better than the current behavior, especially if it tells you about the change that’s made with the remote.

RichardLitt:
I've added in an echo saying 
```sh
remote renamed: origin is now upstream\n
```
This should help.

RichardLitt:
Another workaround that would negate the use of this pr is to encourage the user to set up this alias:

```sh
[alias]
	rr = git remote rename origin upstream && git remote rename RichardLitt origin
```

I'm not sure that is a good idea, though.

ryan-williams:
Just driving-by after noticing this from #762, but I'm skeptical of this level of dictation of users' preferences around `remote` names. It seems like it wouldn't be that hard to allow each user to configure:

1. what they like their `remote`s pointing to github repos to be called
1. what they like their `remote`s pointing to their forks of github repos to be called

and have scripts be robust to various specific values of those two things.

Hoisting the specific `upstream`/`origin` convention described here on people, especially if part of the motivation is to make other scripts/plumbing easier, seems likely to alienate / unpleasantly surprise.

> Having to continually type `git remote -v` ... gets increasingly tedious.

aside: @RichardLitt may I recommend an `alias gr="git remote -vv"` :) I use it often.

RichardLitt:
Allowing the users to configure would also be a good option. I think that the current system is more alienating and unpleasantly surprising than my fix, but that's opinion. 

aside: Huh. `man git-remote | grep vv` shows nothing, although `-vv` appears to be a cool option for `git branch`. Are you sure you don't mean `git branch -vv`, and if not, how is `remote -v` different from `remote -vv`?

ryan-williams:
Yea, weird I guess they're the same, and `-vv` is functionally the same as `-v -v` which is functionally the same as `-v`. So, make that `alias gr="git remote -v"`!

What do you find unpleasantly surprising about the status quo? You expect your existing `origin` remote to be renamed when you run `hub fork`?

RichardLitt:
Well, they're not exactly the same: `branch -vv` is different than `branch -v` (and more useful, actually). 

What's odd is that `origin` doesn't point to the repo where I have write access, and all of the sudden I have to remember for some repositories the remote I want is my username, which I never use otherwise.

ryan-williams:
Interesting. Is it the case that, before you run `hub fork`, you have a remote named `origin` that you lack write-access to? Or is `hub fork` creating the remote named `origin` that you lack write-access to?

RichardLitt:
No, when I clone I normally have a remote named `origin` without write access. The difference is that when I fork I normally do so through github, then clone, which causes `origin` to be my fork. If I've already cloned, I always set it up that `origin` points to my fork and `upstream` points to the origin.

mislav:
> The difference is that when I fork I normally do so through github, then clone, which causes origin to be my fork.

That seems to be what most GitHub users are doing. The only reason why people kept doing this is because if they cloned the canonical repo and started their edits on it, they wouldn't be able to push them anywhere, then they would have to get back to GitHub and fork the repo, then they would have to copy their fork's git URL and create a new remote with it manually, then push to that remote. It's a drag. No wonder people fork first, then make changes, then later decide to push and open PRs.

With hub, it's easy to fork on-demand from the command line when you're ready to push your changes. That's why with hub I suggest simply cloning the canonical repo as "origin" and fork only when you have something to push. That's personally my process. But I guess it's hard to change other people's old habits.

zbeekman:
@ryan-williams  said:

> It seems like it wouldn’t be that hard to allow each user to configure:
>
> - what they like their remotes pointing to github repos to be called
> - what they like their remotes pointing to their forks of github repos to be called

IMO this is the most flexible and appealing option. Provide a means for the user to setup a persistent set of defaults. I do follow @RichardLitt’s remote naming convention, and would be happy if there was some way I could configure hub to use that convention, no matter what the out of the box default is.

RichardLitt:
> With hub, it's easy to fork on-demand from the command line when you're ready to push your changes. That's why with hub I suggest simply cloning the canonical repo as "origin" and fork only when you have something to push. That's personally my process. But I guess it's hard to change other people's old habits.

That's my habit, now, too, but that's after learning that the way I was doing it before was suboptimal. I think that the change in this PR is something like a compromise, helping users who do have this bad habit along. That having been said...

Yes, allowing users to configure the names of their remotes makes more sense than my PR.

mislav:
> It seems like it wouldn’t be that hard to allow each user to configure:
> 
> * what they like their remotes pointing to github repos to be called
> * what they like their remotes pointing to their forks of github repos to be called

I dig those ideas. Those parameters could either be given on the command-line as flags, or they could be configured globally in hub configuration file.

jingweno:
What do you think of this?

* when `hub-fork` is run with a canonical repo (potentially outside of a git directory):

```
$ hub fork github/hub
> fork github/hub under current user's name, e.g., jingweno/hub
> git clone https://github.com/jingweno/hub
> cd hub && git remote add upstream https://github.com/github/hub
```

* existing behavior stays the same when `hub-fork` is run inside a canonical repo:

```
$ hub fork
> fork github/hub under current user's name, e.g., jingweno/hub
> git remote add jingweno https://github.com/jingweno/hub
```

* when `hub-fork` is run inside a canonical repo with the remote name specified:

```
$ hub fork -r fork
> fork github/hub under current user's name, e.g., jingweno/hub
> git remote add fork https://github.com/jingweno/hub
```

ryan-williams:
that doesn't seem to be satisfying the part where we let users specify defaults for:

> * what they like their remotes pointing to github repos to be called
> * what they like their remotes pointing to their forks of github repos to be called

i.e. in your example, the first block hardcodes `upstream` (which I would like to be able to configure differently) and the second block uses `$GITHUB_USER` as the remote name (which I personally would like the ability to override as the default, per #762). The third block, allowing overriding of the default with a cmdline flag, sounds good to me.

RichardLitt:
I'd be happy to take a stab at a global config file, if someone can point me to another command that uses one. Or should that be in the .gitconfig?

mislav:
On Fri, Feb 20, 2015 at 7:17 AM, Jingwen Owen Ou <notifications@github.com>
wrote:

> when hub-fork is run with a canonical repo (potentially outside of a git
> directory):


I don't think it's a good idea to have a command do 2 different operations
depending on the current directory.

mislav:
> I'd be happy to take a stab at a global config file

Take a look at [github/config](https://github.com/github/hub/blob/031657ce7a7257b489bd1b671ec89c07340de984/github/config.go). This configuration should be per GitHub host (the primary being GitHub.com but there can also be any number of Enterprise instances)

jingweno:
@mislav:

> I don't think it's a good idea to have a command do 2 different operations
depending on the current directory.

What if this is in `hub-clone`?

```
$ git clone --fork github/hub
> fork github/hub under current user's name, e.g., jingweno/hub
> git clone https://github.com/jingweno/hub
> cd hub && git remote add upstream https://github.com/github/hub
```

I found myself sometimes use origin as the fork and upstream as the canonical repo. This happens the most when I'm browsing on GitHub.com and I press the "fork" button. Then I'll manually set up my the remotes this way. I guess I'm influenced by [this GitHub doc](https://help.github.com/articles/syncing-a-fork/).

zbeekman:
@jingweno yes, that’s the doc that led me to start doing it that way too, and from that I assumed that it was ‘THE way’ :tm: to do it.

jingweno:
> I assumed that it was ‘THE way’ :tm: to do it.

I think there's no right way. But when I'm working on my fork, I tend to make it the default remote "origin", which is sort of like "origin" is the remote I'm working towards, like any other case. I guess I treat "origin" a bit special is mostly out of perception while `git` treats all remotes the same.

RichardLitt:
@mislav yeah, I don't know Go well enough to be able to do this. I tried looking at that file you suggested, and I'm just totally lost as to how I would implement this. 

I think this PR should be closed for now, since the main code I did for it isn't going to be implemented due to the problems raised in the discussion above. It might be a good interim measure, but I think someone else should work on the global config.

bvalerius:
I would like to add my opinion on this subject. I fully expected `hub fork` to work exactly like the flow on GitHub. You fork a repository, clone it as origin, and add the original repository you forked off of as upstream. This should be the default in my opinion. Here's why: https://help.github.com/articles/fork-a-repo/

Now, anyone who visits and learns about forking and using "origin" and "upstream" as the remotes, is going to have quite a difficult time understanding this backwards (compared to the article) way of forking a repository with `hub fork`.

RichardLitt:
The link to the Help articles doesn't actually say why it is named `upstream` and `origin`, though, just how to do it. 

The question is - are we building a tool to help us with a Git, or a tool to help us with GitHub?

ryan-williams:
seems like that convention is official / de-facto-standard enough that maybe it does make sense for that to be how it works. I think it definitely would have confused / will confuse me but I am not necessarily representative, so I have no further objections to whatever y'all want to do here

bvalerius:
The process on GitHub is Fork -> Clone Fork (origin). Now, the name you choose for the "upstream" repository can be anything you want, because you have to go and add it on after you clone. I think that keeping with conventions and having that remote named "upstream", and then allowing for explicit renaming, would be the best option.

As far as writing a tool to help with Git or a tool to help with GitHub, I think the name implies a tool to help with GitHub, but ends up being a bit of both.

lowne:
Indirectly +1; I have a separate but related problem, `hub issue` (which is *awesome*) lists nothing since everything is in `upstream` as per conventions discussed above - I had to set up an empty repo and `git remote add origin` elsewhere (because I *really* want `hub issue`)

shaleh:
So what is holding this change up? Following the GitHub documented procedure makes sense. It is what we have been taught to do. Is the issue simply that the PR is missing the configuration option as suggested?

mislav:
@shaleh This change is held up mostly by being a radical departure from how the `fork` command behaved previously, as well as lack of consensus about whether we want to perform remotes renaming or not. I think it's an interesting idea, but we also want to follow the principle of least surprise.

If we do make this feature, I'm in favor of adding a configuration option to turn it off. The configuration option would get saved in `~/.config/hub`, but first we'll need to create a subcommand that will allow us to manage these global configuration options.

shaleh:
Good to hear there is still interest. Do let us know if you need help, testers, etc.

RichardLitt:
@mislav Do you have an example of other commands which use the configuration file? I couldn't see any clearly in `commands/`. Would be good to have something else to work off of.

mislav:
@RichardLitt Unfortunately the configuration isn't the cleanest implementation and commands don't currently access it directly. The only thing we use the configuration right now is for authentication info.

My idea for configurable hub commands would be to introduce new command-line flags for a specific feature (e.g. `hub foo --feature` or `hub foo --no-feature`), and then extend hub with a capability to read default flags from files such as `.hubrc` (local directory) or `~/.hubrc`.

simplyianm:
whoops. Didn't see this.

+1 to this change. I think adding a `--upstream` or `-u` flag would be sufficient to make this change non-breaking.

salewski:
Just mentioning this for folks like me who found this thread while looking for a tool to automatically setup the `'upstream'` remote to the parent repo when cloning my own fork (creating the fork, if necessary), while leaving an `'origin'` remote pointing at my fork. The https://github.com/sociomantic-tsunami/git-hub project has a `'clone'` operation that provides this behavior.

dholdren:
+1 I'd like to see this change merged in some fashion.  This is the way I like to work, and it's in line with the GitHub guides: https://help.github.com/articles/configuring-a-remote-for-a-fork/

un1versal:
> +1 I'd like to see this change merged in some fashion. This is the way I like to work, and it's in line with the GitHub guides: https://help.github.com/articles/configuring-a-remote-for-a-fork/

This is just a good example how documentation and what happens in reality doesn't match.

Its confusing to say the least.

The origin is the project that is the original repo, (for these purposes its never been forked) so its natural one would expect it to be called origin right?

Wrong.

What happens is I forked **hub** just now to show how this is just like everything else which has been "GitHubified". So...

I now have a **hub** fork, I clone that **hub** fork locally and open a terminal and run ```git remote -v```
It spits out this....

```
github  https://github.com/github/hub.git (fetch)
github  https://github.com/github/hub.git (push)
origin  https://github.com/un1versal/hub.git (fetch)
origin  https://github.com/un1versal/hub.git (push)
```

So now **what should be named upstream** is actually named github and what should be named something else is called origin, because the logic is I suspect, thats what the origin of the clone....

Lets try this directly.... by cloning the **hub** directly.

```
$ git clone https://github.com/github/hub.git
Cloning into 'hub'...
remote: Counting objects: 12323, done.
remote: Compressing objects: 100% (2/2), done.
remote: Total 12323 (delta 0), reused 0 (delta 0), pack-reused 12321
Receiving objects: 100% (12323/12323), 3.60 MiB | 1.11 MiB/s, done.
Resolving deltas: 100% (7456/7456), done.
$ cd hub
& hub [master ≡]> git remote -v
origin  https://github.com/github/hub.git (fetch)
origin  https://github.com/github/hub.git (push)
```

OK so now **hub** is actually the **origin** as it should be, but this isnt a fork according to GitHub guidelines on using pull requests for contributing to projects is https://help.github.com/articles/using-pull-requests/ which specifically says you going to have to **fork**....

And thus here we are, somewhere in between what should be, what it actually is and hence its a mess. so if you follow https://help.github.com/categories/collaborating-on-projects-using-issues-and-pull-requests/ while eventually you end up with something it takes manual work that should be done behind the scenes.

I guess this is what this PR suggests, but the problem is, no matter what system you implement it will never be perfect...

This is why wherever you go in GitHub, each contributor is confused to find different projects have different expectations on how you should contribute, which as I see it, is a side effect of things being "GitHubified".

;tl;dr

Documentation and what happens in practice should be consistent and by default you shouldn't have to spend time figuring out why things aren't as logical as you expect.

That said this is the fastest best way to develop/contribute imo. Is it perfect? No. is it better than most alternatives, Definitely.

I cant wait to see what a world post Git looks like.

un1versal:
https://github.com/github/hub/issues/762#issuecomment-232025929

adjohnson916:
Can we at least add an option flag for this behavior, leaving the default?

mislav:
@adjohnson916 Sure, there is enough interest for this feature in my opinion to warrant an option. I'm gonna try to work on some sort of configuration mechanism where users could define default options on their system via a configuration file.

un1versal:
> @adjohnson916 Sure, there is enough interest for this feature in my opinion to warrant an option. I'm gonna try to work on some sort of configuration mechanism where users could define default options on their system via a configuration file.

Ideally provide a template along so it can just be edited or options commented out or such

maxandersen:
just my 2 cents - hub's insistent on having the canonical repo be named origin and forks of it by username was what ended up being how I can much better explain and work with fork workflows.

I will argue that calling your own forked repo origin is a bad antipattern. Thus please do *not* change the current awesome defaults.

RichardLitt:
@maxandersen, I agree with you. I've changed my workflow since opening this issue, and I think not using upstream is probably better. At this point, this won't be merged with changed defaults, though; rather, it needs to be reworked to add an optional flag.

RichardLitt:
I am no longer interested in merging this. Feel free to reopen it if it is useful, or to log a new issue elsewhere. For now, I think keeping this open is disingenuous.

PR #797, opened 2015-02-08
Upped Ruby version

RichardLitt:
When running the test file, I got this error: 

```sh
$ script/bootstrap 
go version go1.4.1 darwin/amd64
ruby 2.0.0p481 (2014-05-08 revision 45883) [universal.x86_64-darwin14]
script/bootstrap: line 17: bundle: command not found
You need Ruby 2.1 and Bundler to run hub tests
```

Upped the Ruby version to reflect this.

PR #799, opened 2015-02-08
Update `script/bootstrap` to recommended Go and Ruby version

jingweno:
Related PR: https://github.com/github/hub/pull/797

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -9,14 +9,14 @@ fi
 STATUS=0
 
 if ! go version; then
-  echo "You need to install Go 1.2 to build hub" >&2
+  echo "You need to install Go 1.4.1 or higher to build hub" >&2
   STATUS=1
 fi
 
 { ruby --version
   bundle install --path vendor/bundle -j 4
 } || {
-  echo "You need Ruby 2.1 and Bundler to run hub tests" >&2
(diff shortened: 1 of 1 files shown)


script/bootstrap
```diff
@@ -9,14 +9,14 @@ fi
...
-  echo "You need to install Go 1.2 to build hub" >&2
+  echo "You need to install Go 1.4.1 or higher to build hub" >&2
   STATUS=1
 fi
 
 { ruby --version
   bundle install --path vendor/bundle -j 4
 } || {
-  echo "You need Ruby 2.1 and Bundler to run hub tests" >&2
+  echo "You need Ruby 1.9 or higher and Bundler to run hub tests" >&2
```
mislav: I'm torn about this. On one hand, our test suite does support Ruby 1.9. On the other hand, I put 2.1 in there because I wanted to suggest to people to install the most recent version if they're going to install Ruby for hub development. Ruby versions older than 2.1.4 have all sorts of compilation issues on newer systems that were fixed in the Ruby build process of newer releases.

jingweno: This is to make [`CONTRIBUTING.md`](https://github.com/github/hub/blob/master/CONTRIBUTING.md) and this script consistent - it caused confusion like #797. I don't have a problem of suggesting Ruby 2.1 if that works better. But it'd be nice that our build system (Travis) is running the same version.

PR #800, opened 2015-02-08
Add `script/bootstrap` to gh pages

### LINKED ISSUE #795: Make Go requirement more explicit
Hi,
I am having same issue as in #710.

I see in #712  that 1.3 is the minimum requirment for go.
I visited https://hub.github.com/ and it doens't saying anything in the frontpage about the a required version.

Also, the build script, you should check what version i'm running, to prevent having an error message that doesn't say anything about my Go version being too old.

thanks.

jingweno:
This fixes #795


mislav:
I wouldn't put `script/bootstrap` as prerequisite for building hub because `script/bootstrap` requires Ruby and Bundler, which are not necessary for building Go version.

Why not check Go version in `script/build` instead?

jingweno:
Good call. Closing.

PR #802, opened 2015-02-10
Better arguments processing for `init` command

mislav:
Now the optional `directory` argument to `git init` is detected by looking for the first "word" in arguments that isn't a value for one of `--template`, `--shared`, or `--separate-git-dir` params.

This is done by simply iterating through args instead of pulling in a full-blown flag-parsing library. This is also the approach we take in the `clone` command.

This is an alternative implementation to #793. @jingweno, let me know what you think.

Fixes #792

### DIFF
--- a/commands/init.go
+++ b/commands/init.go
@@ -2,6 +2,8 @@ package commands
 
 import (
 	"path/filepath"
+	"regexp"
+	"strings"
 
 	"github.com/github/hub/github"
 	"github.com/github/hub/utils"
@@ -28,10 +30,8 @@ func init() {
   > git remote add origin git@github.com:USER/REPO.git
 */
 func gitInit(command *Command, args *Args) {
-	if !args.IsParamsEmpty() {
-		err := transformInitArgs(args)
-		utils.Check(err)
-	}
+	err := transformInitArgs(args)
+	utils.Check(err)
 }
 
 func transformInitArgs(args *Args) error {
@@ -39,32 +39,38 @@ func transformInitArgs(args *Args) error {
 		return nil
 	}
 
-	var (
-		name   string
-		newDir bool
-		err    error
-	)
+	var err error
+	dirToInit := "."
+	hasValueRegxp := regexp.MustCompile("^--(template|separate-git-dir|shared)$")
 
-	if args.IsParamsEmpty() {
-		name, err = utils.DirName()
-		if err != nil {
-			return err
+	// Find the first argument that isn't related to any of the init flags.
+	// We assume this is the optional `directory` argument to git init.
+	for i := 0; i < args.ParamsSize(); i++ {
+		arg := args.Params[i]
+		if hasValueRegxp.MatchString(arg) {
+			i++
+		} else if !strings.HasPrefix(arg, "-") {
+			dirToInit = arg
+			break
 		}
-	} else {
-		name = args.LastParam()
-		newDir = true
 	}
 
-	project := github.NewProject("", name, "")
+	dirToInit, err = filepath.Abs(dirToInit)
+	if err != nil {
+		return err
+	}
+
+	// Assume that the name of the working directory is going to be the name of
+	// the project on GitHub.
+	projectName := strings.Replace(filepath.Base(dirToInit), " ", "-", -1)
+	project := github.NewProject("", projectName, "")
 	url := project.GitURL("", "", true)
 
-	cmds := []string{"git"}
-	if newDir {
-		cmds = append(cmds, "--git-dir", filepath.Join(name, ".git"))
+	addRemote := []string{
+		"git", "--git-dir", filepath.Join(dirToInit, ".git"),
+		"remote", "add", "origin", url,
 	}
-
-	cmds = append(cmds, "remote", "add", "origin", url)
-	args.After(cmds...)
+	args.After(addRemote...)
 
 	return nil
 }
--- a/commands/init_test.go
+++ b/commands/init_test.go
@@ -10,13 +10,13 @@ import (
 	"github.com/github/hub/github"
 )
 
-func setup() {
+func setupInitContext() {
 	os.Setenv("HUB_PROTOCOL", "git")
 	github.CreateTestConfigs("jingweno", "123")
 }
 
 func TestEmptyParams(t *testing.T) {
-	setup()
+	setupInitContext()
 
 	args := NewArgs([]string{"init"})
 	err := transformInitArgs(args)
@@ -26,7 +26,7 @@ func TestEmptyParams(t *testing.T) {
 }
 
 func TestFlagToAddRemote(t *testing.T) {
-	setup()
+	setupInitContext()
 
 	args := NewArgs([]string{"init", "-g", "--quiet"})
 	err := transformInitArgs(args)
@@ -48,7 +48,7 @@ func TestFlagToAddRemote(t *testing.T) {
 }
 
 func TestInitInAnotherDir(t *testing.T) {
-	setup()
+	setupInitContext()
 
 	args := NewArgs([]string{"init", "-g", "--template", "mytpl", "my project"})
 	err := transformInitArgs(args)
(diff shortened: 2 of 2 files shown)


commands/init.go
```diff
@@ -28,10 +30,8 @@ func init() {
   > git remote add origin git@github.com:USER/REPO.git
 */
 func gitInit(command *Command, args *Args) {
-	if !args.IsParamsEmpty() {
```
mislav: This check was never needed. `transformInitArgs` will bail out early as a no-op if the `-g` flag is not present, such as in the case where args are an empty list.

commands/init.go
```diff
@@ -39,32 +39,38 @@ func transformInitArgs(args *Args) error {
...
+			dirToInit = arg
+			break
 		}
-	} else {
-		name = args.LastParam()
-		newDir = true
 	}
 
-	project := github.NewProject("", name, "")
+	dirToInit, err = filepath.Abs(dirToInit)
```
mislav: In the standard use case (no explicit `directory` for git init given), the path that will get expanded to absolute path here is `.`, which is just an alternate way of doing `os.Getwd()`, but without having to maintain separate code paths for the two scenarios.

commands/init.go
```diff
@@ -39,32 +39,38 @@ func transformInitArgs(args *Args) error {
...
+	if err != nil {
+		return err
+	}
+
+	// Assume that the name of the working directory is going to be the name of
+	// the project on GitHub.
+	projectName := strings.Replace(filepath.Base(dirToInit), " ", "-", -1)
```
mislav: Replacing spaces with dashes is what `DirName()` did for us internally, but having that logic here makes it more clear how we construct a github project name. Personally, I found the fact that `DirName()` doesn't return the exact directory name kind of surprising. Also, the code here ensures that the project name is sanitized even if it came from the explicit `directory` argument rather than current working directory.

commands/init_test.go
```diff
@@ -1,32 +1,92 @@
...
 	"os"
-	"regexp"
 	"testing"
+
+	"github.com/github/hub/Godeps/_workspace/src/github.com/bmizerany/assert"
+	"github.com/github/hub/github"
 )
 
-func TestTransformInitArgs(t *testing.T) {
+func setup() {
```
mislav: It doesn't seem that Go `testing` package exposes any built-in way of declaring setup/teardown blocks for the whole suite.

commands/init.go
```diff
@@ -39,32 +39,38 @@ func transformInitArgs(args *Args) error {
...
 
-	if args.IsParamsEmpty() {
-		name, err = utils.DirName()
-		if err != nil {
-			return err
+	// Find the first argument that isn't related to any of the init flags.
+	// We assume this is the optional `directory` argument to git init.
+	for i := 0; i < args.ParamsSize(); i++ {
+		arg := args.Params[i]
+		if hasValueRegxp.MatchString(arg) {
```
jingweno: Does this cover the case when flags appear as `--key=val`? For example `--template=DIR`. It'd be nice to have a test case as well.

commands/init.go
jingweno: I agree there's some refactoring needed for `DirName()`.

What do you think of moving the line `strings.Replace(filepath.Base(dirToInit), " ", "-", -1)` to the github package as  `github.CleanName(name)` so that we could use for other places. I'm thinking of removing `DirName()` and make it explicit for other places.

commands/init.go
jingweno: In general, I like the idea of reducing special cases with a default value "." for `dirToInit`.

commands/init.go
```diff
@@ -39,32 +39,38 @@ func transformInitArgs(args *Args) error {
...
+	// the project on GitHub.
+	projectName := strings.Replace(filepath.Base(dirToInit), " ", "-", -1)
+	project := github.NewProject("", projectName, "")
 	url := project.GitURL("", "", true)
 
-	cmds := []string{"git"}
-	if newDir {
-		cmds = append(cmds, "--git-dir", filepath.Join(name, ".git"))
+	addRemote := []string{
+		"git", "--git-dir", filepath.Join(dirToInit, ".git"),
```
jingweno: :1 to less special cases

commands/init_test.go
jingweno: There's no setup/teardown for a test suite. Go 1.4 introduces `TestMain` (https://golang.org/doc/go1.4#minor_library_changes, scroll down the bullet point) but that's a global setup for all tests. However, there're other libraries support such feature, e.g., we switched to https://github.com/stretchr/testify in go-octokit which has setup/teardown for a test suite.

Regarding to the changes, how about renaming `setup()` to something less generic like `setupTestConfigs()`? The method is available to the whole `command` package and it'd be confusing for test cases in other files.

commands/init.go
jingweno: Yup, good call on this :+1:

jingweno:
@mislav It seems parsing the flags manually is more verbose, less clear and more error pone than https://github.com/github/hub/pull/793? It also causes miss cases like https://github.com/github/hub/pull/802#discussion_r24385613. However, there're tradeoffs to hardcode all init flags in https://github.com/github/hub/pull/793: if `git` add or remove flags in the future we need to change our code as well. If hard coding flags is the concern, we could hardcode less by using the flags parsing library to only parse `template`, `shared` and `separate-git-dir`, and ignore the rest of boolean flags.

However, I like the idea of this PR which uses "." as current dir to reduce special cases.

commands/init.go
mislav: Since `--key=val` is a single argument, it's covered by ignoring everything start starts with `-`.

commands/init.go
mislav: I don't think we need to wrap this with a helper method. I would just repeat the same logic whenever we need it. That makes it clear to anybody who's reading the code about how exacly we construct the project name. It also makes it possible to tweak the logic on case-by-case basis. I would just get rid of `DirName()` entirely.

commands/init_test.go
mislav: Oh `setup` will be shared across the whole package? That's no good, because it probably means that we're unable to create a `setup` method in another test file. Is there a way to have a setup block be defined only in the current file, for example with a func and a closure?

Also, our current setup edits an environment variable but doesn't reset it at the end of the test. Not a huge deal, but it would be ideal if we had a way of editing environment variables just for the duration of a single test and have it automatically reset, like Aruba does in our cukes.

commands/init.go
mislav: But yeah it's worth adding a test for that, sure.

jingweno: Ah, I missed the case of ignoring everything starting with `-`: https://github.com/github/hub/blob/git-init-dirs/commands/init.go#L52

commands/init_test.go
jingweno: > Is there a way to have a setup block be defined only in the current file, for example with a func and a closure?

Unfortunately not.  Method name is unique for the same package, regardless which file it's in.

> Also, our current setup edits an environment variable but doesn't reset it at the end of the test.

That's no good. It'd be nice to reset the mutated global states if possible in tests. We've similar cases for fixture where we rest states: https://github.com/github/hub/blob/af9163b699b34ffa167ef99853ff2f7374d7b01a/git/git_test.go#L12-L13.

mislav:
I've renamed `setup` and added a test for `--param=value`. Going ahead and merging this because of imminent release, but in the future we should definitely think about how to avoid hardcoding git flags. Using a flag parsing library doesn't help us in this regard because with it we still need to hardcode a list of flags for init that might match the current git version but might change in future git versions. Ideally we would parse `git init -h` output and detect which flags have values and which don't.

jingweno:
> Going ahead and merging this because of imminent release, but in the future we should definitely think about how to avoid hardcoding git flags. 

:+1:

PR #803, opened 2015-02-10
Extract logic of printing out to console

### LINKED ISSUE #801: One Command test outputs text to stdout
[This specific commands test](https://github.com/github/hub/blob/1409606542e9140324d8b9c113bdb095355f8b4b/commands/commands_test.go#L133-141) outputs `usage: foo` to stdout. This output is only visible if some other test in "commands" package fails.

Ideally, tests should output nothing to stdout, as all their output should be captured and asserted.

jingweno:
Introduce a `console` package which encapsulates logic of printing out
to console. This provides flexibility of discarding output in tests.
It also allows us to unify format of printing to console (if we want to).

This fixes #801.

### DIFF
--- a/commands/commands_test.go
+++ b/commands/commands_test.go
@@ -1,10 +1,18 @@
 package commands
 
 import (
+	"io/ioutil"
+	"os"
 	"testing"
 
 	"github.com/github/hub/Godeps/_workspace/src/github.com/bmizerany/assert"
+	"github.com/github/hub/ui"
 )
+
+func TestMain(m *testing.M) {
+	ui.Default = ui.Console{Stdout: ioutil.Discard, Stderr: ioutil.Discard}
+	os.Exit(m.Run())
+}
 
 func TestCommandUseSelf(t *testing.T) {
 	c := &Command{Usage: "foo"}
--- a/commands/alias.go
+++ b/commands/alias.go
@@ -2,10 +2,12 @@ package commands
 
 import (
 	"fmt"
-	"github.com/github/hub/utils"
 	"os"
 	"path/filepath"
 	"strings"
+
+	"github.com/github/hub/ui"
+	"github.com/github/hub/utils"
 )
 
 var cmdAlias = &Command{
@@ -61,7 +63,7 @@ func alias(command *Command, args *Args) {
 			alias = "alias git=hub"
 		}
 
-		fmt.Println(alias)
+		ui.Println(alias)
 	} else {
 		var profile string
 		switch shell {
@@ -82,7 +84,7 @@ func alias(command *Command, args *Args) {
 		}
 
 		msg := fmt.Sprintf("# Wrap git automatically by adding the following to %s:\n", profile)
-		fmt.Println(msg)
+		ui.Println(msg)
 
 		var eval string
 		switch shell {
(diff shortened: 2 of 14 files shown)


commands/commands_test.go
```diff
@@ -1,11 +1,16 @@
...
 
 import (
+	"io/ioutil"
 	"testing"
 
 	"github.com/github/hub/Godeps/_workspace/src/github.com/bmizerany/assert"
 )
 
+func init() {
+	UI = ui{Stdout: ioutil.Discard, Stderr: ioutil.Discard}
```
jingweno: This is the code muting the print out to console in tests

commands/commands_test.go
```diff
@@ -1,11 +1,19 @@
...
+	"io/ioutil"
+	"os"
 	"testing"
 
 	"github.com/github/hub/Godeps/_workspace/src/github.com/bmizerany/assert"
+	"github.com/github/hub/console"
 )
 
+func TestMain(m *testing.M) {
+	console.Default = console.Console{Stdout: ioutil.Discard, Stderr: ioutil.Discard}
```
jingweno: I did some refactoring again. This is the new code to mute the print out to console in tests

mislav: Since TestMain affects all tests everywhere, should we put it in a file where it's more apparent that this is a global setting?

Also, shouldn't we capture the output instead of discard it?

mislav:
It's a good start, but "console" might be a slight misnomer because a unix executable doesn't know and shouldn't care where its output is going or who's going to read it.

Is there an option to keep our `fmt.Println` and similar but somehow capture and redirect all standard output in tests to a string so we can inspect it? This is trivial in Ruby.

commands/commands_test.go
jingweno: > Since TestMain affects all tests everywhere

This `TestMain` will only affect the commands package. We'd need to put it into a test file in the rest of packages if we want this to be global.

> Also, shouldn't we capture the output instead of discard it?

I'm not aware of any test that asserts the output yet. For [`TestCommandCall`](https://github.com/github/hub/blob/master/commands/commands_test.go#L130) mentioned in #801, we're checking the the function is called and the arguments are correctly passed in.

jingweno:
> "console" might be a slight misnomer because a unix executable doesn't know and shouldn't care where its output is going

Any suggestion? I'm thinking something as generic as `ui.Infoln`. or `log.Infoln`.

> Is there an option to keep our fmt.Println and similar but somehow capture and redirect all standard output in tests to a string so we can inspect it?

Yup, `Console` expects `Stdout` and `Stderr` to be `io.Writer`, which is an interface to all writer, including [bytes of string](http://golang.org/pkg/bytes/#Buffer), so:

```go
var b bytes.Buffer
console.Default = Console{Stdout: &b, Stderr: &b}
// do something with b
```

jingweno:
@mislav Any more feedback here?

mislav:
I think "UI" is a good name, because it's an interface with whoever is using the tool from the outside. I'd also like to keep the `Println` and `Printf` names.

How about:

* `ui.Println`
* `ui.Printf`
* `ui.Errorln`
* `ui.Errorf`

jingweno:
> I think "UI" is a good name, because it's an interface with whoever is using the tool from the outside.

I've extracted `UI` into [an interface](https://github.com/github/hub/blob/fix_test_output/ui/ui.go) and `Console` as an implementation.

> I'd also like to keep the Println and Printf names.

After a second thought, I agree `Println` and `Printf` are better since they are the same as the ones from `fmt`. Fixed with `--amend`.

jingweno:
Merging since all comments are addressed.

PR #806, opened 2015-02-13
Read title and body by taking into account Windows line endings

### LINKED ISSUE #805: pull-request from file on windows does not parse title, body correctly
When creating a pull request on windows using `hub pull-request -F pull.txt` where pull.txt is:
```
my pull title

my description

another line
```
will create a pull request with the title `my pull title my description another line` and with no body.

The root cause comes from readMsg function in [utils.go](https://github.com/github/hub/blob/master/commands/utils.go#L97) where the input file is split on `"\n\n"` to obtain title and body.

In windows this separator should be `"\r\n\r\n"`

jingweno:
`bufio.Scanner` takes care of line endings difference for different OSes. See http://golang.org/pkg/bufio/#ScanLines.

This fixes #805

### DIFF
--- a/commands/utils.go
+++ b/commands/utils.go
@@ -1,14 +1,15 @@
 package commands
 
 import (
+	"bufio"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


mislav:
Nice, thanks!

PR #813, opened 2015-02-17
`Args.InsertParam` resets index to param size if it’s out of bound

### LINKED ISSUE #812: Crash report - Index 1 is out of bound
```
$ git checkout -b https://github.com/<private>/<private>/pull/12
```

Results in this:

Error (*errors.errorString): `Index 1 is out of bound`

Stack:

```
goroutine 1 [running]:
github.com/github/hub/commands.(*Args).InsertParam(0xc20805e1b0, 0x1, 0xc2082b5d38, 0x4, 0x4)
	/var/folders/w8/dp0lmdjs2vs51lsxbrf4s49h0000gn/T/go/src/github.com/github/hub/commands/args.go:99 +0x170
github.com/github/hub/commands.transformCheckoutArgs(0xc20805e1b0, 0x0, 0x0)
	/var/folders/w8/dp0lmdjs2vs51lsxbrf4s49h0000gn/T/go/src/github.com/github/hub/commands/checkout.go:102 +0xd63
github.com/github/hub/commands.checkout(0x6876a0, 0xc20805e1b0)
	/var/folders/w8/dp0lmdjs2vs51lsxbrf4s49h0000gn/T/go/src/github.com/github/hub/commands/checkout.go:36 +0x40
github.com/github/hub/commands.(*Command).Call(0x6876a0, 0xc20805e1b0, 0x0, 0x0)
	/var/folders/w8/dp0lmdjs2vs51lsxbrf4s49h0000gn/T/go/src/github.com/github/hub/commands/commands.go:46 +0x189
github.com/github/hub/commands.(*Runner).Call(0xc20802b120, 0x6876a0, 0xc20805e1b0, 0x0, 0x0, 0x0)
	/var/folders/w8/dp0lmdjs2vs51lsxbrf4s49h0000gn/T/go/src/github.com/github/hub/commands/runner.go:89 +0x54
github.com/github/hub/commands.(*Runner).Execute(0xc20802b120, 0x0, 0x0, 0x0)
	/var/folders/w8/dp0lmdjs2vs51lsxbrf4s49h0000gn/T/go/src/github.com/github/hub/commands/runner.go:81 +0x215
main.main()
	/private/tmp/hub-xGkPdT/hub-2.2.0/main.go:12 +0x40

```

Runtime:

```
GOOS: darwin
GOARCH: amd64
```

Running:

```
$ git version
git version 2.3.0
hub version 2.2.0
```

jingweno:
`Args.InsertParam` panics about index out of bound if the index is larger than param size. These changes reset the index to param size for such case. This is equivalent to appending new items to the end.

This fixes #812.

### DIFF
--- a/commands/checkout.go
+++ b/commands/checkout.go
@@ -64,6 +64,11 @@ func transformCheckoutArgs(args *Args) error {
 		return nil
 	}
 
+	err = sanitizeCheckoutFlags(args)
+	if err != nil {
+		return err
+	}
+
 	id := pullURLRegex.FindStringSubmatch(projectPath)[1]
 	gh := github.NewClient(url.Project.Host)
 	pullRequest, err := gh.PullRequest(url.Project, id)
@@ -99,6 +104,18 @@ func transformCheckoutArgs(args *Args) error {
 
 	remoteName := fmt.Sprintf("%s/%s", user, branch)
 	replaceCheckoutParam(args, checkoutURL, newBranchName, remoteName)
+
+	return nil
+}
+
+func sanitizeCheckoutFlags(args *Args) error {
+	if i := args.IndexOfParam("-b"); i != -1 {
+		return fmt.Errorf("Unsupported flag -b when checking out pull request")
+	}
+
+	if i := args.IndexOfParam("--orphan"); i != -1 {
+		return fmt.Errorf("Unsupported flag --orphan when checking out pull request")
+	}
 
 	return nil
 }
--- a/commands/checkout.go
+++ b/commands/checkout.go
@@ -97,9 +97,14 @@ func transformCheckoutArgs(args *Args) error {
 		args.Before("git", "remote", "add", "-f", "-t", branch, user, u)
 	}
 
-	idx := args.IndexOfParam(checkoutURL)
-	args.RemoveParam(idx)
-	args.InsertParam(idx, "--track", "-B", newBranchName, fmt.Sprintf("%s/%s", user, branch))
+	remoteName := fmt.Sprintf("%s/%s", user, branch)
+	replaceCheckoutParam(args, checkoutURL, newBranchName, remoteName)
 
 	return nil
+}
+
+func replaceCheckoutParam(args *Args, checkoutURL, branchName, remoteName string) {
+	idx := args.IndexOfParam(checkoutURL)
+	args.RemoveParam(idx)
+	args.InsertParam(idx, "--track", "-B", branchName, remoteName)
 }
--- a/commands/checkout_test.go
+++ b/commands/checkout_test.go
@@ -1,0 +1,16 @@
+package commands
+
+import (
+	"testing"
+
+	"github.com/github/hub/Godeps/_workspace/src/github.com/bmizerany/assert"
+)
+
+func TestReplaceCheckoutParam(t *testing.T) {
+	checkoutURL := "https://github.com/github/hub/pull/12"
+	args := NewArgs([]string{"checkout", "-b", checkoutURL})
+	replaceCheckoutParam(args, checkoutURL, "jingweno", "origin/master")
+
+	cmd := args.ToCmd()
+	assert.Equal(t, "git checkout -b --track -B jingweno origin/master", cmd.String())
+}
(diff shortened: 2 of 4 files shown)


commands/checkout_test.go
```diff
@@ -0,0 +1,16 @@
...
+	"github.com/github/hub/Godeps/_workspace/src/github.com/bmizerany/assert"
+)
+
+func TestReplaceCheckoutParam(t *testing.T) {
+	checkoutURL := "https://github.com/github/hub/pull/12"
+	args := NewArgs([]string{"checkout", "-b", checkoutURL})
+	replaceCheckoutParam(args, checkoutURL, "jingweno", "origin/master")
+
+	cmd := args.ToCmd()
+	assert.Equal(t, "git checkout -b --track -B jingweno origin/master", cmd.String())
```
mislav: Not sure about this assertion. The original invocation is an invalid usage of the `PULLREQ-URL` parameter. Maybe we shouldn't transform the arguments in this case?

jingweno: We could do a check of `-b` before everything else, currently it fails on the `git-checkout`:

```
$ ./hub checkout -b https://github.com/github/hub/pull/813
Updating github
From https://github.com/github/hub
 * [new branch]      relax_insert_param_index -> github/relax_insert_param_index
fatal: -b, -B and --orphan are mutually exclusive
```

I would still want to keep the changes of `Args.InsertParam` though since it's more resilient to exceptions.

mislav: > I would still want to keep the changes of Args.InsertParam though since it's more resilient to exceptions.

Sure!

jingweno: I added the check for `-b` and `--orphan` when checking out a pull request: https://github.com/github/hub/commit/6e0d23d6199283d14ac1132d58ed3fcd9db41b26

jingweno:
:shipit: ?

commands/checkout_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+	cmd := args.ToCmd()
+	assert.Equal(t, "git checkout --track -B jingweno origin/master", cmd.String())
+}
+
+func TestTransformCheckoutArgs(t *testing.T) {
+	args := NewArgs([]string{"checkout", "-b", "https://github.com/github/hub/pull/12"})
+	err := transformCheckoutArgs(args)
+
+	assert.Equal(t, "Unsupported flag -b when checking out pull request", fmt.Sprintf("%s", err))
```
mislav: What is `fmt.Sprintf("%s", err)` for?

commands/checkout.go
```diff
@@ -97,9 +102,26 @@ func transformCheckoutArgs(args *Args) error {
...
-	args.InsertParam(idx, "--track", "-B", newBranchName, fmt.Sprintf("%s/%s", user, branch))
+	remoteName := fmt.Sprintf("%s/%s", user, branch)
+	replaceCheckoutParam(args, checkoutURL, newBranchName, remoteName)
 
 	return nil
 }
+
+func sanitizeCheckoutFlags(args *Args) error {
+	if i := args.IndexOfParam("-b"); i != -1 {
+		return fmt.Errorf("Unsupported flag -b when checking out pull request")
```
mislav: It's a good start. In the future we can decide whether we want to support something like this syntax:

```
hub checkout -b LOCAL-BRANCH PULLREQ-URL
```

as an alternative for what we have currently:

```
hub checkout PULLREQ-URL LOCAL-BRANCH
```

But it's not worth worrying about now, especially when I want to rething how this whole command works for the next release.

mislav:
Yep :ship: if it solves the problem; thanks for working on this

commands/checkout_test.go
jingweno: It's to get the string of the error. I changed it to be more concise: `err.Error()`.

commands/checkout.go
jingweno: Yeah, `hub checkout -b LOCAL-BRANCH PULLREQ-URL` makes more sense since it creates a new branch.

> I want to rething how this whole command works for the next release.

Yup!

PR #817, opened 2015-02-22
Remove unnecessary new if statement

RichardLitt:


github/config.go
```diff
@@ -23,9 +23,7 @@ func init() {
 		if u, err := user.Current(); err == nil {
 			homeDir = u.HomeDir
 		}
-	}
-
-	if homeDir == "" {
```
mislav: Why is it unnecessary? Could you provide more rationale?

RichardLitt: There's one right above it? 

```
if y { a }
if y { b } 
```

Can be shortened to `if y { a; b }`, right?

mislav: It can only be shortened that way if `y` never changes in `a` block. During the first conditional block, does `homeDir` has a chance to change?

RichardLitt: Fair. Closed. Good catch, thank you.

PR #824, opened 2015-02-26
Use `.Login` instead of `.Name` for generating a remote or branch name.

knu:
It is `headRepo.Owner.Login` that holds the unix account name of the
owner, and `headRepo.Owner.Name` is omitted from the embedded owner
information, resulting in the following failure:

```
% hub checkout https://github.com/github/hub/issues/{number}
fatal: '' is not a valid remote name
```

### DIFF
--- a/commands/checkout.go
+++ b/commands/checkout.go
@@ -85,7 +85,7 @@ func transformCheckoutArgs(args *Args) error {
 	if headRepo == nil {
 		return fmt.Errorf("Error: that fork is not available anymore")
 	}
-	user := headRepo.Owner.Name
+	user := headRepo.Owner.Login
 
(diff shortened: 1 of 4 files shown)


mislav:
Ouch. Thanks for correcting my oopsie!

This is one demonstration of how tricky it is to ensure compatibility with GitHub API when we're stubbing the API responses. It makes for fast tests, but they are prone to human errors like mine.

knu:
Agreed.  I often feel test code I'm writing would really be testing anything when I just feed what it wants.
Maybe it's an idea to make a test response from a real world snapshot so you can update it from time to time as necessary.

Thanks for the quick merge!

PR #829, opened 2015-03-03
Add new flag to "release" for specifying commitish

### LINKED ISSUE #820: Add "TargetCommitish" argument to "release" command
I think I could do this. Is it okay if I send a pull request for this? Thanks.

darabos:
Resolves #820.

I see this command has no Cucumber tests. I'm too inexperienced to write one I think. I tested the new flag manually: it works fine. Let me know what you think. Thanks!

### DIFF
--- a/commands/release.go
+++ b/commands/release.go
@@ -26,7 +26,7 @@ var (
 	cmdCreateRelease = &Command{
 		Key:   "create",
 		Run:   createRelease,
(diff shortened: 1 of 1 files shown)


mislav:
Thanks for the contribution and sorry for the late reply! I'll review this when I have the time.

mislav:
Thanks! We'll probably change the `release` command interface a little bit (or a whole lot), but this is a good starting point!

PR #834, opened 2015-03-07
Update README.md

prabhanshu1:
standalone edited


mislav:
Sorry for the late reply and thanks for the contribution. New hub versions will ship with an installer script that will take care of this for you automatically!

PR #840, opened 2015-03-11
Document pull-request -o

voidus:


mislav:
We have already documented the option [in the man page](https://github.com/github/hub/blob/26980a9dd79059de4a92b86a1107dbcb297e443a/man/hub.1.ronn#L161), but when you do `hub pull-request --help`, the internal hardcoded help is displayed which you edited here. Thanks, but I believe that we need to fix documentation by consolidating the two help sources, not by trying to keep them up-to-date separately. We'll see what can be done here.

voidus:
Please excuse the drive-by suggestion, but maybe calling `exec man` would be a solution?

mislav:
Yes, it might for non-Windows platforms. But we would also have to split the big man page into man pages for individual commands.

Issue #592, opened 2014-07-12
hub bash completions don't work by default when git's bash completions are loaded dynamically


3flex:
On my system (essentially stock Xubuntu 14.04 with nothing special in my dot files) when I open a new bash shell things are loaded in this order:
* ~/.bashrc which pulls in bash-completion loader
* tbash-completion loader, which pulls in everything in `/etc/bash_completion.d/` (which is where I saved the hub script) and queues up the dynamically loaded scripts in `/usr/share/bash-completion/completions/` (where the git completion file lives)

However - the git bash completion file with the `_git` function isn't loaded until git completion is first invoked. Because the supplied hub completion file checks for the `_git` function when the shell opens and only loads the hub functions if the `_git` function exists it doesn't load the hub completion functions, since `_git` doesn't exist yet.

It's just a matter of forcing the git functions to load before the bash completion loader runs but it took a lot longer than I wanted to figure it all out. A note in the README might help those who are a bit challenged when it comes to bash configuration.

prabhanshu1:
Hi @mislav ,

I am facing the same problem ,even in the latest version of hub (hub-linux-amd64-2.2.0).
hub do not give autocomplete suggestion,if git command has not been used in the terminal before sourcing the hub complition script. I gave too much time to figure out what was happening.

**BTW**, i am a student of computer science ,IIT Kanpur. and am very much excited to do the *project on hub completion* under **Gsoc'15**. Also ,if the issue is not fixed , i am intersted in fixing these kind of bug .

PR #843, opened 2015-03-15
Updated etc/hub.bash_completion.sh : fixed issue #592

prabhanshu1:
fixed issue #592 :by sourcing the /usr/share/bit-completion/completion/git to terminal whenever hub <tab> is pressed, so that hub can use the git completion"
without which hub was not able to complete on pressing <tab>

### DIFF
--- a/etc/hub.bash_completion.sh
+++ b/etc/hub.bash_completion.sh
@@ -1,6 +1,11 @@
 # hub tab-completion script for bash.
 # This script complements the completion script that ships with git.
 
+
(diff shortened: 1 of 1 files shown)


mislav:
Hi, thanks for your attempt to fix this and sorry for the late reply. This only takes care of Ubuntu and probably doesn't cover other bash completion locations if bash lazy completion loading is enabled. Unfortunately, we'll need a more comprehensive solution than this. It would be ideal if we could just tell bash "load completions for `git` if they haven't loaded yet."

Issue #856, opened 2015-04-01
Crash report - Index 1 is out of bound


jcreedcmu:
Error (*errors.errorString): `Index 1 is out of bound`

Stack:

```
goroutine 1 [running]:
github.com/github/hub/commands.(*Args).InsertParam(0xc20805e1b0, 0x1, 0xc2082b9d38, 0x4, 0x4)
	/var/folders/w8/dp0lmdjs2vs51lsxbrf4s49h0000gn/T/go/src/github.com/github/hub/commands/args.go:99 +0x170
github.com/github/hub/commands.transformCheckoutArgs(0xc20805e1b0, 0x0, 0x0)
	/var/folders/w8/dp0lmdjs2vs51lsxbrf4s49h0000gn/T/go/src/github.com/github/hub/commands/checkout.go:102 +0xd63
github.com/github/hub/commands.checkout(0x6876a0, 0xc20805e1b0)
	/var/folders/w8/dp0lmdjs2vs51lsxbrf4s49h0000gn/T/go/src/github.com/github/hub/commands/checkout.go:36 +0x40
github.com/github/hub/commands.(*Command).Call(0x6876a0, 0xc20805e1b0, 0x0, 0x0)
	/var/folders/w8/dp0lmdjs2vs51lsxbrf4s49h0000gn/T/go/src/github.com/github/hub/commands/commands.go:46 +0x189
github.com/github/hub/commands.(*Runner).Call(0xc20802b120, 0x6876a0, 0xc20805e1b0, 0x0, 0x0, 0x0)
	/var/folders/w8/dp0lmdjs2vs51lsxbrf4s49h0000gn/T/go/src/github.com/github/hub/commands/runner.go:89 +0x54
github.com/github/hub/commands.(*Runner).Execute(0xc20802b120, 0x0, 0x0, 0x0)
	/var/folders/w8/dp0lmdjs2vs51lsxbrf4s49h0000gn/T/go/src/github.com/github/hub/commands/runner.go:81 +0x215
main.main()
	/private/tmp/hub-xGkPdT/hub-2.2.0/main.go:12 +0x40

```

Runtime:

```
GOOS: darwin
GOARCH: amd64
```

jcreedcmu:
I was spuriously using the -p flag

jcreedcmu:
in a commandline like

    git checkout -p git://github.com/$USER/$REPO/pull/1234 $BRANCH

mislav:
[...]

PR #861, opened 2015-04-06
Improve crash report with command and version

jingweno:
This adds running command and hub version to crash report. It includes refactoring of hub version to the `version` package for reusability in `commands` package and `github` package.

### DIFF
--- a/github/crash_report.go
+++ b/github/crash_report.go
@@ -12,6 +12,7 @@ import (
 	"github.com/github/hub/git"
 	"github.com/github/hub/ui"
 	"github.com/github/hub/utils"
+	"github.com/github/hub/version"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


jingweno:
An example of the crash report:

screen shot 2015-04-06 at 3 21 57 pm

github/crash_report.go
```diff
@@ -74,19 +75,33 @@ func report(reportedError error, stack string) {
...
 `
 
+func reportTitleAndBody(reportedError error, stack string) (title, body string, err error) {
 	errType := reflect.TypeOf(reportedError).String()
-	message = fmt.Sprintf(message, reportedError, errType, reportedError, stack, runtimeInfo())
+	message := fmt.Sprintf(
+		crashReportTmpl,
+		reportedError,
+		errType,
+		reportedError,
```
mislav: Why does `reportedError` appear twice?

github/crash_report.go
```diff
@@ -74,19 +75,33 @@ func report(reportedError error, stack string) {
...
+		strings.Join(os.Args, " "),
```
mislav: We can't straight-up report used arguments since they might contain sensitive information such as repo/branch names.

jingweno: The report actually won't be created without the user's awareness. It will open an editor so that the users could edit the issue content which they could remove stuff they don't want to report (same workflow like creating a pull request). But I'm open to ways to further obfuscate the content. The intention is to help debug issues like [this](https://github.com/github/hub/issues/856#issuecomment-90263818).

github/crash_report.go
jingweno: One is for issue title, the other is for the body.

github/crash_report.go
mislav: Sure, but we're prefilling it now, and if the user just submits the unchanged text then they're accidentally revealing private information. In fact, the text in this very document, which you haven't changed in this PR, states:

> We're NOT including any information about the command that you were executing, but knowing a little bit more about it would really help us to solve this problem.

jingweno: Alright, let's not push too much to this direction for now. Plus I couldn't think of an easy way to obfuscate the arguments (encrypting the content was one approach came to mind but it's going too far). I removed showing arguments with `--amend`.

version/version.go
```diff
@@ -0,0 +1,16 @@
+package version
```
mislav: Was the new "version" package necessary? Couldn't we have just kept is in one of the existing packages? Since we don't expect that this will grow in functionality.

jingweno: Go doesn't allow circular package dependencies. The logic was previously in `command` and needs to be reused in both `command` and `github`. So it has to be moved to one of the rest of the packages or a new one. I was thinking of a new `hub` package which stores hub specific stuff like main or version. But it requires bigger changes so I made the `version` package for now.

jingweno: Well, the logic could actually go to `github` though, considering `command` depends on `github` but not the other way around.

mislav:
@jingweno Do you still want to ship this? We should wrap it up so it can go in the next bugfix release.

jingweno:
@mislav Yeah, this is definitely good to have. I have rebased on the latest master and it's ready to be merged after the build pass.

mislav:
Thanks!

Issue #599, opened 2014-07-27
Config Issues (host, protocol, ssh...)


KaiSforza:
Hello,

I started using hub as a way to get around using a browser, and I must say that it works great. However, using hub at work has resulted in my finding that there are some configuration options that seem to do one thing but don't do that at all.

### ``hub.protocol``
This can be set to ``https`` to force ``https`` cloning instead of git/ssh. The assumption that I made was that setting it to ``ssh`` would force ``ssh`` connections, and not use ``git://``, however setting ``hub.protocol`` to ``ssh`` does not do this. If there was a way to force all repositories to be thought of as private, that would also work. Having to clone with ``-p`` gets pretty tedious after a while.

### ``hub.host``
Yes, I know that this is only for telling hub what hosts to treat as ``github.com``, but possibly add a ``hub.defaulthost`` that says "Instead of using regular github, default to using ``github.internal.foo`` as the URL. We can do this with exports, but then it is only in the shell environment.

I'm going to look into getting these set up on my own, but I have yet to really work with ruby, and what I make may look like complete and utter crap.

-- 
Thanks,
KaiSforza

janl:
has this landed anywhere, yet? :)

janl:
It appears that at least since 47a9aa9e, private repos automatically get the git@ urls.

My `brew install hub --HEAD` from December didn’t include this, but a current one does, yay :)

KaiSforza:
Any progress on this for non-private repositories?

KaiSforza:
I decided to make progress on it. See #865

KaiSforza:
Fixed in #865

brandonblack:
@KaiSforza thanks for sorting this out!

rjbell4:
What about the default git host part of this?  The question was asked:
> However that would mean that every time you git create or git clone owner/repo, it would use your Enterprise instance by default. You sure you would be OK with that?
My answer in an unqualified "yes".  Our users will be contributing to our GitHub Enterprise instance 99% of the time, and any interaction with github.com will be very rare.  I'd like to be able to config the default host through the Git configuration, instead of through their environment.

rjbell4:
For the part that pertains to the default GitHub host, I have created PR #1184

PR #865, opened 2015-04-10
useHttpProtocol => useProtocol, allow ssh config

KaiSforza:
`useHttpProtocol` has been changed to a more generic `useProtocol`
function that can be used to check more than just https. This allows us to
use the SSH protocol in any location, regardless of the repository being
private or public.

```
% git config --global hub.protocol ssh
% ./hub --noop clone github/hub
git clone git@github.com:github/hub.git
% git config --global hub.protocol https
% ./hub --noop clone github/hub
git clone https://github.com/github/hub.git
% git config --global hub.protocol none
% ./hub --noop clone github/hub
git clone git://github.com/github/hub.git
```

(Got tired of waiting for #599 :) )

### DIFF
--- a/github/project.go
+++ b/github/project.go
@@ -69,9 +69,9 @@ func (p *Project) GitURL(name, owner string, isSSH bool) (url string) {
 
 	host := rawHost(p.Host)
 
-	if useHttpProtocol() {
+	if preferredProtocol() == "https" {
 		url = fmt.Sprintf("https://%s/%s/%s.git", host, owner, name)
-	} else if isSSH {
+	} else if isSSH || preferredProtocol() == "ssh" {
 		url = fmt.Sprintf("git@%s:%s/%s.git", host, owner, name)
 	} else {
 		url = fmt.Sprintf("git://%s/%s/%s.git", host, owner, name)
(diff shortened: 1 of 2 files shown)


KaiSforza:
Added tests in github/

github/project.go
```diff
@@ -92,13 +92,13 @@ func rawHost(host string) string {
 	}
 }
 
-func useHttpProtocol() bool {
-	https := os.Getenv("HUB_PROTOCOL")
-	if https == "" {
-		https, _ = git.Config("hub.protocol")
+func useProtocol(proto string) bool {
```
mislav: Let's call this method `preferredProtocol()`. Then its usage would be:

```go
if "https" == preferredProtocol() {
  # ...
}
```

Simplicity.

github/project.go
```diff
@@ -92,13 +92,13 @@ func rawHost(host string) string {
...
+	cfgproto := os.Getenv("HUB_PROTOCOL")
+	if cfgproto == "" {
+		cfgproto, _ = git.Config("hub.protocol")
```
mislav: I'm not a fan of `cfgproto` variable. Call it something human-readable, such as `protocol` or `userProtocol`

KaiSforza:
Do you want this rebased or just as a separate commit?

mislav:
I like simpler histories, so rebase the hell out of this. There's no reason why this would be more than 1 commit.

KaiSforza:
Squashed and reworded.

mislav:
Thanks. Bonus points for carefully crafting a nice commit message.

However, as someone who always strives for better commit messages, I would suggest that a good commit message doesn't just describe the code change (`oldFunction → newFunction` in your case, clear from the commit diff itself), but it describes what is the functional, desired change. In your case, a better commit message would be "Allow selecting SSH as preferred clone protocol everywhere".

KaiSforza:
I'll keep that in mind. Thanks!

PR #874, opened 2015-04-22
Update README.md

shadyvb:


mislav:
Thanks for the contribution, but I think that the example of how to use hub to clone a repo from GitHub should best clone a repo *unrelated* to itself, to reduce confusion.

PR #875, opened 2015-04-22
fix non-portable GNU find(1)-ism

sl4mmy:
GNU find(1)'s -not operator is not POSIX compliant.  Use ! operator
instead in order to work with other find(1) implementations.


mislav:
Thanks!

PR #905, opened 2015-05-21
Use `--no-tags` for new remotes in `checkout/cherry-pick`

### LINKED ISSUE #879: hub checkout pulls down tags from user's fork
If the user has tags in the history on their fork, doing a `hub checkout <pull req URL>` will also pull down their tags.  Would it be possible to add the `--no-tags` option to the `fetch` call to prevent that?

mislav:
This is so the tags from the automatically-added remote (typically someone's fork) don't spill in and mix with the current tags.

Fixes #879

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,11 +124,11 @@ superpowers:
 ### git cherry-pick
 
     $ git cherry-pick http://github.com/mislav/REPO/commit/SHA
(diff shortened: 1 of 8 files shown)


jingweno:
:+1:

PR #916, opened 2015-06-02
Add OSX build to Travis

jingweno:
Adding OSX build to Travis. Build time has increased 4 times. But it helps reduce cross platform issues.

screen shot 2015-06-02 at 6 07 12 am

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+os:
+  - linux
+  - osx
 sudo: false
 before_install:
   - export BUNDLE_GEMFILE=$PWD/Gemfile
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,8 @@ cache:
     - vendor/bundle
 notifications:
   email: false
+env:
+  global:
+    - AMAZON_S3_BUCKET=ci-cache
+    - AMAZON_ACCESS_KEY_ID=AKIAJQCVTDEWQHRPBPGQ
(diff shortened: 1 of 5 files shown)


mislav:
Seems that osx build doesn't have tmux for testing (understandably, because I only pre-compiled it for Linux) and that Travis caching doesn't work for some reason. Trying to solve both issues right now.

jingweno:
> Seems that osx build doesn't have tmux for testing (understandably, because I only pre-compiled it for Linux) and that Travis caching doesn't work for some reason.

:+1:

mislav:
For the record, how I created a tarball of precompiled tmux for Travis's user directory:

```sh
prefix=/Users/travis
sudo mkdir $prefix
sudo chown "$USER":staff $prefix

unset LDFLAGS
export PATH=/usr/bin:/bin:/usr/sbin:/sbin

cd /tmp
wget https://downloads.sourceforge.net/project/levent/libevent/libevent-2.0/libevent-2.0.22-stable.tar.gz  | tar -xz
cd libevent-2.0.22-stable
./configure -disable-dependency-tracking --disable-debug-mode --prefix=$prefix
make && make install

cd /tmp
wget https://downloads.sourceforge.net/project/tmux/tmux/tmux-2.0/tmux-2.0.tar.gz | tar -xz
cd tmux-2.0
LDFLAGS="-lresolv" LIBEVENT_LIBS="-L${prefix}/lib -levent" LIBEVENT_CFLAGS="-I${prefix}/include" ./configure --prefix=$prefix --disable-dependency-tracking
make && make install

cd $prefix
tar -czf tmux.osx.tgz bin lib share
```

jingweno:
I assume we could also "brew install tmux" on OSX. But your approach will make the build faster by downloading pre-built binaries.

mislav:
Yes, `brew install tmux` worked but it sometimes took 20 seconds because of openssl dependency. Now it's almost instantaneous.

PR #919, opened 2015-06-03
05/25/2015 All setup done. First commit. `hub.fish` file created.

hasit:
05/26/2015
2 support functions: __fish_hub_needs_command, __fish_hub_using_command
Tab completions: help, fork [--no-remote]


mislav:
Thanks for the update, @hasit! However, we won't pull this into `github/hub` yet because I'd like the script to mature more before merging into master branch here. For instance, I would like to see it wrapping the official git completion script in fish and expanding it to include hub commands and extra flags. Let's continue the work in your fork until then.

PR #922, opened 2015-06-07
Automatically publish a GitHub release from Travis CI

mislav:
Whenever CI processes a new git tag successfully, automatically create a GitHub release and upload cross-compiled assets to it.

The new release body is populated by the git changelog of Go runtime files since the last release. The new release is created as a draft so that we have a chance to review its text and assets before going public.

This reduces the whole release process to simply pushing a git tag to GitHub.


mislav:
@jingweno From my rudimentary tests, cross-compiling seems to produce valid binaries for Windows, although we'll have to test on both 32-bit and 64-bit before being sure.

However, FreeBSD doesn't compile due to:

```
Godeps/_workspace/src/github.com/howeyc/gopass/pass.go:18: undefined: getch
```

You might know what's going on there.

We didn't offer prebuilt binaries specifically for FreeBSD but I decided to add them to our matrix as an experiment. Because they fail, their assets are never getting uploaded to the release, but that doesn't get block other assets from getting uploaded. So, this still publishes binaries for OS X, Linux, and Windows even though FreeBSD build fails.

script/package
```diff
@@ -1,166 +1,55 @@
...
+
+exename="${tmpdir}/bin/hub"
+[ "$os" != "windows" ] || exename="${exename}.exe"
+mkdir -p "${exename%/*}"
+script/build -o "$exename"
+
+if [ "$os" = "windows" ]; then
+  cp README.md "${tmpdir}/README.txt"
+  cp LICENSE "${tmpdir}/LICENSE.txt"
+  cp man/hub.1.html "${tmpdir}/hub.html"
```
mislav: For Windows, package docs as `.txt` so they're easily opened by Notepad. Also, copy the man page over as HTML representation since Windows doesn't have `man`.

script/cross-compile
```diff
@@ -0,0 +1,30 @@
...
+version="${1?}"
+
+echo '
+  darwin   amd64    OS X
+  freebsd  386      FreeBSD 32-bit
+  freebsd  amd64    FreeBSD 64-bit
+  linux    386      Linux 32-bit
+  linux    amd64    Linux 64-bit
+  windows  386      Windows 32-bit
+  windows  amd64    Windows 64-bit
```
mislav: Such cross-compile. Very build matrix.

script/github-release
```diff
@@ -0,0 +1,66 @@
...
+STDIN.each do |line|
+  filename, label = line.chomp.split("\t", 2)
+  name = File.basename filename
+
+  if asset = release["assets"].find { |a| a.fetch("name") == name }
+    api asset.fetch("url"), "-X", "DELETE"
+  end
+
+  upload = upload_url.sub(/\{\?.+?\}/, "?name=#{escape name}&label=#{escape label}")
+  api upload, "-H", "Content-Type: application/octet-stream", "--data-binary", "@#{filename}"
```
mislav: The magic of auto-uploaded assets happens here :sparkles:

script/package
```diff
@@ -1,166 +1,55 @@
...
+export GIMME_ARCH="$arch"
+eval "$(gimme "$TRAVIS_GO_VERSION")" >/dev/null
+
+tmpdir="tmp/${release}"
+rm -rf "$tmpdir"
+
+exename="${tmpdir}/bin/hub"
+[ "$os" != "windows" ] || exename="${exename}.exe"
+mkdir -p "${exename%/*}"
+script/build -o "$exename"
```
jingweno: How would `script/build` handle cross compile? To cross compile, I used to use [gox](https://github.com/mitchellh/gox). My understanding is `GOOS` and `GOARCH` need to be set. Also the tool chain needs to be built to be ready for cross compilation (`gox -build-toolchain`). There is a [bash alternative](https://github.com/davecheney/golang-crosscompile/blob/master/crosscompile.bash) to gox.

script/github-release
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env ruby
```
jingweno: It would be nice to use `hub release create` in the future :wink: - I'm actually using it to upload all our releases.

script/package
jingweno: How about for Windows shell environments that has `man`, e.g., Cygwin?

jingweno:
@mislav:

> FreeBSD doesn't compile due to
> `Godeps/_workspace/src/github.com/howeyc/gopass/pass.go:18: undefined: getch`

Hmm...I'm actually not sure how it is failed. The build tag for FreeBSD is [there](https://github.com/github/hub/blob/master/Godeps/_workspace/src/github.com/howeyc/gopass/bsd.go#L1) and there's [such method defined](https://github.com/github/hub/blob/master/Godeps/_workspace/src/github.com/howeyc/gopass/bsd.go#L27-L29). I'm suspecting it's because [`GOOS` and `GOARCH` are not set correctly](https://github.com/github/hub/pull/922#discussion_r31873309) but I could be wrong.

hkdobrev:
Why don't you use the Travis CI deployment feature with GitHub releases as a provider?  http://docs.travis-ci.com/user/deployment/releases/.

You can use the `before_deploy` hook to prepare the files you want for the release.

script/package
mislav: I don't care much about Cygwin. It's not a proper unix environment and things don't often work in it. But I can look into where it looks up man pages.

script/package
mislav: I'm settting GIMME_OS and GIMME_ARCH above for Gimme that is preinstalled on Travis. It automatically sets up the runtime for the desired OS and takes care of setting GOOS and GOARCH properly.

script/github-release
mislav: It would be, after it stabilizes in its API a bit. For now I've coded it up manually since I wanted to possibly reuse this script in other Go projects that might not have hub available— although, the script itself could download and install hub just for the purposes of creating a release :)

mislav:
> The build tag for FreeBSD is there and there's such method defined.

I'm pretty sure that GOOS and GOARCH were set correctly because Gimme did a good job for OSs other than FreeBSD. Could you try to compile on your machine using gox? Before we didn't specifically create binaries for FreeBSD, so it's possible that this never worked for some reason.

mislav:
> Why don't you use the Travis CI deployment feature with GitHub releases as a provider?

@hkdobrev Good point, I didn't know about that feature. I would say that the advantage of my script is that it creates a new release text with the changelog since last release, and uploads assets with pretty labels so that they are human-readable.

screen shot 2015-06-07 at 1 58 13 pm

script/github-release
mislav: For instance, something that hub doesn't let me do right now is upload the compiled asset even if it exists. My script detects the duplicate and deletes the old one under the same filename. This is so botched binaries might get updated.

script/package
jingweno: :+1:

script/github-release
jingweno: > For instance, something that hub doesn't let me do right now is upload the compiled asset even if it exists.

This is valid point. It would be nice to support this in the future.

script/package
jingweno: I was just thinking there's no harm to leave the man page there. [Same argument](https://github.com/github/hub/pull/751#issuecomment-68610283) was made for the autocomplete scripts. However, we could ship as is and put it back if people ask for it.

jingweno:
:shipit:

script/cross-compile
jingweno: I will try out freebsd and fix in another PR if `gopass` is failing the builds

script/package
mislav: You're right, it doesn't hurt to leave them in. I forgot about win-bash being a thing. I'll make the change.

script/cross-compile
mislav: I'll test but I don't think the recent gopass bump will fix the issue, since it only contained fixes for Windows. I have no idea why it doesn't compile.

jingweno:
@mislav Are we missing anything from here? Can't wait to see this going.

mislav:
This is largely done, but I still want to get a Windows `.bat` installer in here, as well as include the man page for Windows as well per your comment.

mislav:
@jingweno I have updated to Go 1.5.1 which has native cross-compile support and verified that it works, at least for Windows build. I have also added installer scripts for Windows and *nix platforms. This is good to go! We can test it out fix a bugfix release next.

technoweenie:
We had all kinds of problems with that batch file. Git LFS currently has an NSIS installer: https://github.com/github/git-lfs/blob/master/script/nsis/installer.nsi

When I make a new Git LFS release, I'll document the required steps again. The only hard part is it requires Windows :)

mislav:
> We had all kinds of problems with that batch file.

Yes, my main problem was its `git.exe` detection. If run from the GitHub Shell, it wouldn't properly detect that git was present after gaining sudo privileges. I've removed that detection to simplify the script. I've also added a `pause` in the end so that the Command Prompt window stays open after finishing installation so that you get a confirmation of successful install.

I've seen the NSIS installer and might adopt that approach as well, but for now I might keep it simple just to avoid introducing a GUI program to install a CLI tool.

jingweno:
:shipit:

script/build
```diff
@@ -41,7 +41,7 @@ up_to_date() {
 
 build_hub() {
   setup_gopath
-  [ -n "$1" ] && (up_to_date "$1" || go build -ldflags "-X github.com/github/hub/commands.Version `./script/version`" -o "$1")
+  [ -n "$1" ] && (up_to_date "$1" || go build -ldflags "-X github.com/github/hub/commands.Version=`./script/version`" -o "$1")
```
jingweno: Note that potential conflict with this PR: https://github.com/github/hub/pull/861

Issue #446, opened 2013-12-09
Templatize pull-request message/title


ajf:
It would be nice to be able to specify a `-t` argument to `pull-request` that takes a file name.  Hub could then use this file to pre-populate the pull-request message in the case of 1 or more commits.

When making pull-requests, we have a few standard formats that we'd like people to use, and this would make it easier for them.

rtuin:
I know this is an old issue, but having the option to configure a template would improve the workflow when creating pull-requests.
Judging by the current code there is no option for this yet. Correct? Is this on the roadmap, or can I help, perhaps?

mislav:
Internally we do use a template I think. We would just need to make this configurable so that you can specify a file to read the template from.

rjmunro:
`git commit` uses `-F` to specify the filename and `-e` or `--edit` to say it should be edited, which achieves the same result. It would be better to support the same options.

PR #924, opened 2015-06-08
Add -e option to pull request

m-nakamura145:
Hi. ```-e``` option improves pull-request with template.

For example 
run ```pull-request -F template_file -e``` then open editor and configure a template file. :tada: 

related issue: https://github.com/github/hub/issues/446

### DIFF
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -43,7 +43,8 @@ var (
 	flagPullRequestAssignee,
 	flagPullRequestFile string
 	flagPullRequestBrowse,
-	flagPullRequestForce bool
+	flagPullRequestForce,
+	flagPullRequestEdit bool
 )
 
 func init() {
@@ -55,6 +56,7 @@ func init() {
 	cmdPullRequest.Flag.BoolVarP(&flagPullRequestForce, "force", "f", false, "FORCE")
 	cmdPullRequest.Flag.StringVarP(&flagPullRequestFile, "file", "F", "", "FILE")
 	cmdPullRequest.Flag.StringVarP(&flagPullRequestAssignee, "assign", "a", "", "USER")
+	cmdPullRequest.Flag.BoolVarP(&flagPullRequestEdit, "edit", "e", false, "EDIT")
 
 	CmdRunner.Use(cmdPullRequest)
 }
@@ -180,6 +182,14 @@ func pullRequest(cmd *Command, args *Args) {
 		utils.Check(err)
 
 		editor, err = github.NewEditor("PULLREQ", "pull request", message)
+		utils.Check(err)
+
+		title, body, err = editor.EditTitleAndBody()
+		utils.Check(err)
+	}
+
+	if flagPullRequestFile != "" && flagPullRequestEdit {
+		editor, err = github.NewEditor("PULLREQ", "pull request", title + "\n\n" + body)
 		utils.Check(err)
 
 		title, body, err = editor.EditTitleAndBody()
--- a/features/pull_request.feature
+++ b/features/pull_request.feature
@@ -279,6 +279,35 @@ Feature: hub pull-request
     Then the output should contain exactly "https://github.com/mislav/coral/pull/12\n"
     And the file ".git/PULLREQ_EDITMSG" should not exist
 
+  Scenario: Edit title and body from file
+    Given the GitHub API server:
+      """
+      post('/repos/mislav/coral/pulls') {
+        assert :title => 'Edit title from file',
+               :body  => "Edit body from file as well.\n\nMultiline, even!"
+        json :html_url => "https://github.com/mislav/coral/pull/12"
+      }
+      """
+    And a file named "pullreq-msg" with:
+      """
+      Title from file
+
+      Body from file as well.
+
+      Multiline, even!
+      """
+    When I run `hub pull-request -F pullreq-msg --edit` interactively
+    And I pass in:
+      """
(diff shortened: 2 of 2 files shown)


commands/pull_request.go
```diff
@@ -55,6 +56,7 @@ func init() {
 	cmdPullRequest.Flag.BoolVarP(&flagPullRequestForce, "force", "f", false, "FORCE")
 	cmdPullRequest.Flag.StringVarP(&flagPullRequestFile, "file", "F", "", "FILE")
 	cmdPullRequest.Flag.StringVarP(&flagPullRequestAssignee, "assign", "a", "", "USER")
+	cmdPullRequest.Flag.BoolVarP(&flagPullRequestEditor, "editor", "e", false, "EDITOR")
```
mislav: For parity with `git commit -e`, I think the long name for this flag should be `--edit`, not `--editor`.

commands/pull_request.go
```diff
@@ -186,6 +188,14 @@ func pullRequest(cmd *Command, args *Args) {
 		utils.Check(err)
 	}
 
+	if flagPullRequestEditor {
+		editor, err = github.NewEditor("PULLREQ", "pull request", title + "\n\n" + body)
```
mislav: Why does this repeat the above logic? Couldn't we just have the above conditional branch happen unconditionally if `--edit` was used?

mislav:
Thanks! This will need a test in [`features/pull_request.feature`](https://github.com/github/hub/blob/d45511c7aaca000c49ac1a2ef1a162a6153b0c44/features/pull_request.feature) to get accepted. Look for existing tests using `-F` or "text editor".

commands/pull_request.go
m-nakamura145: fd09d23
Fix!! :bow:

commands/pull_request.go
m-nakamura145: fd09d23
Oh...sorry, I want to open editor when template file exits. Fix! :bow:

m-nakamura145:
fd09d23
Thanks!! I added test to ``` features/pull_request.feature``` :+1:

mislav:
Thanks for updating! This is getting really close to being done. However, I'm not satisfied with the test completely. We're not verifying in any way that the template was seen by the user and offered for editing. However, setting up a proper test is tricky and I won't blame you if I don't want to deal with that right now. When I catch some time I'll try to work on top of your code to improve the test.

michaeljs1990:
Sorry for jumping in late! this is an awesome addition but to me it seems like this provides a much different feature then -F. It would be preferable to have a completely different flag for this IMO. -F to me is when you have already completely prepared a message where if would like to use a template It would seem like you should be able to do something like -T instead of having to set -F and then -e. It is also just odd to me because you very rarely see command line flags that manipulate another command line flag in a way like this and i think would lead to some confusion among end users.

michaeljs1990:
@m-nakamura145 this is a feature I would very much like are you still working on it? I don't want it to sit in PR purgatory :cry:  so let me know if you are no longer invested in getting this up to current standards and I will take over.

m-nakamura145:
@michaeljs1990 I'm sorry for beging late. :pray:  I will do this PR until the last. :runner:

michaeljs1990:
@m-nakamura145 you have any idea when you will get to make the changes requested above?

mislav:
Do we still want this feature in? :wink:

m-nakamura145:
I want this feature but couldn't write test code. sorry :bow: :weary:

mislav:
@m-nakamura145 That's OK! I'll see what I can do.

michaeljs1990:
I would still very much like this feature :+1:

dhilton:
OK - I'm just looking at the tests - getting 7 failures that are unrelated to this PR, so ignoring for now, assuming they'll go away when rebased against current master.

* To summarise, we need to change the flag from '-f' to '-T'
* The PR needs a working cucumber test to cover the rendering of a commit message given a template

Does it need a golang written test?

mislav:
The most useful tests for this should be added to `feature/pull_request.feature`.

Why changing `-F` to `-T`?

mironal:
:+1:

halkeye:
I see the tests are passing, does it need more tests? Can I help in any way to get this landed?

keyan:
:+1: would really like to see this too, happy to help if help is needed. Looks like a rebase is all it needs though.

Tonkpils:
Is this still valid with https://github.com/github/hub/pull/1187 ?

dhilton:
No I think this has been dealt with thanks!

dcosson:
I'm not sure that this is fully addressed by templates in #1187. With the -F flag, I can use bash process substitution to dynamically generate the PR description based on some script (e.g. as a simple example `hub pull-request -F <(git log --format=%B origin/master..HEAD)`). I wish I could do this as a dynamic default but have the content open in an editor before creating the PR.

Can I accomplish this with the templates?

bsander:
@dcosson probably not, though I've been using a short bash script for a while to accomplish this feature using a tmpfile. You should be able to do it with something like this:

```shell
PR_MSG=$(mktemp --suffix=".md" -t PR_MSG.XXXXXX )
git log --format=%B origin/master..HEAD > "$PR_MSG"
$EDITOR -w "$PR_MSG"
hub pull-request -F "$PR_MSG" 
rm -f "$PR_MSG"
```

Issue #837, opened 2015-03-08
Have to reenter credentials on Windows


twsouthwick:
I cloned a repo on Win10 64-bit and it worked fine.  However, when I tried removing a remote branch (git push origin :[branch name]) it asked me for my credentials again.

My version:
git version 1.9.5.msysgit.0
hub version 2.2.0

Nicofuma:
I noticed the same thing and in my case it seems to be due to a '|' in the password: hub stop aking for my password as soon as I type the '|' character

mislav:
We're having some trouble on Windows, although we're not sure why: #815

From what I can tell, the password prompt is definitely busted. However, @twsouthwick said that he managed to input the password and it worked, but it wasn't persisted. The next time that happens, can you check that the successful password prompt did persist the auth credentials in `$HOME/.config/hub`? (I'm not sure what is the user's home directory on Windows.)

Also, please try with latest hub version v2.2.1, thanks!

/cc @jingweno

jingweno:
@Nicofuma I was able to type "|" on a 32 bit Windows with v2.2.1. Let me try on 64 bit as well. I'm using the Git shell comes with [GitHub Windows](https://windows.github.com/). What's your terminal?

We could try upgrading the [gopass](https://github.com/howeyc/gopass) lib and see if it fixes the issues. We're missing 2 latest Windows fixes.

Nicofuma:
@jingweno I will update hub and try tonight to see if I can reproduce the issue.

Nicofuma:
@jingweno I just reproduced it and actually it's not related to a specific char but to the password length: the prompt stop just after the 15th. So if your password has more than 15 chars you're stuck.

Reproduced with `git bash` and `powershell` (with powershell it even stops after the 13th)

mislav:
I was able to reproduce the problem in Windows 8.1 Powershell with hub 2.2.1:

```
C:\Users\IEUser\Documents> .\hub clone playground
github.com username: mislav
github.com password for mislav (never stored):
←[36m> POST https://api.github.com/authorizations←[0m
←[36m> Authorization: Basic [REDACTED]←[0m
←[36m{"scopes":["repo"],"note":"hub for IE11Win8_1\\IEUser@IE11WIN8_1","note_url":"http://hub.github.com/"}
←[0m
←[36m< HTTP 401←[0m
←[36m{"message":"Bad credentials","documentation_url":"https://developer.github.com/v3"}←[0m
[...]

PR #925, opened 2015-06-09
Fix password insertion more than 15 characters on Windows

jingweno:
Bumping to the latest gopass. This fixes https://github.com/github/hub/issues/837.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -26,7 +26,7 @@
 		},
 		{
 			"ImportPath": "github.com/howeyc/gopass",
-			"Rev": "62ab5a80502a82291f265e6980d72310b8f480d5"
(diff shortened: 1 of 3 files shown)


mislav:
Nice! Have you tested on Windows?

I restarted the intermittent failing build, btw

jingweno:
@mislav Yes, I tested on Windows and it works.

mislav:
Awesome!

I must say I'm looking at the diff of the fix and I don't understand.

jingweno:
>  I must say I'm looking at the diff of the fix and I don't understand.

I should have put more description. [This commit](https://github.com/howeyc/gopass/commit/2c70fa70727c953c51695f800f25d6b44abb368e) fixes NUL (0 byte) returning from getch on 64 bit Windows.

PR #927, opened 2015-06-12
Support coloring console on Windows

jingweno:
As pointed in https://github.com/github/hub/issues/837#issuecomment-108581171, `hub` doesn't display color correctly on Windows. I introduce `github.com/mattn/go-colorable` to solve the issue. But `bitbucket.org/kardianos/osext` (dependency of `github.com/inconshreveable/go-update`) was removed and it prevented me from adding `github.com/mattn/go-colorable` with `godep`. So I upgraded `github.com/inconshreveable/go-update` with the new location to `kardianos/osext`. Nothing should be broken.


jingweno:
Color in action on Windows:

screen shot 2015-06-12 at 1 00 39 pm

mislav:
Nice!

PR #788, opened 2015-02-01
WIP: Improve and simplify zsh completion, using `user-commands` zstyle

blueyed:
Defining the hub commands via the user-commands zstyle makes it less
fragile and more straightforward.

This also drops support for the bash-completion from the git
distribution.

PR #933, opened 2015-06-26
All supported shells added to hub.zsh_completion

hasit:
Added all supported shells (bash zsh sh ksh csh tcsh fish) to keep it consistent with commands/alias.go

### DIFF
--- a/etc/_hub
+++ b/etc/_hub
@@ -1,0 +1,1 @@
+hub.zsh_completion
--- a/etc/hub.zsh_completion
+++ b/etc/hub.zsh_completion
@@ -9,6 +9,39 @@
 # This gets also done for `alias git=hub` by default, and can be forced using
 # `compdef _hub git`.
 
+# INSTALLATION DETAILS: {{{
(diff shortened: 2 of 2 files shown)


mislav:
Sorry, we can't advertise support for `sh`, `tcsh`, `fish` before we tested them! For instance, we don't yet have fish completion. Also, I feel current completion script for bash wouldn't run in `sh` because it wouldn't have all the bash features needed. I don't think many people are using `sh` for their login shell anyway.

I will work on merging #788 instead.

hasit:
@mislav This PR is for aliasing hub as git. commands/alias.go supports alias script for all those shells.

mislav:
Ah, correct; thanks for clarifying.

mislav:
I have decided that this is not necessary; i.e. correct tab completion for `hub alias` is not important since `hub alias` is a semi-internal command that serves just for hub configuration and users will not interact with it daily. Plus, I am not ready to pull in the additional commits in this PR that were adopted from another PR.

PR #942, opened 2015-07-03
Add support for the `--literal-pathspecs` flag

whilp:
Extend support for other global flags as introduced in 2699a6b718ce8af11cb4f4dc7d069cf09891ff71. `--literal-pathspecs` is fairly rare, but /is/ used by the Emacs porcelain Magit. Without this patch, `hub` falls back to the `git` executable when called via Magit, preventing the use of extended commands like `pull-request` in that context.

### DIFF
--- a/commands/args.go
+++ b/commands/args.go
@@ -201,6 +201,9 @@ func slurpGlobalFlags(args *[]string, globalFlags *[]string, noop *bool) {
 		noReplaceObjects bool
 		bare             bool
 		literalPathspecs bool
+		globPathspecs    bool
+		noglobPathspecs  bool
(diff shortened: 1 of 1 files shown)


mislav:
Thanks. Could you add all other `*-pathspecs` options as well?

whilp:
@mislav sure thing. I pushed the dumbest possible approach in b1b2b3d.

mislav:
Looks good, but I didn't like our general approach of hardcoding all these flags. I made an alternate PR that solves your problem and that I pinged you on for review. Thanks

whilp:
Happy to close this in favor of https://github.com/github/hub/pull/951

PR #946, opened 2015-07-07
Fix spurious test failure when GIT_EDITOR is set

whilp:
`git var` checks the `$GIT_EDITOR` environment variable before the `core.editor` config knob. `TestGitDir` intends to verify that the latter is respected, so sanitize the environment before running the test and restore it upon completion.

### DIFF
--- a/git/git_test.go
+++ b/git/git_test.go
@@ -1,6 +1,7 @@
 package git
 
 import (
+	"os"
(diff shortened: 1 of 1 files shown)


parkr:
:+1:

mislav:
Thanks!

PR #947, opened 2015-07-07
Update EDITMSG copy

justincampbell:
* Put "The" before description
* Add comma
* Reflow to 80 columns

### DIFF
--- a/commands/issue.go
+++ b/commands/issue.go
@@ -95,8 +95,8 @@ func writeIssueTitleAndBody(project *github.Project) (string, string, error) {
 	message := `
 # Creating issue for %s.
 #
-# Write a message for this issue. The first block
-# of text is the title and the rest is description.
+# Write a message for this issue. The first block of
+# text is the title and the rest is the description.
 `
 	message = fmt.Sprintf(message, project.Name)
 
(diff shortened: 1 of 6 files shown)


mislav:
Sorry, I don't see a value in this.

* The reflow doesn't affect users much. This text gets ignored anyway.
* The definite article before "description" doesn't sound like it's stricly necessary.
* A comma before "and" is not necessary unless it's an Oxford comma.

justincampbell:
"and the rest is description" is not proper English. Happy to open a PR with just the "the" before description.

mislav:
Sure, thanks. You should re-use this same PR.

justincampbell:
Updated. I did reflow a tiny bit just so that the second line wasn't longer than the first.

mislav:
The last bit you need to do is update the test suite so it reflects the change. Right now there is a test failure.

justincampbell:
Weird, script/test passes locally for me. Can you re-run?

-Justin

github.com/justincampbell
twitter.com/justincampbell

On July 8, 2015 at 12:14:42 PM, Mislav Marohnić (notifications@github.com) wrote:

The last bit you need to do is update the test suite so it reflects the change. Right now there is a test failure.

mislav:
Does this pass for you? `bundle exec cucumber features/pull_request.feature:115`

justincampbell:
Ah I see, updated!

mislav:
Thanks for being patient.

PR #951, opened 2015-07-08
Avoid depending on a hardcoded list of git global flags

mislav:
Now, every argument except `--help` and `--version` that starts with a `-` is considered to be a global flag. In the long run, this ensures that hub deals with global flags that git might add in the future.

Alternative to https://github.com/github/hub/pull/942

/cc @whilp

### DIFF
--- a/commands/args.go
+++ b/commands/args.go
@@ -169,7 +169,7 @@ func NewArgs(args []string) *Args {
 	)
 
 	slurpGlobalFlags(&args, &globalFlags)
-	noop = removeValue(&globalFlags, "--noop")
+	noop = removeValue(&globalFlags, noopFlag)
 
 	if len(args) == 0 {
 		params = []string{}
@@ -190,6 +190,7 @@ func NewArgs(args []string) *Args {
 }
 
 const (
+	noopFlag    = "--noop"
 	versionFlag = "--version"
 	helpFlag    = "--help"
 	configFlag  = "-c"
@@ -238,14 +239,13 @@ func removeItem(slice []string, index int) (newSlice []string, item string) {
 }
 
 func removeValue(slice *[]string, value string) (found bool) {
-	var newSlice []string
-	for _, item := range *slice {
-		if item == value {
+	aa := *slice
+	for i := len(aa) - 1; i >= 0; i-- {
+		arg := aa[i]
+		if arg == value {
 			found = true
-		} else {
-			newSlice = append(newSlice, item)
+			*slice, _ = removeItem(*slice, i)
 		}
 	}
-	*slice = newSlice
 	return found
 }
--- a/commands/args.go
+++ b/commands/args.go
@@ -168,7 +168,8 @@ func NewArgs(args []string) *Args {
 		globalFlags []string
 	)
 
-	slurpGlobalFlags(&args, &globalFlags, &noop)
+	slurpGlobalFlags(&args, &globalFlags)
+	noop = removeValue(&globalFlags, "--noop")
 
 	if len(args) == 0 {
 		params = []string{}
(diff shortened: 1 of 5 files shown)


whilp:
Love it! I noticed your similar previous proposal in https://github.com/github/hub/pull/802 to solve https://github.com/github/hub/pull/792 (plus related in https://github.com/github/hub/pull/793 and https://github.com/github/hub/pull/764); it does feel like it's worth the refactor to avoid future breakage and reduce maintenance. I also like the new tests. Seems legit to me. Thank you!

mislav:
@jingweno What do you think?

parkr:
:+1:

commands/args.go
```diff
@@ -191,90 +189,29 @@ func NewArgs(args []string) *Args {
...
+	slurpNextValue := false
+	commandIndex := 0
 
-	if gitDir != "" {
-		*globalFlags = append(*globalFlags, "--git-dir", gitDir)
+	for i, arg := range *args {
+		if slurpNextValue {
+			commandIndex = i + 1
+			slurpNextValue = false
+		} else if arg == "--version" || arg == "--help" || !strings.HasPrefix(arg, "-") {
```
jingweno: `"--version"` and `"--help"` could be a `const`

commands/args.go
```diff
@@ -191,90 +189,29 @@ func NewArgs(args []string) *Args {
...
+				slurpNextValue = true
+			}
+		}
 	}
 
-	if workTree != "" {
-		*globalFlags = append(*globalFlags, "--work-tree", workTree)
+	if commandIndex > 0 {
+		aa := *args
+		*globalFlags = aa[0:commandIndex]
```
jingweno: I wonder what happens if there're multiple `-c` and they're not consequent to each other, e.g., `hub -c a=b --git-dir=X -c c=d`. It doesn't seem to work. Could you add a test for that?

commands/runner.go
```diff
@@ -54,8 +54,11 @@ func (r *Runner) All() map[string]*Command {
 	return r.commands
 }
 
-func (r *Runner) Use(command *Command) {
+func (r *Runner) Use(command *Command, aliases ...string) {
```
jingweno: How about adding a alias field to `Command` which would make it more modular:

```go
type Command struct {
   Aliases []string
}
```

In `Use`, we could add all aliases:

```go
for _, a := range command.Aliases {
   r.commands[a] = command
}
```

commands/args.go
```diff
@@ -171,7 +168,8 @@ func NewArgs(args []string) *Args {
 		globalFlags []string
 	)
 
-	slurpGlobalFlags(&args, &globalFlags, &noop)
+	slurpGlobalFlags(&args, &globalFlags)
+	noop = removeValue(&globalFlags, "--noop")
```
jingweno: I think `removeValue` could be in `slurpGlobalFlags`'s for loop as well by iterating with index only (no `range`):

```go
for i := len(*args) - 1; i >= 0; i-- {
  arg := *args[i] 
  // other ifs
  if arg == "--noop" {
    *args, _ = removeItem(*args, i) // note that we may not need removeValue since removeItem does the job here.
  }
}
```

What do you think?

jingweno:
@mislav In general I think your approach is cleaner than mine :+1:

commands/runner.go
mislav: I went for the simplest possible solution right now. I don't think we'll use aliases much going forward— these flags-that-are-actually-subcommands is sorta like an edge case. So instead of making true alias support, I'll just stick with the current solution and see whether there will be any need for a more sophisticated system in the future.

commands/args.go
mislav: Good idea. I've reversed the loop to handle cases where `--noop` might have been specified multiple times (an edge case, but we don't want the program to segfault).

commands/args.go
mislav: Multiple `-c`s were already handled well in a loop, but I added a test as well to prove it.

mislav:
@jingweno Thanks for the thorough review!

jingweno:
:shipit:

Issue #948, opened 2015-07-08
Use **GITHUB_TOKEN** environment variable to save the token


stevemao:
This module looks for **GITHUB_USER** and **GITHUB_PASSWORD** environment variables and get a token then store in "~/.config/hub". Why not just look for **GITHUB_TOKEN** environment variable straight away? If not found try to fallback to use username and password to get the token then store in **GITHUB_TOKEN**?

mislav:
Planned! https://github.com/github/hub/issues/463#issuecomment-32387755

Thanks for the suggestion.

PR #953, opened 2015-07-09
config: if GITHUB_TOKEN is set, do not attempt to create one

### LINKED ISSUE #463: Obtaining authentication from environment
I don't want to have a `.config/hub` file lying around on remote systems where I use hub. I'd prefer to communicate my password through the environment (since I can use ssh SendEnv, etc).

I've tried using GITHUB_USER and GITHUB_PASSWORD with a token generated from https://github.com/settings/applications, but when I try to send a pull request I get "404 not found". It works if I send a pull request by entering my credentials and using `.config/hub`. Any ideas?

parkr:
Fixes #463 and #948.

I looked for tests, but couldn't find any for config.

/cc @jingweno @mislav

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,6 +286,9 @@ the API and exchange it for an OAuth token, which it saves in "~/.config/hub".
 To avoid being prompted, use **GITHUB_USER** and **GITHUB_PASSWORD** environment
 variables.
 
+Alternatively, you may provide **GITHUB_TOKEN**, an access token with
+**repo** permissions. This will not be written to `~/.config/hub`.
+
 ### HTTPS instead of git protocol
 
 If you prefer the HTTPS protocol for GitHub repositories, you can set
--- a/features/authentication.feature
+++ b/features/authentication.feature
@@ -117,6 +117,25 @@ Feature: OAuth authentication
     Then the output should not contain "github.com password for mislav"
     And the file "../home/.config/hub" should contain "oauth_token: OTOKEN"
 
+  Scenario: Credentials from GITHUB_TOKEN
+    Given the GitHub API server:
+      """
+      get('/user') {
+        halt 401 unless request.env["HTTP_AUTHORIZATION"] == "token OTOKEN"
+        json :login => 'mislav'
+      }
+      post('/user/repos') {
+        halt 401 unless request.env["HTTP_AUTHORIZATION"] == "token OTOKEN"
+        p request.env
+        json :full_name => 'mislav/dotfiles'
+      }
+      """
+    Given $GITHUB_TOKEN is "OTOKEN"
+    When I successfully run `hub create`
+    Then the output should not contain "github.com password"
+    And the output should not contain "github.com username"
+    And the file "../home/.config/hub" should not exist
+
   Scenario: Wrong password
     Given the GitHub API server:
       """
--- a/github/config.go
+++ b/github/config.go
@@ -57,29 +57,37 @@ func (c *Config) PromptForHost(host string) (h *Host, err error) {
 		return
 	}
 
-	user := c.PromptForUser(host)
-	pass := c.PromptForPassword(host, user)
-
 	client := NewClient(host)
-	var code, token string
-	for {
-		token, err = client.FindOrCreateToken(user, pass, code)
-		if err == nil {
-			break
-		}
 
-		if ae, ok := err.(*AuthError); ok && ae.IsRequired2FACodeError() {
-			if code != "" {
-				ui.Errorln("warning: invalid two-factor code")
+	var user, pass, token string
+	token = c.DetectToken(host)
+	usingEnvVarToken := true
+
+	if token == "" {
+		usingEnvVarToken = false
+		user = c.PromptForUser(host)
+		pass = c.PromptForPassword(host, user)
+
+		var code string
+		for {
+			token, err = client.FindOrCreateToken(user, pass, code)
+			if err == nil {
+				break
 			}
-			code = c.PromptForOTP()
-		} else {
-			break
-		}
-	}
 
-	if err != nil {
-		return
+			if ae, ok := err.(*AuthError); ok && ae.IsRequired2FACodeError() {
+				if code != "" {
+					ui.Errorln("warning: invalid two-factor code")
+				}
+				code = c.PromptForOTP()
+			} else {
+				break
+			}
+		}
+
+		if err != nil {
+			return
+		}
 	}
 
 	client.Host.AccessToken = token
@@ -95,7 +103,10 @@ func (c *Config) PromptForHost(host string) (h *Host, err error) {
 		Protocol:    "https",
 	}
 	c.Hosts = append(c.Hosts, *h)
-	err = newConfigService().Save(configsFile(), c)
+
+	if !usingEnvVarToken {
+		err = newConfigService().Save(configsFile(), c)
+	}
 
 	return
 }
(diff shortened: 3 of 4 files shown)


mislav:
Thanks for working on this. Please see [features/authentication.feature](https://github.com/github/hub/blob/7ea2575ed96d150ee19f70edea4bd42c7c2f0b83/features/authentication.feature)

parkr:
Thanks for your feedback, @mislav. I have added two tests here. If this is insufficient, please advise. I wasn't sure whether you wished for a test to see if this token fails etc as that is covered by other tests.

features/authentication.feature
```diff
@@ -117,6 +117,45 @@ Feature: OAuth authentication
...
+      get('/user') {
+        json :login => 'mislav'
+      }
+      post('/user/repos') {
+        json :full_name => 'mislav/dotfiles'
+      }
+      """
+    Given $GITHUB_TOKEN is "feline01234"
+    When I run `hub create` interactively
+    When I type "mislav"
```
mislav: I feel that if you provide a GITHUB_TOKEN, you shouldn't also need to provide the username. The token itself is enough to authenticate.

parkr: Ah, ok. From your comment ( https://github.com/github/hub/issues/463#issuecomment-32387755 ) I thought that the username was required. I'll rework this at your request, however!

README.md
```diff
@@ -284,7 +284,8 @@ Hub will prompt for GitHub username & password the first time it needs to access
 the API and exchange it for an OAuth token, which it saves in "~/.config/hub".
 
 To avoid being prompted, use **GITHUB_USER** and **GITHUB_PASSWORD** environment
-variables.
+variables. If you would prefer not to expose your password, you may use
+**GITHUB_TOKEN**. The token must have **repo** permissions.
```
parkr: Maybe 

> Alternatively, you may provide **GITHUB_TOKEN**, an access token with **repo** permissions.

?

parkr:
@mislav I have addressed your concerns about the username. We're good to go.

jingweno:
LGTM. @mislav?

features/authentication.feature
```diff
@@ -117,6 +117,25 @@ Feature: OAuth authentication
     Then the output should not contain "github.com password for mislav"
     And the file "../home/.config/hub" should contain "oauth_token: OTOKEN"
 
+  Scenario: Credentials from GITHUB_TOKEN
+    Given the GitHub API server:
+      """
+      post('/authorizations') {
```
mislav: `POST /authorizations` should never happen in this case. There is no auth phase. We provided an existing token.

features/authentication.feature
```diff
@@ -117,6 +117,25 @@ Feature: OAuth authentication
...
+        assert_basic_auth 'mislav', 'kitty'
+        json :token => 'OTOKEN'
+      }
+      get('/user') {
+        json :login => 'mislav'
+      }
+      post('/user/repos') {
```
mislav: We need to verify that the token "feline0123" got used for this request.

features/authentication.feature
```diff
@@ -117,6 +117,25 @@ Feature: OAuth authentication
...
+        json :full_name => 'mislav/dotfiles'
+      }
+      """
+    Given $GITHUB_TOKEN is "feline0123"
+    When I successfully run `hub create`
+    Then the output should not contain "github.com password"
+    And the file "../home/.config/hub" should contain "oauth_token: feline0123"
```
mislav: The token provided by environment variable should **never** result in `~/.config/hub` changing.

features/authentication.feature
parkr: How might I go about doing that? `params["token"]`? Is there a `request.headers["Authorization"]` variable?

parkr:
@mislav I believe this latest push will be more to your liking.

features/authentication.feature
```diff
@@ -117,6 +117,24 @@ Feature: OAuth authentication
...
+        json :login => 'mislav'
+      }
+      post('/user/repos') {
+        assert({ "HTTP_AUTHORIZATION" => "token feline0123" }, env)
+        json :full_name => 'mislav/dotfiles'
+      }
+      """
+    Given $GITHUB_TOKEN is "feline0123"
+    When I successfully run `hub create`
+    Then the exit status should be 0
```
mislav: Exit status is already covered by an assertion above it

features/authentication.feature
```diff
@@ -117,6 +117,24 @@ Feature: OAuth authentication
...
     And the file "../home/.config/hub" should contain "oauth_token: OTOKEN"
 
+  Scenario: Credentials from GITHUB_TOKEN
+    Given the GitHub API server:
+      """
+      get('/user') {
+        json :login => 'mislav'
+      }
+      post('/user/repos') {
+        assert({ "HTTP_AUTHORIZATION" => "token feline0123" }, env)
```
mislav: I don't think you need to tweak the `assert` method here. How about simply using the approach like other tests?

```rb
halt 401 unless request.env['HTTP_AUTHORIZATION'] == 'token OTOKEN'
```

features/support/local_server.rb
```diff
@@ -102,7 +102,7 @@ def generate_patch(subject)
 --- /dev/null
 +++ b/README.md
 +hello
--- 
+--
```
mislav: Watch out when letting your editor clean up whitespace. This is inside a patch file, so it's best to avoid making any changes to it.

github/config.go
```diff
@@ -95,7 +103,10 @@ func (c *Config) PromptForHost(host string) (h *Host, err error) {
 		Protocol:    "https",
 	}
 	c.Hosts = append(c.Hosts, *h)
-	err = newConfigService().Save(configsFile(), c)
+
+	if !usingEnvVarToken {
+		err = newConfigService().Save(configsFile(), c)
```
mislav: Nice, simple approach :+1:

mislav:
Looks good! Could you also add a cuke that verifies that it detects your username correctly? Since the only piece of information you're passing to hub is the OAUTH token, we want to verify that hub will be able to fetch your user info from the API for commands like `hub clone foo` which will automatically clone `YOURUSER/foo`.

parkr:
> Could you also add a cuke that verifies that it detects your username correctly? Since the only piece of information you're passing to hub is the OAUTH token, we want to verify that hub will be able to fetch your user info from the API for commands like hub clone foo which will automatically clone YOURUSER/foo.

I gave this a shot, but it seems to fail. If you have any insight, that'd be great. It seems it's not hitting the API I have setup in that Cuke at all.

parkr:
Would you mind taking a look at this, @mislav?

features/authentication.feature
```diff
@@ -117,6 +117,25 @@ Feature: OAuth authentication
...
+  Scenario: Credentials from GITHUB_TOKEN
+    Given the GitHub API server:
+      """
+      get('/user') {
+        halt 401 unless request.env["HTTP_AUTHORIZATION"] == "token OTOKEN"
+        json :login => 'mislav'
+      }
+      post('/user/repos') {
+        halt 401 unless request.env["HTTP_AUTHORIZATION"] == "token OTOKEN"
+        p request.env
```
mislav: Debugging statement slipped in here

mislav:
@parkr Found the problem, solving it. It requires some refactoring, though. I'll take it from here, thanks!

parkr:
> @parkr Found the problem, solving it. It requires some refactoring, though. I'll take it from here, thanks!

Thanks a million, @mislav!

PR #959, opened 2015-07-17
Add installation instructions for zsh tab-completion

heitortsergent:
Added a README to the folder `etc` containing instructions on how to use the `hub.zsh_completion` file.

### DIFF
--- a/etc/README.md
+++ b/etc/README.md
@@ -1,0 +1,22 @@
+# Installation instructions
+
+## zsh
+
+Create a new folder for completions:
+
+```sh
+mkdir -p ~/.zsh/completions
+```
+
+Download the file hub.zsh_completion and rename it to `_hub`:
+
+```sh
+curl https://github.com/github/hub/blob/master/etc/hub.zsh_completion >  ~/.zsh/completions/_hub
+```
+
+Then add the following lines to your .zshrc file:
+
+```sh
+fpath=(~/.zsh/completions $fpath) 
+autoload -U compinit && compinit
+```
--- a/etc/README.md
+++ b/etc/README.md
@@ -1,5 +1,28 @@
 # Installation instructions
 
+## bash
+
+Open your `.bashrc` file and add:
+
+> If you want to set up Git to automatically have Bash shell completion for all users, copy the `hub.bash_completion` script to the `/opt/local/etc/bash_completion.d` directory on Mac systems or to the `/etc/bash_completion.d/` directory on Linux systems. ([Source](https://git-scm.com/book/en/v1/Git-Basics-Tips-and-Tricks#Auto-Completion))
+
+* [Link to git auto-completion bash file](https://github.com/git/git/blob/master/contrib/completion/git-completion.bash)
+
+```sh
+# Make sure you've aliased hub to git
+eval "$(hub alias -s)"
+
+# And make sure that the git auto-completion is being loaded
+if [ -f /path/to/git-completion.bash ]; then
+    . /path/to/git-completion.bash
+fi
+
+# Load hub autocompletion
+if [ -f /path/to/hub.bash_completion ]; then
+    . /path/to/hub.bash_completion
+fi
+```
+
 ## zsh
 
 Create a new folder for completions:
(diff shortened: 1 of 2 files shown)


mislav:
Thanks for the suggestion! You'll have to sign the GitHub CLA in order for this to be accepted. See the failure in the PR checks section on this page.

What about bash instructions? Most of our users use bash.

etc/README.md
```diff
@@ -0,0 +1,22 @@
...
+Create a new folder for completions:
+
+```sh
+mkdir -p ~/.zsh/completions
+```
+
+Download the file hub.zsh_completion and rename it to `_hub`:
+
+```sh
+curl https://github.com/github/hub/blob/master/etc/hub.zsh_completion >  ~/.zsh/completions/_hub
```
mislav: This needs to follow redirects by adding the `-L` flag. But why download it from the server when people should probably already have hub downloaded on their computer? Maybe it would be better to instruct them to simply copy the `etc/hub.zsh_completion` from their download the the zsh completions location.

heitortsergent:
@mislav thanks for the feedback! I signed the CLA, changed the zsh instructions to copy the file instead of using cURL, and added instructions for bash. Let me know what you think. 💃

Honghe:
the installation instructions for zsh is nice. But why this pull request can not be merged?

heitortsergent:
@Honghe thanks! It looks like all tests are :+1: now, I did a rebase with master.

etc/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+* [Link to git auto-completion bash file](https://github.com/git/git/blob/master/contrib/completion/git-completion.bash)
+
+```sh
+# Make sure you've aliased hub to git
+eval "$(hub alias -s)"
+
+# And make sure that the git auto-completion is being loaded
+if [ -f /path/to/git-completion.bash ]; then
+    . /path/to/git-completion.bash
```
mislav: This is not necessary. Let's assume they're already successfully loading the git completion script, and they just want to extend the functionality with hub.

etc/README.md
```diff
@@ -0,0 +1,46 @@
+# Installation instructions
+
+## bash
+
+Open your `.bashrc` file and add:
```
mislav: This should be `~/.bashrc`. And, for OS X, we should suggest that they edit `~/.bash_profile` instead since that seems to be the convention right now.

etc/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+> If you want to set up Git to automatically have Bash shell completion for all users, copy the `hub.bash_completion` script to the `/opt/local/etc/bash_completion.d` directory on Mac systems or to the `/etc/bash_completion.d/` directory on Linux systems. ([Source](https://git-scm.com/book/en/v1/Git-Basics-Tips-and-Tricks#Auto-Completion))
```
mislav: Why is this marked up as a blockquote?

Also note that `/opt/local/etc/bash_completion.d` directory doesn't exist by default on OS X. This is the directory which is used for bash completion scripts by Homebrew if you installed it in `/opt/local`, which is minority of users. Most of them will have Homebrew in `/usr/local`. However if they use Homerew, then they can just use `brew install hub` and they won't have to do any extra steps to set up bash completion. So I guess that Homebrew needs its own mention.

heitortsergent: I just marked it as a blockquote because it's more of a side-note. Do you think it's better to just remove it?

mislav: Blockquote is usually reserved for quoting what somebody said, for instance if I reply to you and quote a part of what you said, just like in email:

> Do you think it's better to just remove it?

heitortsergent: Ah, got it. :)  I'm not sure on what to change, though, do you think it's better to just remove that whole snippet? Or should I add a h2 header above bash for Homebrew and mention that doing `brew install hub` automatically sets up autocompletion?

mislav: Yeah, remove this whole snippet and add a special section for bash + Homebrew above.

heitortsergent:
@mislav I updated the PR, please let me know if the "bash + Homebrew" section looks okay. :)

mislav:
Thanks!

PR #961, opened 2015-07-18
client.Releases() fetches all pages.

pda:
Currently `hub release` only shows the first page of results from the [GitHub Releases API][api]. It doesn't indicate that results have been truncated, and doesn't provide a way to fetch more/all results.

With this patch, `client.Releases()` (and thus `hub release`) follows the `rel="next"` URL from [GitHub's Link header][link] until all results have been fetched.

---

Some design changes could be made around this — e.g. providing pagination/limit controls to the user, or streaming the results rather than fetching them all up front. However compared to truncation of results, I think this change is an improvement worth merging prior to considering further changes.

[api]: https://developer.github.com/v3/repos/releases/
[link]: https://developer.github.com/v3/#link-header

### DIFF
--- a/github/client.go
+++ b/github/client.go
@@ -257,10 +257,21 @@ func (client *Client) Releases(project *Project) (releases []octokit.Release, er
 		return
 	}
 
-	releases, result := api.Releases(client.requestURL(url)).All()
-	if result.HasError() {
-		err = FormatError("getting release", result.Err)
-		return
+	for url != nil {
+		page, result := api.Releases(client.requestURL(url)).All()
+		if result.HasError() {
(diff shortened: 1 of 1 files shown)


mislav:
Good catch; thanks for the workaround. Ideally this should be handled by the underlying library (go-octokit), but I totally get your concern re: potential slowness when fetching all the pages unbounded.

My dream is that the underlying API library should handle both pagination concerns and provide some sort of caching so that only new releases get fetched, and old ones get restored from disk cache without having to paginage through all results unnecessarily.

We are still figuring out how the `hub release` command should behave. Right now I don't think just listing everything is very useful since the output is neither machine-readable nor very nice for humans.

PR #968, opened 2015-08-05
Allow to set pull request milestone and labels.

calavera:
Add the `-M/--milestone` and `-l/--labels` flags to `hub pull-request` to
set the milestone or tags when a pull request is created.

Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -41,9 +41,11 @@ var (
 	flagPullRequestIssue,
 	flagPullRequestMessage,
 	flagPullRequestAssignee,
+	flagPullRequestLabels,
 	flagPullRequestFile string
 	flagPullRequestBrowse,
 	flagPullRequestForce bool
+	flagPullRequestMilestone uint64
 )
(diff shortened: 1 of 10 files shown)


jingweno:
Just tried this out, working like a charm! :+1:

mislav:
Yes, it works great!

The milestone option is a bit clumsy since it expects you to know a milestone ID. My ideal feature would be that it accepts a milestone name. More thoughts here: #1089

jingweno:
Have been using attaching labels. Haven't tried attaching milestones yet.

Maybe this discussion could go in another thread. But it would be nice if we could also define labels and milestones when editing the title and body. Maybe we could introduce a new meta data section. The template could go like:

```
Title

Body

# label: one, two
# milestone: M1

# Requesting a pull to master from branch
#
# Write a message for this pull request. The first block
# of text is the title and the rest is description.
```

We could discuss more how the syntax should go.

PR #970, opened 2015-08-06
Handle git worktrees when computing the head repo

### LINKED ISSUE #969: hub pull-request broken in combination with git-worktree
`headProject` is incorrectly computed as the upstream project.

Example:

```
/code/hub> git co -b topic/foo; git commit -m dummy --allow-empty
Switched to a new branch 'topic/foo'
[topic/foo 2a36c7c] dummy

topic/foo /code/hub> git push
Counting objects: 1, done.
Writing objects: 100% (1/1), 182 bytes | 0 bytes/s, done.
Total 1 (delta 0), reused 0 (delta 0)
To git@github.com:retronym/hub.git
 * [new branch]      topic/foo -> topic/foo

topic/foo /code/hub> hub pull-request # $EDITOR pops up with "dummy\n\nRequesting a pull to github:master from retronym:topic/foo"

topic/foo /code/hub> rm .git/PULLREQ_EDITMSG

topic/foo /code/hub> git worktree add ../hub2 -b topic/bar
Enter ../hub2 (identifier hub2)
Switched to a new branch 'topic/bar'

topic/foo /code/hub> cd ../hub2

topic/bar /code/hub2> git push
Total 0 (delta 0), reused 0 (delta 0)
To git@github.com:retronym/hub.git
 * [new branch]      topic/bar -> topic/bar

topic/bar /code/hub2> git pull-request -b master  # $EDITOR shows up without the commit info and with  "Requesting a pull to github:master from github:topic/bar"
```

A workaround is to explicitly specify the head `repo:ref`:

```
topic/bar /code/hub2 > git pull-request -h retronym:topic/bar # $EDITOR pops up with "dummy\n\nRequesting a pull to github:master from retronym:topic/bar"
```

The following snippets of the code seem responsible. 

```go
// git.go
func HasFile(segments ...string) bool {
	dir, err := Dir()
	if err != nil {
		return false
	}

	s := []string{dir}
	s = append(s, segments...)
	path := filepath.Join(s...)
	if _, err := os.Stat(path); err == nil {
		return true
	}

	return false
}
```

```go
// branch.go
			if git.HasFile("refs", "remotes", remote.Name, shortName) {
				name := fmt.Sprintf("refs/remotes/%s/%s", remote.Name, shortName)
				branch = &Branch{b.Repo, name}
				break
			}
```

The location of the `refs` file should be resolved with `git rev-parse --git-path refs`, rather than `$(git rev-parse --git-dir)/refs`.

For backwards compatbility, the current scheme will need used a a fallback as `--git-path` was only added in Git 2.5.0: https://github.com/git/git/commit/557bd833bb4e2bd1f109f6f911ab517b6ba3d27a

retronym:
Git 1.5.0 introduces support for worktrees. A worktree has a
.git file, rather than a folder. (The file contents points to to a
.git folder buried underneath the .git folder of the main repository.)

A new option has been added to resolve paths within the git dir,
`git rev-parse --git-path foo/bar`.

The commit uses this new facility, but falls back to the old approach
for backwards compability.

Fixes #969

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -37,6 +37,17 @@ func Dir() (string, error) {
 }
 
 func HasFile(segments ...string) bool {
+	// The blessed way to resolve paths in git dir, since Git 1.5.0
+	// Needed when working with git worktrees.
+	output, err := gitOutput("rev-parse", "-q", "--git-path", filepath.Join(segments...))
+	if err == nil {
+		_, err := os.Stat(output[0])
+		if err == nil {
+			return true
+		}
+	}
+
+	// Fallback to $(git rev-parse --git-dir)/segment1/segment2 for older git versions
 	dir, err := Dir()
 	if err != nil {
 		return false


git/git.go
```diff
@@ -37,11 +37,21 @@ func Dir() (string, error) {
...
 
 func HasFile(segments ...string) bool {
-	dir, err := Dir()
-	if err != nil {
-		return false
+	// The blessed way to resolve paths in git dir, since Git 1.5.0
+	// Needed when working with git worktrees.
+	output, err1 := gitOutput("rev-parse", "-q", "--git-path", filepath.Join(segments...))
+	if (err1 == nil) {
+	    _, err2 := os.Stat(output[0])
```
mislav: Feel free to reuse a `err` variable instead of creating `err1` and `err2`.

git/git.go
```diff
@@ -37,11 +37,21 @@ func Dir() (string, error) {
...
+		if (err2 == nil) {
+			return true
+		}
 	}
 
+	// Fallback to $(git rev-parse --git-dir)/segment1/segment2 for older git versions
+    dir, err := Dir()
```
mislav: Seems like indentation is wonky here? Please run `script/format`

retronym:
Done and done.

git/git.go
```diff
@@ -37,11 +37,21 @@ func Dir() (string, error) {
 }
 
 func HasFile(segments ...string) bool {
-	dir, err := Dir()
-	if err != nil {
-		return false
+	// The blessed way to resolve paths in git dir, since Git 1.5.0
+	// Needed when working with git worktrees.
+	output, err1 := gitOutput("rev-parse", "-q", "--git-path", filepath.Join(segments...))
```
mislav: It seems to me that this logic belongs to `Dir()` function instead of `HasFile`. The `Dir()` function is supposed to return the path to the `.git` directory, right?

retronym: Nope:

```
topic/bar /code/hub2> git rev-parse --git-dir
/Users/jason/code/hub/.git/worktrees/hub2
topic/bar /code/hub2> git rev-parse --git-path refs
/Users/jason/code/hub/.git/refs
```

mislav: Hmm so it's more complicated than my idea, huh? OK let's go with your approach.

mislav:
We'll need tests for this, just to make sure we don't regress. A test in `git_test.go` might be tricky to set up but would be worth it. @retronym feeling up to it?

git/git.go
retronym: If you could assume git 1.5 you could use An alternative approach would be to take advantage of:

```
% git rev-parse --symbolic-full-name @{upstream}
refs/remotes/retronym/topic/typer-debug-implicits
```

But while you are being backwards compatible, it probably isn't worth using that conditionally.

retronym:
I'm not sure I have time to familiarize myself with the testing approach (this was already a bigger diversion in my afternoon that I had planned for!).

mislav:
Merged; thanks!

PR #975, opened 2015-08-12
Allow --ff with merge

### LINKED ISSUE #973: Allow --ff with merge
At the moment when you try and merge a pull request using --ff (not --ff-only) it adds the --no-ff argument to the end.
There should be a check [here](https://github.com/github/hub/blob/master/commands/merge.go#L80)(?) which makes sure --no-ff is not added when using --ff.

This is so people can merge pull requests using fast-forward and also do so recursively using hub!

CraigyDavi:
Closes #973

### DIFF
--- a/commands/merge.go
+++ b/commands/merge.go
@@ -77,7 +77,7 @@ func transformMergeArgs(args *Args) error {
 	mergeMsg := fmt.Sprintf("Merge pull request #%v from %s\n\n%s", id, mergeHead, pullRequest.Title)
 	args.AppendParams(mergeHead, "-m", mergeMsg)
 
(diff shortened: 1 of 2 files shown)


mislav:
Looks good! This would be great if it had a test as well. You could [duplicate this test](https://github.com/github/hub/blob/4c018ba9d5bb8e27d139527b5445c45ca507d4fa/features/merge.feature#L34-L58) and change `--ff-only` in that test to `-ff`.

You can run these test with `bundle exec cucumber features/merge.feature`. If you're not able to perform the necessary setup for testing locally, just push new commits to this branch and Travis CI will perform the checks for us.

CraigyDavi:
Added tests and they appear to work.

mislav:
Thanks!

Issue #33, opened 2010-04-16
Updated to use ronn v0.5


xoebus:
Because of the name change of `ron` to `ronn` it's now difficult for someone who didn't have the old version of `ron` to build the documentation to test changes. This branch changes the documentation and Rakefile to be compatible with the current gem version of `ronn`, v0.5:

http://github.com/xoebus/hub/tree/ronn

PR #981, opened 2015-08-18
Add hub-apr

ryo33:
Applying patches from PR number like `$ hub apr 33` or `$ hub apr "#33"`.

### DIFF
--- a/commands/apply.go
+++ b/commands/apply.go
@@ -4,7 +4,9 @@ import (
 	"io"
 	"io/ioutil"
 	"regexp"
+	"strings"
 
+	"github.com/github/hub/git"
 	"github.com/github/hub/github"
 	"github.com/github/hub/utils"
 )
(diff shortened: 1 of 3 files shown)


mislav:
Hi thanks for the excellent pull request. However, is this basically equivalent to `hub apply <pullrequest-url>`?

ryo33:
`hub apr <pullrequest-number>` will be transformed to `hub am -3 <pullrequest-url>` for now.
Please see https://github.com/ryo33/hub/blob/hub-apr/commands/apply_test.go#L11-L16 for reference.

mislav:
So this is just a shortcut for `am -3 <pullrequest-url>`? I feel we need something like this, although I'm not a fan of a new command called `apr`.

I was thinking about making a new command called `hub pr` or something that will handle everything PR related: from opening them, browsing them, checking them out as branches, to merging.

ryo33:
Yes, it's just a shortcut.
So the shortcut should be `hub pr --am -3 <pullrequest-number>`?

PR #982, opened 2015-08-19
Add hub-pr

ryo33:
`hub pr` is a command handling pull requests with their numbers and `--*` argument.
It has been already implemented that `--apply`, `--am` and `--browse`.
Please see #981 for reference.
And, can I get your opinion?

### DIFF
--- a/commands/pr.go
+++ b/commands/pr.go
@@ -1,0 +1,102 @@
+package commands
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/github/hub/github"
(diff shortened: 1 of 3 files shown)


mislav:
Hi, sorry for the late reply, but I'm not really a fan of the syntax `hub pr --am -q -3 387`. Since there is already a `hub am` subcommand, it's confusing to use it as `hub pr --am`.

I have some ideas for the `pr` command, but it would be completely different than just allowing the PR number to get specified for existing commands.

Issue #225, opened 2012-08-17
Dedicated command for authentication & configuration


stuartpb:
All I really use hub for is pulling down my own repos. As a result, I've never been prompted for my username/password to connect to the API, so I'm always prompted for my username. I'd appreciate some way to just trigger the configuration by itself.

kamalmarhubi:
Is there anything approaching a spec for this feature? I may be willing to implement browser-based authentication (https://github.com/github/hub/issues/889). Ideally I'd do it in a way that didn't require big rework when this `auth` command eventually gets worked on.

mislav:
I think we want the command to be called `hub auth` and have several subcomands, like `login`, `whoami` and possibly `config` to see per-host hub configuration.

By default you will be authenticating with `github.com` but you could specify your Enterprise instance with `--site git.company.com` or `http://git.mycompany.com` if the Enterprise instance is to be communicated with over HTTP.

When implementing browser-based authentication, how do you plan to deal with the following problems:

1. For browser-based authentication, hub needs to be registered as an OAuth app. How will we embed securely our own OAuth client tokens in the binary?
2. An OAuth app needs to be registered will a callback URL. I presume you want a callback URL such as `http://localhost:<port>`. However, that means that we have to depend on whatever `<port>` we choose to be available on localhost.
3. People need to be able to authenticate to Enterprise instances. An OAuth app that we register with github.com (presumably) won't be known to the Enterprise instance because it won't be in its own database. How do you plan to compensate for that?

kamalmarhubi:
> For browser-based authentication, hub needs to be registered as an OAuth app. How will we embed securely our own OAuth client tokens in the binary?

This problem applies to any client application so I don't think it's really in scope. See also the [Threat: Obtaining Client Secrets][rfc-client-secrets] section from the spec on the topic.

[rfc-client-secrets]: http://tools.ietf.org/html/rfc6819#section-4.1.1

> An OAuth app needs to be registered will a callback URL. I presume you want a callback URL such as http://localhost:<port>. However, that means that we have to depend on whatever <port> we choose to be available on localhost.

In my last comment on #889, I pointed to the Python utility Google uses in its apps: https://github.com/google/oauth2client/blob/master/oauth2client/tools.py

They attempt to open two ports, and use those as redirection URLs. If they fail, they pass a special URN that signifies that a code should be given back in the browser for the user to paste in.

[...]

PR #992, opened 2015-09-09
Add user command

sbp:
This new command writes the default user from hub config to stdout with a
trailing newline. For example, if you are me you get:

```sh
$ hub user
sbp
```

This is useful for automating contribution scripts. Say, for example, that you
want to write a script that pushes your contributions to some feature branch.
Using this command, you can do:

```sh
$ hub push `hub user` `git rev-parse --abbrev-ref HEAD`
```

Whereas otherwise you would have to manually fill in the user name, even though
hub knows what user name it would normally use from config.

### DIFF
--- a/commands/user.go
+++ b/commands/user.go
@@ -1,0 +1,35 @@
+package commands
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/github/hub/github"
+	"github.com/github/hub/utils"
+)
+
+var cmdUser = &Command{
+	Run:   user,
+	Usage: "user",
+	Short: "Show the default user name",
+	Long: `Show which user name will be used by hub commands.`,
+}
+
+func init() {
+	CmdRunner.Use(cmdUser)
+}
+
+/*
+  $ gh user
+  YOUR_USER
+*/
+func user(cmd *Command, args *Args) {
+    var host *github.Host
+    config := github.CurrentConfig()
+    host, err := config.DefaultHost()
+    if err != nil {
+            utils.Check(github.FormatError("reading config", err))
+    }
+    fmt.Print(host.User + "\n")
+    os.Exit(0)
+}


mislav:
Thanks, this looks good!

1. I would like that the command is called something more specific, like `whoami`, in case we want to add a `user` command in the future that looks up other users or whatnot.
2. This needs tests. Specifically, I would like to see some cukes created in `features/whoami.feature`; one that verifies that your username is displayed when proper configuration is in place, and one with the command failure in case there is no configuration (i.e. the user hasn't yet authenticated). Look at other cukes to get a feeling of how to write them.

mislav:
This functionality is related to #225.

Dechowmedia:
Hmmm how about the command was `hub account` ? :)

timofonic:
@mislav  Any news?

sbp:
@timofonic This kinda bifurcated #225 where @mislav had [already written](https://github.com/github/hub/issues/225#issuecomment-109766782) in July 2015:

> I think we want the command to be called `hub auth` and have several subcomands, like `login`, `whoami` and possibly `config` to see per-host hub configuration.

I hadn't seen that at the time I made this PR (#992). Writing all of `hub auth` plus associated tests is not something that I planned or plan to do.

I originally wrote this PR because I think contributing back to github projects using the web based PR mechanism is a lot more difficult than it needs to be. For some of my projects I want people contributing to them as much as possible, and so I started to explore anything that I could do to encourage that. A command line option seemed like a pretty good encouragement, because I could just make a script that does everything for you—it could build a local copy of hub, prompt for your username, and then automatically push your changes. You could clone my code, make your changes, check it in, and then simply run a script in the repo in order to contribute back!

When this pull request was not accepted, I lost interest, and to be honest I haven't even used `hub` myself since then even though I'm a pretty active github user. It's entirely reasonable for the `hub` maintainers to want a nice, unified API and extensive tests for this feature, and I hope that eventually they land in `hub`.

kandice1978:
Um

PR #994, opened 2015-09-14
Update README - Add Brew Support

IWriteThings:
Most users will know to check BREW but it can be helpful for folks to know that Brew is a viable option for installing this tool.


parkr:
:+1:

PR #999, opened 2015-09-27
Enhance `hub ci-status` to consume and display combined statuses

### LINKED ISSUE #943: Show all build statuses with `ci-status`
If a repository has multiple CI's setup, `ci-status -v` will only show the first one in the list. I would expect it to list all CI's

Tested against version `2.2.1`

mislav:
Now `ci-status` consumes the new combined statuses API endpoint designed for when multiple contexts report statuses independently.

The verbose output is now colorized and is in format where each line is:

    ✔︎ <Tab> [CONTEXT] <Tab> [URL]

Fixes #943

@jingweno Since Octokit only supported the old API endpoint and I needed access to the new one, here I did an experiment called "simple client" where I bypassed Octokit and used net/http and JSON parsing directly. I wanted to see how much would be involved, because for some future commands I need to do certain things that Octokit doesn't support and I don't have the bandwidth to maintain both libraries right now. What do you think?

### DIFF
--- a/github/client.go
+++ b/github/client.go
@@ -336,8 +336,8 @@ func (client *Client) UploadReleaseAsset(uploadUrl *url.URL, asset *os.File, con
 }
 
 type CIStatusResponse struct {
-	State    string      `json:"state"`
-	Statuses []*CIStatus `json:"statuses"`
+	State    string     `json:"state"`
+	Statuses []CIStatus `json:"statuses"`
 }
 
 type CIStatus struct {
@@ -352,7 +352,7 @@ func (client *Client) FetchCIStatus(project *Project, sha string) (status *CISta
 		return
 	}
 
-	res, err := api.Get("repos/" + project.Owner + "/" + project.Name + "/commits/" + sha + "/status")
+	res, err := api.Get(fmt.Sprintf("repos/%s/%s/commits/%s/status", project.Owner, project.Name, sha))
 	if err != nil {
 		return
 	}
@@ -362,7 +362,7 @@ func (client *Client) FetchCIStatus(project *Project, sha string) (status *CISta
 	}
 
 	status = &CIStatusResponse{}
-	err = res.Parse(status)
+	err = res.Unmarshal(status)
 
 	return
 }
--- a/github/client.go
+++ b/github/client.go
@@ -335,27 +335,34 @@ func (client *Client) UploadReleaseAsset(uploadUrl *url.URL, asset *os.File, con
 	return
 }
 
-func (client *Client) CIStatus(project *Project, sha string) (status *octokit.Status, err error) {
-	url, err := octokit.StatusesURL.Expand(octokit.M{"owner": project.Owner, "repo": project.Name, "ref": sha})
+type CIStatusResponse struct {
+	State    string      `json:"state"`
+	Statuses []*CIStatus `json:"statuses"`
+}
+
+type CIStatus struct {
+	State     string `json:"state"`
+	Context   string `json:"context"`
+	TargetUrl string `json:"target_url"`
+}
+
+func (client *Client) FetchCIStatus(project *Project, sha string) (status *CIStatusResponse, err error) {
+	api, err := client.simpleApi()
 	if err != nil {
 		return
 	}
 
-	api, err := client.api()
+	res, err := api.Get("repos/" + project.Owner + "/" + project.Name + "/commits/" + sha + "/status")
 	if err != nil {
-		err = FormatError("getting CI status", err)
+		return
+	}
+	if res.StatusCode != 200 {
+		err = fmt.Errorf("Unexpected HTTP status code: %d", res.StatusCode)
 		return
 	}
 
-	statuses, result := api.Statuses(client.requestURL(url)).All()
-	if result.HasError() {
-		err = FormatError("getting CI status", result.Err)
-		return
-	}
-
-	if len(statuses) > 0 {
-		status = &statuses[0]
-	}
+	status = &CIStatusResponse{}
+	err = res.Parse(status)
 
 	return
 }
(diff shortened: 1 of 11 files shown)


github/client.go
```diff
@@ -335,27 +335,34 @@ func (client *Client) UploadReleaseAsset(uploadUrl *url.URL, asset *os.File, con
 	return
 }
 
-func (client *Client) CIStatus(project *Project, sha string) (status *octokit.Status, err error) {
-	url, err := octokit.StatusesURL.Expand(octokit.M{"owner": project.Owner, "repo": project.Name, "ref": sha})
+type CIStatusResponse struct {
+	State    string      `json:"state"`
+	Statuses []*CIStatus `json:"statuses"`
```
jingweno: This could be `[]CiStatus`. A array of pointers doesn't seem to buy us anything.

mislav: Yeah I wanted to avoid copying structs when iterating over this array, but I guess it doesn't matter?

github/client.go
```diff
@@ -544,14 +551,20 @@ func (client *Client) FindOrCreateToken(user, password, twoFactorCode string) (t
...
 }
 
-func (client *Client) api() (c *octokit.Client, err error) {
+func (client *Client) ensureAccessToken() (err error) {
 	if client.Host.AccessToken == "" {
-		host, e := CurrentConfig().PromptForHost(client.Host.Host)
-		if e != nil {
-			err = e
-			return
+		host, err := CurrentConfig().PromptForHost(client.Host.Host)
```
jingweno: `:=` creates a local variable `err` which is not the same as the one to return `(err, error)`.  Currently `(err, error)` is always `nil`. We could directly return the err if it's not nil:

```go
host, err := CurrentConfig().PromptForHost(client.Host.Host)
if err == nil {
  client.Host = host
} else {
  return err
}

return nil
```

Or like before, use a different local variable name and assign it to `(err error)`:

```go
host, e := CurrentConfig().PromptForHost(client.Host.Host)
if e != nil {
  err = e
  return
} 

client.Host = host
...
```

github/client.go
```diff
@@ -574,15 +604,11 @@ func (client *Client) newOctokitClient(auth octokit.AuthMethod) *octokit.Client
 	return c
 }
 
-func (client *Client) absolute(endpoint string) *url.URL {
-	u, _ := url.Parse(endpoint)
-	if u.Scheme == "" && client.Host != nil {
+func (client *Client) absolute(host string) *url.URL {
+	u, _ := url.Parse("https://" + host)
```
jingweno: Previously `endpoint` includes the http schema. Would it break anything?

github/http.go
```diff
@@ -79,7 +76,7 @@ func (t *verboseTransport) dumpResponse(resp *http.Response) {
 }
 
 func (t *verboseTransport) dumpHeaders(header http.Header, indent string) {
-	dumpHeaders := []string{"Authorization", "X-GitHub-OTP", "Localtion"}
+	dumpHeaders := []string{"Authorization", "X-GitHub-OTP", "Location"}
```
jingweno: Good catch!

github/client.go
mislav: Previously `absolute()` created a URL that looks like `https://example.com`, but its Host was blank, instead its Path property was `/example.com`. So I decided to fix it. We were only passing plain hostnames to `endpoint` before anyway, so I changed the argument and tests to reflect that.

github/client.go
mislav: Oops, good catch!

github/http.go
```diff
@@ -175,3 +172,45 @@ func proxyFromEnvironment(req *http.Request) (*url.URL, error) {
...
+	}
+
+	return
+}
+
+type simpleResponse struct {
+	*http.Response
+}
+
+func (res *simpleResponse) Parse(dest interface{}) (err error) {
```
jingweno: I'm a bit indifference to this. But Go community seems to have the convention of naming deserialized method `Encode/Decode` or the object `Encoder/Decoder`, e.g., the JSON decoder etc. FWIW.

mislav: Good point. I might rename the method. I might also name it `Unmarshal` to indicate what's going on under the hood?

jingweno:
> Since Octokit only supported the old API endpoint and I needed access to the new one, here I did an experiment called "simple client" where I bypassed Octokit and used net/http and JSON parsing directly. I wanted to see how much would be involved, because for some future commands I need to do certain things that Octokit doesn't support and I don't have the bandwidth to maintain both libraries right now. What do you think?

I think this is prob. the best for `hub`, considering not much progress has been made on `go-octokit`. Besides, the current implementation of `go-octokit` may need some reevaluation.

/cc @pengwynn

github/client.go
jingweno: Oh okay, I see the confusion. Array/Slice and Map are pointers. So no copy will be made when they're passed around.

github/http.go
jingweno: > I might also name it Unmarshal to indicate what's going on under the hood?

Yeah, I think both `Unmarshal` and `Decode` would be a good indicator of what's happening.

mislav:
> Besides, the current implementation of `go-octokit` may need some reevaluation.

Yes, ideally go-octokit would be a true hypermedia client: following the links, caching etc.

Right now I feel that I would be faster if I prototyped some new commands using manual API calls and only deserialize the few JSON keys that I actually consume. I might extract this into a new subpackage so that it's clear it's standalone.

github/client.go
jingweno: Gotya.

github/client.go
mislav: Actually I just tested it now and it seems OK. The empty `return`s will correctly return the `err` object even if it has been "redefined" by `:=` operator. We're using the construct quite a lot throughout the codebase and it seems OK. I will keep it like this.

PR #1000, opened 2015-09-27
Check if GitHub repo exists before every `hub clone`

### LINKED ISSUE #868: `git clone resque` fails
```
$ git clone resque
Cloning into 'resque'...
fatal: remote error:
  Repository not found.
```

However,
```
$ git clone resque/resque
Cloning into 'resque'...
remote: Counting objects: 10977, done.
remote: Total 10977 (delta 0), reused 0 (delta 0), pack-reused 10977
Receiving objects: 100% (10977/10977), 3.21 MiB | 2.17 MiB/s, done.
Resolving deltas: 100% (4726/4726), done.
```

I have installed `hub` via homebrew and aliased with `eval "$(hub alias -s)"` in my `.bash_profile`

```
$ hub version
git version 2.3.5
hub version 2.2.1
```

mislav:
If it doesn't exist, print a more helpful error message about it so the user knows which "owner/repo" pair was tested. This aims to reduce confusion when cloning own repos such as `hub clone dotfiles`.

The downside is now there is an extra HTTP request on every `hub clone` because the repository existence is checked every time.

Fixes #868


features/clone.feature
```diff
@@ -113,6 +134,17 @@ Feature: hub clone
...
 
+  Scenario: Clone my repo that doesn't exist
+    Given the GitHub API server:
+      """
+      get('/repos/mislav/dotfiles') { status 404 }
+      """
+    When I run `hub clone dotfiles`
+    Then the exit status should be 1
+    And the stdout should contain exactly ""
+    And the stderr should contain exactly "Error: repository mislav/dotfiles doesn't exist\n"
```
mislav: Hmm maybe this message should also say "on github.com" so it's clear that the repo was searched for online and on which host?

parkr:
The extra network request is a bummer. I work on a lot of high-latency networks and this would likely reduce efficiency and effectiveness of this tool for me. Perhaps, if the 'git clone' operation fails, hub can print a message containing the relevant information?

"fatal: failed to clone mislav/dotfiles from GitHub.com"

mislav:
Then we would have to execute `git clone` internally, inspect its exit code and message, and replace the message with our own. Doable, but we would have to change more significantly how `clone` works. It's not a bad idea.

It's just that the API request will likely take the negligible amount of time compared to the clone operation.

parkr:
Optimizing for the successful case would seem preferable to me. Assume it's going to clone, clone it, and if it fails, print a message. I'm not suggesting you remove the git clone output, just add a line at the bottom of the output with the details about your clone attempt.

How expensive is a failed git clone operation? 

What if GitHub's API is down but clone operations are fine? Do I have to wait 30 seconds for it to time out for every clone? If the API is up, this might only add a second to the total real time for the user in the mean case. Is that second, multiplied over thousands of users who don't get 404's, worth it?

Just trying to save some network time here, because latency is horrifying for me. The intent is to print owner and host name information, right? I'm interested in providing that without any added overhead so it doesn't degrade the successful case (where the repo exists).

features/clone.feature
parkr: `Error: repository git@github.com:mislav/dotfiles doesn't exist`? Do you think the full path is too confusing? If this were used on an enterprise instance, for example, it would be great to specify.

parkr:
I don't go to Asia much anymore so I'm feeling less defensive about the extra network request, haha.

This would certainly be a great addition for the next release of hub!

PR #1010, opened 2015-10-13
unawkward the work header a bit

muhmuhten:
the other two headers ("As a ...") are attributive, that one was predicative, which was kind of inconsistent and also the "one who" construction just seemed off.


mislav:
Agreed!

Issue #876, opened 2015-04-22
Hub remote parsing is incomplete


frozencemetery:
For instance:

```
rharwood@thriss:~/python-gssapi$ git remote -v
frozencemetery	git+ssh://git@github.com/frozencemetery/python-gssapi (fetch)
frozencemetery	git+ssh://git@github.com/frozencemetery/python-gssapi (push)
origin	git+ssh://git@github.com/pythongssapi/python-gssapi (fetch)
origin	git+ssh://git@github.com/pythongssapi/python-gssapi (push)
sross	git+ssh://git@github.com/DirectXMan12/python-gssapi (fetch)
sross	git+ssh://git@github.com/DirectXMan12/python-gssapi (push)
rharwood@thriss:~/python-gssapi$ git pull-request
Aborted: the origin remote doesn't point to a GitHub repository.
rharwood@thriss:~/python-gssapi$ git fetch origin
rharwood@thriss:~/python-gssapi$ echo $?
0
rharwood@thriss:~/python-gssapi$ 
```

Clearly origin does point to a Github repo, so something has gone wrong here.

mislav:
I don't think we recognize the `git+ssh` scheme. How did you end up with it, as opposed to `git@github.com:frozencemetery/python-gssapi` syntax? If it's a valid scheme, we should definitely support it. In the meantime, it will help to change the URL syntax.

frozencemetery:
It's a valid scheme as per [git source](https://github.com/git/git/blob/master/transport.c#L963).  I tend to use it because I don't want to keep track of the implicit part of, e.g., `git@github.com:frozencemetery/python-gssapi`, which is admittedly not the best reason, but it is accepted by get.

retronym:
Perhaps relatedly, I recently upgraded to `hub` 2.2.1 and found that it no longer seems to handle my remote configuration:

```
% git remote -v | grep origin
origin	git://github.com/scala/scala.git (fetch)
origin	no_push_for_you (push)
ticket/9285 ~/code/scala3 git pull-request -b 2.11.x
Aborted: the origin remote doesn't point to a GitHub repository.
```

I use separate pull/push URLs in this way to ensure that changes are only made through pull requests.

hakatashi:
I'm also experiencing the issue when I'm explicitly pointing port number for ssh

```
[~/dotfiles (master)]$ git remote -v
origin  ssh://git@github.com:22/hakatashi/dotfiles.git (fetch)
[...]

PR #1019, opened 2015-10-29
Fix pull-request command for fetch only remotes.

madhusudancs:
There are no tests in this PR yet. I am working on them. But I wanted to get the code review going and see if any of you have any opinions about this change. Please take a look.

Fixes one part of issue #876.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -176,7 +176,7 @@ func Remotes() (map[string]string, error) {
 			return nil, fmt.Errorf("couldn't get url for remote %s: %v", name, err)
 		}
 		if len(out) != 1 {
-			return nil, fmt.Errorf("")
+			return nil, fmt.Errorf("unexpected number of remote urls %s", out)
 		}
 		remotes[name] = out[0]
 	}
--- a/git/git.go
+++ b/git/git.go
@@ -163,8 +163,24 @@ func Log(sha1, sha2 string) (string, error) {
 	return outputs, nil
 }
 
-func Remotes() ([]string, error) {
-	return gitOutput("remote", "-v")
+func Remotes() (map[string]string, error) {
+	names, err := gitOutput("remote")
+	if err != nil {
+		return nil, fmt.Errorf("couldn't list remote names: %v", err)
+	}
+
+	remotes := make(map[string]string)
+	for _, name := range names {
+		out, err := gitOutput("ls-remote", "--get-url", name)
+		if err != nil {
+			return nil, fmt.Errorf("couldn't get url for remote %s: %v", name, err)
+		}
+		if len(out) != 1 {
+			return nil, fmt.Errorf("")
+		}
+		remotes[name] = out[0]
+	}
+	return remotes, nil
 }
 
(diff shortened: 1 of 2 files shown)


git/git.go
```diff
@@ -163,8 +163,24 @@ func Log(sha1, sha2 string) (string, error) {
...
+		return nil, fmt.Errorf("couldn't list remote names: %v", err)
+	}
+
+	remotes := make(map[string]string)
+	for _, name := range names {
+		out, err := gitOutput("ls-remote", "--get-url", name)
+		if err != nil {
+			return nil, fmt.Errorf("couldn't get url for remote %s: %v", name, err)
+		}
+		if len(out) != 1 {
```
mislav: What is this check for?

github/remote.go
```diff
@@ -27,25 +25,12 @@ func (remote *Remote) Project() (*Project, error) {
 }
 
 func Remotes() (remotes []Remote, err error) {
-	re := regexp.MustCompile(`(.+)\s+(.+)\s+\((push|fetch)\)`)
-
-	rs, err := git.Remotes()
+	remotesMap, err := git.Remotes()
```
mislav: Nice simplification 👍

mislav:
Thanks for starting this conversation! Few notes:

1. What is the earliest version where `ls-remote --get-url` is available?
2. Is there a way to read `pushUrl` as well, but fallback to `url` if the former is not a valid URL?

git/git.go
madhusudancs: Sorry, I wanted to add an error message for the return which would have made the intention here clear, but I got distracted.

This is an assertion or a defense to ensure that git-ls-remote --get-url always returns one and only one URL, in the unlikely event that this changes.

madhusudancs:
> 1. What is the earliest version where ls-remote --get-url is available?

It looks like it was added in 1.7.5 -https://github.com/git/git/commit/45781adb9a89c0c47f61ccf8062be26b86a38a54 but the documentation did not appear until 1.7.12.2 -https://github.com/git/git/commit/2303cad242a6ccc4745d7994ecc0425bd8513a0d

> 2. Is there a way to read pushUrl as well, but fallback to url if the former is not a valid URL?

I don't think it is possible to do that with ls-remote --get-url, unless there is a trick that I am unaware of. But if we want to do that we might have to end up parsing the output that git remote produces.

mislav:
Thanks for the detective work. We don't have a defined requirement for minimum git version yet, but I'm worried about requiring 1.7.5 because that might be too high, and the payoff isn't really big.

Could you instead take a stab at a separate PR that keeps reading `git remote -v`, but ignores `pushUrl`s that are invalid URLs (some people use an invalid `pushUrl` to prevent accidental pushes)?

madhusudancs:
> Could you instead take a stab at a separate PR that keeps reading git remote -v, but ignores pushUrls that are invalid URLs (some people use an invalid pushUrl to prevent accidental pushes)?

Sure.

mislav:
Closing in favor of your new PR, thanks!

PR #1025, opened 2015-11-03
Another attempt at fixing pull-request command for fetch-only remotes.

madhusudancs:
Please see PR #1019 for previous discussion.

### DIFF
--- a/features/pull_request.feature
+++ b/features/pull_request.feature
@@ -643,3 +643,19 @@ Feature: hub pull-request
       """
     When I successfully run `hub pull-request -m hereyougo -l feature,release`
     Then the output should contain exactly "the://url\n"
+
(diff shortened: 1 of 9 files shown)


mislav:
Thanks for working on this. I love the added tests.

I'm glad that the problem is solved, but I'm sad that this makes the remote parsing logic even more complex. I'm not happy with the current state of the code that analyzes state of git remotes, branches, and GitHub projects. I wish that in the future we could refactor the code to make all this simpler.

You will need to accept the CLA in order for this PR to get accepted. See CI results for this PR.

madhusudancs:
Thanks. I agree that this makes the parsing code even more complex, but sadly I did not find a better way.

I have accepted the CLA now. Please accept the PR.

PR #1031, opened 2015-11-14
Update redirects in Readme

ReadmeCritic:
Created with https://github.com/dkhamsing/frankenstein

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ in your executable path.
 
 #### Source
 
-To install `hub` from source, you need to have a [Go development environment](http://golang.org/doc/install),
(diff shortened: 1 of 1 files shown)


stevemao:
:+1:

README.md
```diff
@@ -138,7 +138,7 @@ superpowers:
 
 ### git am, git apply
 
-    $ git am https://github.com/defunkt/hub/pull/55
```
mislav: Sorry, I don't want to change these examples. These don't have to be real URLs, and having "github" twice in the URL might be confusing, even though that's the latest name of this repo. I might change the name of the example repo in the future.

Issue #1036, opened 2015-11-24
hub fork -- set upstream


simplyianm:
I think doing a rename of origin to upstream and creating a new remote as origin is much better than the current behavior (create new remote with name as username)

mislav:
See https://github.com/github/hub/pull/798

PR #1038, opened 2015-11-27
Alter fork behavior to rename "origin" to "upstream".

simplyianm:
This makes a lot more sense in my opinion. Related to #1036.

### DIFF
--- a/commands/fork.go
+++ b/commands/fork.go
@@ -11,10 +11,8 @@ import (
 var cmdFork = &Command{
 	Run:   fork,
 	Usage: "fork [--no-remote]",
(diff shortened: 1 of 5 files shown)


simplyianm:
I accepted the CLA. Not sure how that will get updated.

mislav:
Hi, thanks for your contribution. The PR looks alright, but this was already attempted in https://github.com/github/hub/pull/798, and not yet accepted since it doesn't keep backwards compatibility with how hub worked before. If this feature was added, it needs to include some sort of configurability option so that people can choose to explicitly opt into new behavior.

For updates, please follow the discussion around the original PR. Thanks!

PR #1044, opened 2015-12-03
commands/fork: handle missing "origin" remote

### LINKED ISSUE #1042: Crash report - runtime error: invalid memory address or nil pointer dereference
Error (runtime.errorString): `runtime error: invalid memory address or nil pointer dereference`

Stack:

```
goroutine 1 [running]:
github.com/github/hub/commands.fork(0x721e80, 0xc8200a6240)
	/var/folders/08/6djc5fjs7gq20zv_1gx0_c3w0000gn/T/go/src/github.com/github/hub/commands/fork.go:75 +0x396
github.com/github/hub/commands.(*Command).Call(0x721e80, 0xc8200a6240, 0x0, 0x0)
	/var/folders/08/6djc5fjs7gq20zv_1gx0_c3w0000gn/T/go/src/github.com/github/hub/commands/commands.go:47 +0x188
github.com/github/hub/commands.(*Runner).Call(0xc820065920, 0x721e80, 0xc8200a6240, 0x0, 0x0, 0x0)
	/var/folders/08/6djc5fjs7gq20zv_1gx0_c3w0000gn/T/go/src/github.com/github/hub/commands/runner.go:93 +0x5a
github.com/github/hub/commands.(*Runner).Execute(0xc820065920, 0x0, 0x0, 0x0)
	/var/folders/08/6djc5fjs7gq20zv_1gx0_c3w0000gn/T/go/src/github.com/github/hub/commands/runner.go:85 +0x367
main.main()
	/private/tmp/hub20151119-59033-n5wdgh/main.go:12 +0x40

```

Runtime:

```
GOOS: darwin
GOARCH: amd64
```

Version:

```
git version 2.3.2 (Apple Git-55)
hub version 2.2.0-g1202fc4
```

parkr:
Fixes nil reference crash. Fixes #1042.

### DIFF
--- a/features/fork.feature
+++ b/features/fork.feature
@@ -113,11 +113,7 @@ Scenario: Related fork already exists
     And the stderr should contain "fatal: Not a git repository"
 
   Scenario: Origin remote doesn't exist
-    Given the GitHub API server:
-      """
-      post('/repos/mislav/dotfiles/forks') { '' }
-      """
-    And I run `git remote rm origin`
+    Given I run `git remote rm origin`
     And the "mislav" remote has url "https://github.com/mislav/dotfiles.git"
     When I run `hub fork`
     Then the exit status should be 1
--- a/features/fork.feature
+++ b/features/fork.feature
@@ -112,6 +112,21 @@ Scenario: Related fork already exists
     Then the exit status should be 1
     And the stderr should contain "fatal: Not a git repository"
 
+  Scenario: Origin remote doesn't exist
+    Given the GitHub API server:
+      """
+      post('/repos/mislav/dotfiles/forks') { '' }
+      """
+    And I run `git remote rm origin`
+    And the "mislav" remote has url "https://github.com/mislav/dotfiles.git"
+    When I run `hub fork`
+    Then the exit status should be 1
+    And the stderr should contain exactly:
+      """
+      Error creating fork: No git remote with name origin\n
+      """
+    And there should be no "origin" remote
+
   Scenario: Unknown host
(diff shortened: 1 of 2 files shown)


mislav:
Thanks for fixing! That should take care of the crash. But, I'm confused about the functionality of this command right now.

If the command aborted and exited with status `1`, was forking successful or not? It seems like it was, but then we abort just prior to final `git remote add`. What did we fork from if there was no origin remote to start from?

jingweno:
Good point @mislav. @parkr, it's better we check the error before [all the ceremonies](https://github.com/parkr/hub/blob/handle-missing-origin-remote/commands/fork.go#L51-L53).

parkr:
> What did we fork from if there was no origin remote to start from?

@mislav From what I understand from the code, the fork is performed by the remote returned by `localRepo.MainProject()`, which appears to just pull in the first remote listed by git.

> it'd be better we check the error before all the ceremonies.

@jingweno Updated.

features/fork.feature
```diff
@@ -112,6 +112,21 @@ Scenario: Related fork already exists
     Then the exit status should be 1
     And the stderr should contain "fatal: Not a git repository"
 
+  Scenario: Origin remote doesn't exist
+    Given the GitHub API server:
+      """
+      post('/repos/mislav/dotfiles/forks') { '' }
```
mislav: Do we need the API server?

parkr: No, we do not. I have removed it.

mislav:
Thanks @parkr!! :tada:

parkr:
My pleasure. Apologies for all the back and forth. Looking forward to the next release! :sparkles:

PR #1052, opened 2015-12-09
Added `upstream` option to `browse`

RichardLitt:
I constantly want to be able to go directly to the remote upstream webpage for
the repo I am currently working on. This normally enters my workflow when I am
working on a fork, push to a feature branch, and want to PR to the remote repo
upstream. I would run `gh browse`, then click on the upstream link on the website,
then start the PR process. Having `gh browse upstream` will make this much easier.

This can be extended to deal with other remotes as well, perhaps using a subpage
like system: `git browse remote upstream`. I am happy to do this change, as well,
but figured a smaller fix like this would open discussion easier.

### DIFF
--- a/commands/browse.go
+++ b/commands/browse.go
@@ -47,6 +47,9 @@ func init() {
 
   $ gh browse gh wiki
   > open https://github.com/YOUR_LOGIN/gh/wiki
+
+  $ gh browse upstream
+  > open https://github.com/UPSTREAM_REMOTE_REPO
 */
 func browse(command *Command, args *Args) {
 	var (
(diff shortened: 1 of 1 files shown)


sindresorhus:
:+1:

mislav:
Hi, thanks for the contribution.

I'm not sure exactly what you're referring to by saying "upstream", but I understand your process of opening pull requests. The `browse` command was supposed to open the representation of current branch as it was pushed to a remote that's your fork on GitHub, but this functionality was broken. It got fixed in a77104fc395597842f9178b74688e1a985365519, but it wasn't released yet. It will be soon.

The process of opening a PR from web will then be:

```sh
git push -u <my-remote> feature
hub browse
# when branch opens in browser, press Pull Request button.
```

There shouldn't be a need for a subcommand that only works with remotes literally named "upstream". Some people name their upstream remotes as "origin" or "github".

RichardLitt:
In that case, it might be smart to enable `hub browse <remote>`, because this is a use case that would be highly useful.

mislav:
In my opinion, automatically detecting which remote has the current branch been pushed to and opening that branch on that remote should be enough. Do you have a use-case for opening an explicit remote? Which branch would be opened on that remote if you specify it via argument?

penman:
I have a use-case for this:

I work on my own fork of a shared repository. My fork is called `origin` and the shared repository is called `upstream`. I then make pull requests from `origin` to `upstream`.

I use `git browse` for opening the repository in the browser so I can look at issues and stuff. But, currently, this takes me to my fork, which doesn't have any issues or pull requests, and I have to click on the "forked from" link before I get to where I want.

I'd like to be able to `git browse upstream` and be taken to the upstream repository on GitHub (probably the default branch unless otherwise specified).

PR #1054, opened 2015-12-13
Don't corrupt the PATH if longer than 1024 characters on Windows

jamesqo:
setx automatically truncates the PATH to 1024 characters if it is longer than that, which is no good. It also mucks up the local user's PATH with the machine-wide PATH.

This PR fixes that by using PowerShell to manually alter the PATH.

### DIFF
--- a/script/install.bat
+++ b/script/install.bat
@@ -2,14 +2,22 @@
 CLS
 goto checkPrivileges
 
-:appendToPath
-set "RUNPS=powershell -NoProfile -ExecutionPolicy Bypass -Command"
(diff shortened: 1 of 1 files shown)


mislav:
Thanks for the fix! Unfortunately, this new code is quote hard to interpret as a human. Is it possible to try to clean up the code, or at least document it? I'm reluctant to pull in code that I don't know what it does.

jamesqo:
Sure, no problem. Thanks for responding :smile:

mislav:
Thanks for adding docs! Have you tested this manually?

jamesqo:
Just tested, it seems to be working after I made a few changes:

image

mislav:
Thanks!

PR #1060, opened 2015-12-27
Add another commmand for alias

laithshadeed:


mislav:
Thanks for the recommendation, but we won't add this since it does exactly the same thing as eval'ing `hub alias -s`. If there's a bug with `hub alias`, let's work on fixing that instead of tweaking the alternatives in the documentation.

PR #1061, opened 2015-12-27
Use XDG Base Directory Spec for default config file. 

### LINKED ISSUE #1048: respect the XDG_CONFIG_HOME environment variable
Thank you for using `~/.config/hub` for your configuration file, as opposed to polluting the base home directory with yet another top-level dot file (`~/.hub`).

It would be even better if this configuration file was relocatable.  This is covered by the [XDG Base Directory Specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html).  Basically, this means to check for the environment variable `XDG_CONFIG_HOME`.  If it exists, make your configuration file relative to that (`${XDG_CONFIG_HOME}/hub`).  If it does not exist, make your configuration file relative to `~/.config` (matching the existing behavior).  Because you are already using the default relative path, it would be a backwards compatible change for existing users.

pabryan:
Fixes #1048. Note that the dependency cep21/xdgbasedir referenced in Godeps breaks the tests, though the committed version in _workspaces passes. Once cep21/xdgbasedir#1 is resolved upstream, then everything will be fine.


pabryan:
I don't know why the Travis build failed. The Travis build https://travis-ci.org/pabryan/hub/jobs/99089471 worked fine after committing to my fork pabryan/hub/@4c612cfa4

pabryan:
I've added a write test now. It's a little bit of a kludge since it uses authenticates first because as far as I know, the only place hub writes to the config file is when authenticating: if the authentication code breaks this test will fail even if the xdgbasedir code is working. I suppose I should also add a read config test too.

github/config.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/github/hub/Godeps/_workspace/src/github.com/howeyc/gopass"
 	"github.com/github/hub/ui"
 	"github.com/github/hub/utils"
+	"github.com/github/hub/Godeps/_workspace/src/github.com/cep21/xdgbasedir"
```
mislav: Instead of pulling in a new dependency, couldn't we simply read the value of `XDG_CONFIG_HOME` environment variable and default to `$HOME/.config` if not set?

pabryan: Sure. I'll do it later today if that's preferred. I just figured that it might be nice to have the xdgbasedir spec available in it's entirety, and why not pull it in from a library?. For instance there is an argument that passwords should go in data not in config, and the dependency has code for data directories as well. Hub config would go in the config directory. If a user wants to cache their password, but not save it permanently that could go in the cache directory etc.

README.md
```diff
@@ -279,16 +279,26 @@ superpowers:
 Configuration
 -------------
 
+### Configuration file
+
+Hub stores its configuration in a file named "hub", located in a directory
+determined in accordance with the 
+[XDG Base Directory Specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html).
```
mislav: I don't think this section is at all necessary. We can just mention somewhere that hub respects XDG_CONFIG_HOME, and that should be enough for people familiar with XDG to understand what's going on. We can keep referring to `~/.config/hub` in the rest of the documentation as before.

github/config.go
mislav: My general rule for software is not to pull in dependencies when you can simply inline something. This is the entirety of the logic that we're obtaining by pulling in an external library:

```go
// DataHomeDirectory is a single base directory relative to which user-specific data files should be written
func DataHomeDirectory() (string, error) {
	return getInEnvOrJoinWithHome("XDG_DATA_HOME", defaultDataHomeDirectory)
}
```

we can do the same ourselves. Also, this library uses `user.Current()`, which we want to avoid due to cross-compilation issues: https://github.com/github/hub/pull/1082

README.md
pabryan: Sorry, I just moved house and have had intermittent internet access. I thought I'd get a chance to do this before I moved, but it didn't happen. Lucky this is not very important :smile: I'll sort it out when I get a chance over the next couple of days.

mislav:
Due to some changes in master, this PR will now heavily conflict at merge. @pabryan do you want me to take this over, since you're busy at the moment? I can sort it out quickly.

pabryan:
Okay, thanks.

PR #1068, opened 2016-01-05
Add --id option to issue command

saidie:
This pull request adds `--id` (`-i`) option to the `issue` command. If you set an issue number to this option, you can get that issue information.
```
% hub issue --id=1
      1] hub pages ( https://github.com/github/hub/issues/1 )
```
This would be useful to get an issue title and URL when only an issue number is given.

### DIFF
--- a/commands/issue.go
+++ b/commands/issue.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"fmt"
 
+	"github.com/github/hub/Godeps/_workspace/src/github.com/octokit/go-octokit/octokit"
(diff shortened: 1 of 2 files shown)


mislav:
Thanks for the suggestion, but the `issue` command is half-baked and only in beta until we decide what its API and functionality is going to look like. For now, the `--id` flag is not very useful since the only piece of information you get about the issue is its title (the URL you could have constructed using the ID).

In the future specifying an issue will probably be supported, but for the purpose of editing it (e.g. adding labels).

PR #1077, opened 2016-01-19
rename gh command to hub in docs

kbrock:
followup to #1073 where I was reminded that `gh` is now called `hub`

took the liberty to change docs in all files

/cc @jingweno @mislav thanks


mislav:
Thanks!

PR #1078, opened 2016-01-19
Clarify "Unknown Shell" Error Message

### LINKED ISSUE #1065: `hub alias` on Windows 10, git bash shell reports "Unknown Shell"
On Windows 10, running the git bash shell, installed hub and attempted to use:
`eval $(hub alias -s)` command

I get the response `Unknown shell`.

Bash Version:
```
$ bash --version
GNU bash, version 3.1.23(6)-release (i686-pc-msys)
Copyright (C) 2005 Free Software Foundation, Inc.
```

I don't know what that command is supposed to output, but I replaced it with `alias git=hub`, which appears to work.

Regardless, not clear why the above is not working.

SaintPeter:
In instances where the `SHELL` environment variable is not set, nor is the `-s` parameter set it is not clear what action should occur.  Clarify the error message to make it obvious.

Closes #1065

Untested.

### DIFF
--- a/commands/alias.go
+++ b/commands/alias.go
@@ -36,7 +36,7 @@ func alias(command *Command, args *Args) {
 	}
 
 	if shell == "" {
-		utils.Check(fmt.Errorf("Unknown shell"))
+		utils.Check(fmt.Errorf("Error: couldn't detect shell type. Please specify your shell with 'hub alias <shell>'"))
 	}
 
 	shells := []string{"bash", "zsh", "sh", "ksh", "csh", "tcsh", "fish"}


commands/alias.go
```diff
@@ -36,7 +36,7 @@ func alias(command *Command, args *Args) {
 	}
 
 	if shell == "" {
-		utils.Check(fmt.Errorf("Unknown shell"))
+		utils.Check(fmt.Errorf("Unknown shell: Set the SHELL enviornment variable or execute 'hub alies -s <shell name>'"))
```
parkr: Did you mean `hub alias`, perhaps?

SaintPeter:
@parkr Yes, corrected typo.

commands/alias.go
```diff
@@ -36,7 +36,7 @@ func alias(command *Command, args *Args) {
 	}
 
 	if shell == "" {
-		utils.Check(fmt.Errorf("Unknown shell"))
+		utils.Check(fmt.Errorf("Unknown shell: Set the SHELL enviornment variable or execute 'hub alias -s <shell name>'"))
```
mislav: Let's not suggest that people set the SHELL environment variable, and instead let's just suggest that they explicitly pass in the shell type as argument.

How about if the error message said:

> Error: couldn't detect shell type. Please specify your shell with `hub alias -s <shell>`

Also, note that this error message can be shown when `hub alias` was called without the `-s` flag. In that case, the error messsage *shouldn't* suggest that you execute `hub alias -s`. Instead:

> Error: couldn't detect shell type. Please specify your shell with `hub alias <shell>`

SaintPeter: What does the `-s` do?  Is it the same as `hub alias <shell>`?

mislav: No, you run `hub alias` shell to get the instructions, and `hub alias -s` to get the code output suitable for eval'ing. Basically, `-s` opts into eval mode and isn't related to `<shell>` argument at all.

SaintPeter:
Updated error message as per @mislav's suggestion.

mislav:
Thanks for that!

PR #1079, opened 2016-01-20
Switch from Godep to GO15VENDOREXPERIMENT

mislav:
Move all dependencies from `Godeps/_workspace/src` to `vendor/` directory.

Pros:

* No more path rewriting
* Standard Go toolchain

Cons:

* Jumped through hoops to get `go test` and `fmt` working again
* Tooling to manage dependencies still unclear

/cc @jingweno


jingweno:
AFAIK @freeformz is making Godep to work with 1.5's GO15VENDOREXPERIMENT.

freeformz:
godep works" with the GO15VENDOREXPERIMENT and writes stuff to `vendor/` instead of the workspace, although there are some rough edges (things still need to be in the $GOPATH first to be saved). But over the next month or so most of those should be worked out.

PS: You should still use a tool to manage the deps IMO regardless of `vendor/`. I suggest either godep || gb || govendor.

freeformz:
See also https://github.com/tools/godep#go-15-vendor-experiment

mislav:
Thanks for the tips!

PR #1087, opened 2016-01-24
Split help into individual man pages

mislav:
* The canonical version of documentation for each command is now maintained inside the `*.go` implementation file for that command. Duplicate documentation is removed from README and the main `hub(1)` man page.

* Add `script/man` that generates individual man pages for each command. Generated pages are not kept in version control; instead, they will be bundled with each release.

* `hub help pull-request` will now show `hub-pull-request(1)` man page on systems that support `man`. On Windows, plain text help will be displayed.

* `hub help hub-clone` (note the `hub-` prefix) now looks up `hub-clone(1)` man page. As before, `hub help clone` displays ordinary `git-clone(1)` documentation for compatibility with git.

* The HTML representation of man pages now has cross-reference links:

  screen shot 2016-01-25 at 1 59 54 am


mislav:
New man pages are gorgeous! :sparkles: :lipstick: 

screen shot 2016-01-25 at 2 32 58 am

PR #1088, opened 2016-01-25
Fix bash_completion double reading bug

kawaz:
```console
$ # When first read, it is no ploblem.
$ . hub/etc/hub.bash_completion.sh
$ hub [tab][tab]
add                  describe             pull-request
alias                diff                 push
am                   difftool             rebase
:
$ # When read again, it is problem.
$ . hub/etc/hub.bash_completion.sh
$ hub [tab]-bash: __git_list_all_commands_without_hub_without_hub: command not found
```

This pull request correct the problem.

### DIFF
--- a/etc/hub.bash_completion.sh
+++ b/etc/hub.bash_completion.sh
@@ -4,12 +4,13 @@
 # Check that git tab completion is available
 if declare -F _git > /dev/null; then
   # Duplicate and rename the 'list_all_commands' function
-  eval "$(declare -f __git_list_all_commands | \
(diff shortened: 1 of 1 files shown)


mislav:
Thanks for the suggestion? But, when is bash completion script read twice?

kawaz:
I'm verry sorry. I found the problem in my bashrc.
I read a local copy of hub.bash_completion.sh, and one more it installed by brew too.

mislav:
Ok, so if it was a problem with your bashrc, I'm going to close this out.

PR #1090, opened 2016-01-26
Adding support for arm processors

IWriteThings:
Updating the cross compile script here to add support for arm processors.

### DIFF
--- a/script/cross-compile
+++ b/script/cross-compile
@@ -13,6 +13,8 @@ echo '
   freebsd  amd64    FreeBSD 64-bit
   linux    386      Linux 32-bit
   linux    amd64    Linux 64-bit
+  linux    arm      Linux ARM 32-bit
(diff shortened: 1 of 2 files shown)


mislav:
This looks good! Have you verified that a non-ARM linux machine can cross-compile ARM binaries? I couldn't on OS X, and I'm wondering why.

IWriteThings:
@mislav ,
I was able to  cross-compile the files in arm and arm64. Can you pull my latest commit and try it on your box? I am on Mac OSX 10.10.

Steps I took -

1. 
```
$ ./script/cross-compile comp
```

Do you have go > 1.5 ?
What errors are you getting?

mislav:
Ah, your 2nd commit was key :) thanks!

mislav:
The FreeBSD failure is known, unfortunately. We don't know what's causing it.

PR #1096, opened 2016-02-01
Filter project issues by assignee

cornwe19:
* Supports displaying project issues and pull requests assigned to a particular user

### DIFF
--- a/commands/issue.go
+++ b/commands/issue.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/github/hub/github"
 	"github.com/github/hub/ui"
@@ -75,7 +76,8 @@ func issue(cmd *Command, args *Args) {
 					url = issue.HTMLURL
 				}
 
-				if flagIssueAssignee == "" || issue.Assignee.Login == flagIssueAssignee {
+				if flagIssueAssignee == "" ||
+					strings.EqualFold(issue.Assignee.Login, flagIssueAssignee) {
 					// "nobody" should have more than 1 million github issues
 					ui.Printf("% 7d] %s ( %s )\n", issue.Number, issue.Title, url)
 				}
--- a/commands/issue.go
+++ b/commands/issue.go
@@ -33,6 +33,7 @@ var (
 `,
 	}
 
+	flagIssueAssignee,
 	flagIssueMessage,
 	flagIssueFile string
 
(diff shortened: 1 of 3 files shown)


mislav:
Looks good! You will have to accept the CLA. The link is among the status checks for this PR.

cornwe19:
Thanks! I just accepted the CLA

commands/issue.go
```diff
@@ -66,8 +74,11 @@ func issue(cmd *Command, args *Args) {
 				} else {
 					url = issue.HTMLURL
 				}
-				// "nobody" should have more than 1 million github issues
-				ui.Printf("% 7d] %s ( %s )\n", issue.Number, issue.Title, url)
+
+				if flagIssueAssignee == "" || issue.Assignee.Login == flagIssueAssignee {
```
mislav: Can you make it so that the comparison is case-insensitive?

features/steps.rb
```diff
@@ -82,6 +82,15 @@
   run_silent %(git reset --quiet --hard HEAD^)
 end
 
+Given(/^the remote issues of "(.*?)" are:$/) do |proj, json_value|
```
mislav: I don't feel there's a need for this custom step yet. How about simply using `Given the GitHub API server` directly?

mislav:
After a few minor nitpicks this will be ready to merge! Thank you for the exceptionally high quality pull request.

One more open question: should there be a way to list all issues assigned to yourself, without having to type your own login name? It would be nice if we could do `-a me` or something, but unfortunately a user with login name "me" exists.

I plan to overhaul the `hub issue` command soon just as I did with `release` #1095, but this will make a great feature to keep.

cornwe19:
Thanks for the feedback! I'm totally on board with the idea of short hand to find issues assigned to 'me' I just couldn't think of an intuitive way to include it off the top of my head. That's actually the way I'd most commonly use this feature.

I'll give it some more thought and fix up the suggestions from your other comments in the mean time.

cornwe19:
I think I've addressed your concerns with the code.

As for the suggestion about referencing yourself automatically, it might be useful to try and interpret `-a` with no assignee login given as the current user. I'm not sure if that's the most intuitive implementation or if its even possible for the pflag library to treat a flag like both a string value and a simple boolean at the same time.

Regardless, if someone wanted that functionality, they could also create an alias like the following:
```
git config --global alias.myissues 'issue -a cornwe19'
```

mislav:
Thanks for your patience!

mislav:
> it might be useful to try and interpret `-a` with no assignee login given as the current user

No that's not possible. A flag should either accept a value or not accept a value, but it can't be both. The only exception are boolean flags which usually don't take a value (e.g. `-p`) but can optionally take a value using equal signs e.g. `-p=false`.

cornwe19:
Makes sense. I got the impression that wouldn't work out. Thanks for confirming and for merging the change through.

PR #1099, opened 2016-02-03
Revamp `hub issue` commands

mislav:
This marks `hub issue` family of commands coming out of beta.

* `hub issue` cleaner output
  * doesn't include pull requests
  * red/green color based on closed/open state
  * list both open and closed issues with `hub issue -s all`

  screen shot 2016-02-03 at 11 32 53 pm
* All commands reimplemented using `simpleApi` and not go-octokit
* `hub issue create` small tweaks for compatibility with `hub release create`

Todo:

* [x] ~~Implement other `hub issue` [filtering options](https://developer.github.com/v3/issues/#list-issues-for-a-repository)~~
  * [x] Add `hub issue --format=...` option to choose which fields to output a la `git log`
* [x] List all pages of results from API (currently consuming just 1st page)
  * [x] ~~Cache previously fetched pages locally~~
* [x] ~~`hub issue edit` command to change labels, assignee, closed/open state~~
  * [x] ~~`hub issue comment` command to add a comment to the thread?~~
* [x] ~~Ability to list all your issues a la Issues Dashboard, i.e. not project-specific?~~
* [x] ~~Flag to specify a repository to manage issues for, e.g. if not in a git working directory?~~


commands/issue.go
```diff
@@ -39,7 +39,15 @@ var (
 `,
 	}
 
+	cmdCreateIssue = &Command{
+		Key:   "create",
+		Run:   createIssue,
+		Usage: "issue create [-m <MESSAGE>|-f <FILE>] [-l <LABELS>]",
```
pcorpet: I believe Usage and Long of subcommand are never used. Actually putting something in there can be misleading.

commands/issue.go
```diff
@@ -48,80 +56,146 @@ var (
...
-			utils.Check(err)
-			for _, issue := range issues {
-				var url string
-				// use the pull request URL if we have one
-				if issue.PullRequest.HTMLURL != "" {
-					url = issue.PullRequest.HTMLURL
-				} else {
-					url = issue.HTMLURL
-				}
+func listIssues(cmd *Command, args *Args) {
```
pcorpet: This is a quite long function, consider splitting something out in another method. This would help a lot for documentation and understanding it

commands/issue.go
```diff
@@ -48,80 +56,146 @@ var (
...
+				ui.Printf(" \033[38;5;%d;48;2;%d;%d;%dm %s \033[m", textColor, color.Red, color.Green, color.Blue, label.Name)
+			}
 
-			ui.Println(issue.HTMLURL)
+			ui.Printf("\n")
 		}
-	})
+	}
+
+	os.Exit(0)
```
pcorpet: That seems kind of extreme. Why do you need to do that? It seems that the framework should be able to understand a command doesn't need further processing somehow.

commands/issue.go
```diff
@@ -48,80 +56,146 @@ var (
...
+	project, err := localRepo.MainProject()
+	utils.Check(err)
+
+	gh := github.NewClient(project.Host)
+
+	var title string
+	var body string
+	var editor *github.Editor
+
+	if cmd.FlagPassed("message") {
```
pcorpet: The pattern is not very clean here: the fact that flagIssueMessage is linked to "message" is encoded in two different places. I would just check whether flagIssueMessage is empty or not. The only difference would be if someone actually sets the flag to the empty string which doesn't make much sense.

Same for file.

commands/issue.go
```diff
@@ -48,80 +56,146 @@ var (
...
+
+		title, body, err = editor.EditTitleAndBody()
+		utils.Check(err)
+	}
 
-	editor, err := github.NewEditor("ISSUE", "issue", message)
-	if err != nil {
-		return "", "", err
+	if title == "" {
+		utils.Check(fmt.Errorf("Aborting creation due to empty issue title"))
```
pcorpet: You use a lot of this "fmt.Errorf" without any actual formatting. Is there a reason you prefer it to errors.New?

commands/issue.go
```diff
@@ -48,80 +56,146 @@ var (
...
 
-	defer editor.DeleteFile()
+	params := &github.IssueParams{
+		Title:  title,
+		Body:   body,
+		Labels: flagIssueLabels,
+	}
+
+	if args.Noop {
+		ui.Printf("Would create issue `%s' for %s\n", params.Title, project)
```
pcorpet: Is it worth to also add the body and the labels as well?

github/client.go
```diff
@@ -420,50 +420,79 @@ func (client *Client) ForkRepository(project *Project) (repo *octokit.Repository
 	return
 }
 
-func (client *Client) Issues(project *Project) (issues []octokit.Issue, err error) {
-	url, err := octokit.RepoIssuesURL.Expand(octokit.M{"owner": project.Owner, "repo": project.Name})
+type IssueParams struct {
```
pcorpet: Can we document those types? for instance if you have a reference to the structure that it's reading, that would be awesome: it would auto-document each field, and add a big hint to anybody wanting to extend them.

utils/utils.go
```diff
@@ -88,3 +89,34 @@ func DirName() (string, error) {
...
 	return strings.EqualFold(confirm, short) || strings.EqualFold(confirm, long)
 }
+
+type Color struct {
+	Red   int64
+	Green int64
+	Blue  int64
+}
+
+func NewColor(hex string) (*Color, error) {
```
pcorpet: The logic is not rocket science, but I guess it wouldn't hurt to add a test with some basic values and some basic errors.

github/client.go
```diff
@@ -420,50 +420,79 @@ func (client *Client) ForkRepository(project *Project) (repo *octokit.Repository
...
+	path := fmt.Sprintf("repos/%s/%s/issues", project.Owner, project.Name)
+	if filterParams != nil {
+		query := url.Values{}
+		for key, value := range filterParams {
+			switch v := value.(type) {
+			case string:
+				query.Add(key, v)
+			}
+		}
+		path += "?" + query.Encode()
```
pcorpet: It's a minor optimization but I guess that if filterParams contains only one value that is not a string, then query.Encode() would be empty.

commands/issue.go
mislav: `Usage` should be displayed in cases when you passed an invalid flag, and `Usage` and `Long` are both used for generating documentation for a command.

commands/issue.go
mislav: hub was originally designed to only wrap existing commands such as `git remote`. This, it's designed to allow you to change the original arguments to the invocation, but will default to eventually executing `git <command> [<args>...]` after the Go function has executed. Right now `os.Exit(0)` is a clunky way to opt out of that and say "no, the command actually ends here".

commands/issue.go
mislav: I'm not absolutely happy about this pattern as well, but it's the only way I could discern between the "this argument was passed as an empty string" scenario vs. "this argument wasn't passed".

Consider `git commit -m ""`. It detects that an empty message argument was passed and aborts execution. If you don't pass `-m` at all, however, it will open a text editor. Well, if we check `flagIssueMessage == ""`, this actually doesn't tell us whether any value has been passed at all. That's why I created `FlagPassed(...)` to answer that.

If you have ideas how to encapsulate this pattern better, I'm all ears.

commands/issue.go
mislav: No reason whatsoever. However, if the package already imports `fmt`, then I feel that `fmt.Errorf` is just as good as `errors.New`.

commands/issue.go
mislav: I don't feel that the `--noop` output should be absolutely detailed here. The `--noop` flag is mostly useful for other hub commands that result in a git command being executed because it allows you to inspect the final git command.

github/client.go
mislav: These structs map very closely to the JSON representation of these resources as returned from GH API. I'm open to document them better, but I'm not sure how. Do you have an idea of what you would like to see here?

github/client.go
mislav: Oh I think that's OK for now. But good eye catching that!

utils/utils.go
mislav: Agreed; will try to add rudimentary Go unit tests.

commands/issue.go
mislav: Agreed that this function is too long! I just wanted to get it working at first, but now I'll try to split it.

commands/issue.go
pcorpet: >Usage and Long are both used for generating documentation for a command.

That is not true today for subcommands: either open an issue to make `hub help issue create` display the help for the subcommand, or just remove the doc here.

commands/issue.go
pcorpet: > hub was originally designed to only wrap existing commands 

I understand that, however there seems to be more and more commands that are not only wrappers. My point was that we probably should make the framework evolve to account for those commands that are not wrappers. It could be something like a flag in the Command struct or a return value for the command function for instance.

commands/issue.go
mislav: Oops, you're right. For subcommands, these fields aren't consulted yet. For now, let's keep them up to date still, at least as some sort of "code" documentation, and we'll figure out how to hook subcommands into the global help system.

commands/issue.go
mislav: You're right. This isn't a huge code debt yet (since it's just a few identical lines like this throughout the codebase), but we should definitely solve this through CmdRunner system going forward.

commands/issue.go
pcorpet: > If you have ideas how to encapsulate this pattern better, I'm all ears.

It's too bad that the flag module that you are using is not returning the Flag object: that would ease everything. Let's keep it like this for now.

github/client.go
pcorpet: >  Do you have an idea of what you would like to see here?

A link to https://developer.github.com/v3/issues/#create-an-issue that contains all the documentation for those fields and also list the other fields that we don't use (yet) in hub like assignee and milestone.

krlmlr:
I can confirm that this PR also works with Ubuntu 15.10 (where hub 2.2.3 failed to retrieve anything). Sweet, thanks!

mislav:
@pcorpet I've pushed some extra features and will merge this to master as the first iteration of the new `issue` command. Additional upgrades should come in form of separate PRs.

* Add multiple assignees & milestone support
* Add `hub issue create --browse` flag
* `--format` placeholders renamed & expanded (see below)
* All pages of API results now fetched. We'll have to figure out some caching strategy.

```
%I: issue number

%i: issue number prefixed with "#"

%U: the URL of this issue

%S: state (i.e. "open", "closed")

%sC: set color to red or green, depending on issue state.

%t: title

%l: colored labels

%L: raw, comma-separated labels

%b: body

%au: login name of author

%as: comma-separated list of assignees

%Mn: milestone number

%Mt: milestone title

%NC: number of comments

%Nc: number of comments wrapped in parentheses, or blank string if zero.

%cD: created date-only (no time of day)

%cr: created date, relative

%ct: created date, UNIX timestamp

%cI: created date, ISO 8601 format

%uD: updated date-only (no time of day)

%ur: updated date, relative

%ut: updated date, UNIX timestamp

%uI: updated date, ISO 8601 format
```

pcorpet:
Cool, thanks. Do you have an idea of a feature / issue I could help with to get a new release of hub with the code currently in master? (I saw you created a 2.2.5 but it seems to cherry-pick only some small fixes).

mislav:
> Do you have an idea of a feature / issue I could help with to get a new release of hub with the code currently in master?

Your help is greatly appreciated. Here are my ideas for the next release: https://github.com/github/hub/issues/1232

Obviously, not everything will probably make it in. I'm leaving on vacation for 2 weeks next week, so by the end of this week I was thinking to try to push a prerelease with the stuff currently in master.

Issue #1102, opened 2016-02-05
Add an option to download releases.


glennpratt:
[feature request]
I see I can list releases now, but I cannot download them from private repositories with just the URLs.

```
~/bin/hub release show 0.0.0-gpratt4 --show-downloads


## Downloads

https://github.com/topsecretorg/foobar/releases/download/0.0.0-gpratt4/foobar-linux-3.19.0-30-generic-x86_64-0.0.0-gpratt4.tar.gz
https://api.github.com/repos/topsecretorg/foobar/zipball/0.0.0-gpratt4
https://api.github.com/repos/topsecretorg/foobar/tarball/0.0.0-gpratt4
```

glennpratt:
See #1103

glennpratt:
Merged in e046b8d

mislav:
@glennpratt Thanks for following up by closing this. In the future, you can use the phrase `Fixes/closes #123` in your pull request body. That will automatically close the referenced issue when the PR is merged! :wink:

PR #1103, opened 2016-02-05
Add `hub release download` command that downloads all assets.

glennpratt:
This is probably overly simplistic, it downloads all assets if there are any, writes them in the current directory and prints the names.

I was lazy and avoided touching octokit.

Addresses #1102

### DIFF
--- a/commands/release.go
+++ b/commands/release.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
@@ -46,6 +47,9 @@ With '--include-drafts', include draft releases in the listing.
 		pre-populated with current release title and body. To re-use existing title
 		and body unchanged, pass '-m ""'.
 
+	* _download_:
+	  Download the assets attached to release for the specified <TAG>.
+
 ## Options:
 	-d, --draft
 		Create a draft release.
@@ -64,7 +68,7 @@ With '--include-drafts', include draft releases in the listing.
 
 	-F, --file <FILE>
 		Read the release title and description from <FILE>.
-	
+
 	-c, --commitish <TARGET>
 		A SHA, tag, or branch name to attach the release to (default: current branch).
 
@@ -90,6 +94,11 @@ hub(1), git-tag(1)
 	cmdEditRelease = &Command{
 		Key: "edit",
 		Run: editRelease,
+	}
+
+	cmdDownloadRelease = &Command{
+		Key: "download",
+		Run: downloadRelease,
 	}
 
 	flagReleaseIncludeDrafts,
@@ -126,6 +135,7 @@ func init() {
 	cmdRelease.Use(cmdShowRelease)
 	cmdRelease.Use(cmdCreateRelease)
 	cmdRelease.Use(cmdEditRelease)
+	cmdRelease.Use(cmdDownloadRelease)
 	CmdRunner.Use(cmdRelease)
 }
 
@@ -193,6 +203,52 @@ func showRelease(cmd *Command, args *Args) {
 	}
 
 	os.Exit(0)
+}
+
+func downloadRelease(cmd *Command, args *Args) {
+	tagName := cmd.Arg(0)
+	if tagName == "" {
+		utils.Check(fmt.Errorf("Missing argument TAG"))
+	}
+
+	localRepo, err := github.LocalRepo()
+	utils.Check(err)
+
+	project, err := localRepo.MainProject()
+	utils.Check(err)
+
+	gh := github.NewClient(project.Host)
+
+	release, err := gh.FetchRelease(project, tagName)
+	utils.Check(err)
+
+	for _, asset := range release.Assets {
+		ui.Printf("Downloading %s ...\n", asset.Name)
+		err := downloadReleaseAsset(asset, gh)
+		utils.Check(err)
+	}
+
+	os.Exit(0)
+}
+
+func downloadReleaseAsset(asset github.ReleaseAsset, gh *github.Client) (err error) {
+	assetReader, err := gh.DownloadReleaseAsset(asset.ApiUrl)
+	if err != nil {
+		return
+	}
+	defer assetReader.Close()
+
+	assetFile, err := os.OpenFile(asset.Name, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0644)
+	if err != nil {
+		return
+	}
+	defer assetFile.Close()
+
+	_, err = io.Copy(assetFile, assetReader)
+	if err != nil {
+		return
+	}
+	return
 }
 
 func createRelease(cmd *Command, args *Args) {
--- a/features/release.feature
+++ b/features/release.feature
@@ -294,3 +294,46 @@ MARKDOWN
       """
       Attaching release asset `hello-1.2.0.tar.gz'...\n
       """
+
+    Scenario: Download a release asset.
+      Given the GitHub API server:
+        """
+        get('/repos/mislav/will_paginate/releases') {
+          json [
+            { url: 'https://api.github.com/repos/mislav/will_paginate/releases/123',
+              upload_url: 'https://api.github.com/uploads/assets{?name,label}',
+              tag_name: 'v1.2.0',
+              name: 'will_paginate 1.2.0',
+              draft: true,
+              prerelease: false,
+              assets: [
+                { url: 'https://api.github.com/repos/mislav/will_paginate/assets/9876',
+                  name: 'hello-1.2.0.tar.gz',
+                },
+              ],
+            },
+          ]
+        }
+        get('/repos/mislav/will_paginate/assets/9876') {
+          halt 401 unless request.env['HTTP_AUTHORIZATION'] == 'token OTOKEN'
+          halt 415 unless request.accept?('application/octet-stream')
+          status 302
+          headers['Location'] = 'https://github-cloud.s3.amazonaws.com/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz'
+          ""
+        }
+        get('/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz', :host_name => 'github-cloud.s3.amazonaws.com') {
+          halt 400 unless request.env['HTTP_AUTHORIZATION'].nil?
+          halt 415 unless request.accept?('application/octet-stream')
+          headers['Content-Type'] = 'application/octet-stream'
+          "ASSET_TARBALL"
+        }
+        """
+        When I successfully run `hub release download v1.2.0`
+        Then the output should contain exactly:
+          """
+          Downloading hello-1.2.0.tar.gz ...\n
+          """
+        And the file "hello-1.2.0.tar.gz" should contain exactly:
+          """
+          ASSET_TARBALL
+          """
--- a/github/client.go
+++ b/github/client.go
@@ -371,6 +371,20 @@ func (client *Client) DeleteReleaseAsset(asset *ReleaseAsset) (err error) {
 	return
 }
 
+func (client *Client) DownloadReleaseAsset(url string) (asset io.ReadCloser, err error) {
+	api, err := client.simpleApi()
+	if err != nil {
+		return
+	}
+
+	resp, err := api.GetFile(url, "application/octet-stream")
+	if err = checkStatus(200, "downloading asset", resp, err); err != nil {
+		return
+	}
+
+	return resp.Body, err
+}
+
 type CIStatusResponse struct {
 	State    string     `json:"state"`
 	Statuses []CIStatus `json:"statuses"`
--- a/github/http.go
+++ b/github/http.go
@@ -3,6 +3,7 @@ package github
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -135,7 +136,33 @@ func newHttpClient(testHost string, verbose bool) *http.Client {
 		Out:         ui.Stderr,
 		Colorized:   ui.IsTerminal(os.Stderr),
 	}
-	return &http.Client{Transport: tr}
+
+	// Implement CheckRedirect callback to fix issues with net/http.
+	fixupCheckRedirect := func(req *http.Request, via []*http.Request) error {
+		// net/http doesn't send a Host header on redirect requests.
+		// TODO: Find or file a Go bug.
+		if req.Host == "" {
+			req.Host = req.URL.Host
+		}
+
+		// Maintain headers after redirect.
+		// https://github.com/golang/go/issues/4800
+		for key, val := range via[0].Header {
+			if req.Host != via[0].Host && strings.EqualFold(key, "Authorization") {
+				continue
+			}
+			req.Header[key] = val
+		}
+
+		// remainder should match http/Client.defaultCheckRedirect()
+		if len(via) >= 10 {
+			return errors.New("stopped after 10 redirects")
+		}
+
+		return nil
+	}
+
+	return &http.Client{Transport: tr, CheckRedirect: fixupCheckRedirect}
 }
 
 func cloneRequest(req *http.Request) *http.Request {
@@ -217,6 +244,12 @@ func (c *simpleClient) jsonRequest(method, path string, body interface{}) (*simp
 
 func (c *simpleClient) Get(path string) (*simpleResponse, error) {
 	return c.performRequest("GET", path, nil, nil)
+}
+
+func (c *simpleClient) GetFile(path string, mimeType string) (*simpleResponse, error) {
+	return c.performRequest("GET", path, nil, func(req *http.Request) {
+		req.Header.Set("Accept", mimeType)
+	})
 }
 
 func (c *simpleClient) Delete(path string) (*simpleResponse, error) {


glennpratt:
I'm thinking for completeness this should have some args:

```
--all
--source-zip
--source-tar
--assets
```

Not sure what the default should be (or perhaps there shouldn't be one). A filter for --assets might be nice, but I don't need one :sunglasses:

github/client.go
```diff
@@ -371,6 +371,20 @@ func (client *Client) DeleteReleaseAsset(asset *ReleaseAsset) (err error) {
 	return
 }
 
+func (client *Client) DownloadReleaseAsset(url string) (asset io.ReadCloser, err error) {
+	api, err := client.simpleApi()
+	if err != nil {
+		return
+	}
+
+	asset, err = api.GetBody(url, "application/octet-stream")
```
mislav: Could we avoid adding a new `GetBody` function and simply use `api.Get` to obtain a response object and return its `.Body` property from this function? I think it will result in less overall added code.

glennpratt: @mislav `Get()` doesn't appear to allow setting the Accept header. Should I change it's signature?

glennpratt: @mislav I just assumed the answer was yes. See new diff.

mislav: > `Get()` doesn't appear to allow setting the Accept header. Should I change it's signature?

Oh I see. I don't think we should be changing Get signature across the board. I forgot that Get doesn't support a configure block. I would go back to your initial idea of creating a specialized method.

mislav:
Thanks for your pull request! This is a good start, but we'll have to polish this command's outside API a bit. First of all, what would be the use-case for this command? Will people usually want to download all assets, or just a specific one? Will the GET http request need to follow redirects? Etc.

Thanks for your patience!

glennpratt:
@mislav 
 - Use case: I'm using it to download official releases of our private software (compiled), so that we aren't compiling it over and over and have consistent builds.  In our case, everyone has a checkout available already, but it could be nice to even support a project path arg (:user/:repo) and skip the checkout.

- Asset filter: I'm downloading all because I only have one asset per release (only have one target). That would probably be different for others - but should that be a blocker or a future feature? I definitely do not wan't the source assets and am not sure why many would vs Git, but that's easy enough to support.

- Redirects: according to the docs, the GET request must follow redirects and presumably does.

mislav:
> * but it could be nice to even support a project path arg (:user/:repo) and skip the checkout.

I had this idea that supporting a `:user/:repo` arg across the board to specify a non-checked out repo would be a great addition to hub.

> * but should that be a blocker or a future feature?

Not necessarily a blocker for this PR, but I would definitely like to reach a decision before this subcommand hits the next feature release.

> * according to the docs, the GET request must follow redirects and presumably does.

I don't think it does… the `release` family of commands uses a rudimentary API client called `simpleApi` internally that has a very lightweight implementation and no redirect following. I think you would have to implement that for DownloadAsset.

glennpratt:
@mislav it is following redirects. The final URL is available in the response:

```go
response.Request.URL.String()
```

A test, just adding a couple prints to the current code (sensitive parts changed):
```
url
https://api.github.com/repos/myorg/my-project/releases/assets/1292217

response.Request.URL.String()
https://github-cloud.s3.amazonaws.com/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=AKIAI000000000000000%FFFFFFFFF%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20160209T233542Z&X-Amz-Expires=300&X-Amz-Signature=0000000000000000000000000000000000000000000000000000&X-Amz-SignedHeaders=host&actor_id=0&response-content-disposition=attachment%3B%20filename%3Dmy-project-0.0.0-gpratt7-linux-3.19.0-30-generic-x86_64.tar.gz&response-content-type=application%2Foctet-stream
```

github/client.go
```diff
@@ -371,6 +372,22 @@ func (client *Client) DeleteReleaseAsset(asset *ReleaseAsset) (err error) {
 	return
 }
 
+func (client *Client) DownloadReleaseAsset(url string) (asset io.ReadCloser, err error) {
+	api, err := client.simpleApi()
+	if err != nil {
+		return
+	}
+
+	result, err := api.Get(url, func(req *http.Request) {
```
mislav: Could this use `performRequest` so we don't have to change `Get()` signature?

mislav:
Oh wow, I never realized this API client would follow redirects automatically. I'm not sure where it happens, but it's awesome if it works.

Could you add a cuke for this feature to `features/release.feature`?

github/client.go
glennpratt: Not directly because performRequest isn't exported, but I added a `GetFile()`.

commands/release.go
```diff
@@ -195,6 +202,45 @@ func showRelease(cmd *Command, args *Args) {
...
+	for _, asset := range release.Assets {
+		assetReader, apiError := gh.DownloadReleaseAsset(asset.ApiUrl)
+		utils.Check(apiError)
+		if assetReader == nil {
+			continue
+		}
+
+		ui.Printf("Downloading %s ...\n", asset.Name)
+
+		assetFile, err := os.OpenFile(asset.Name, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0600)
```
mislav: How about `os.CreateFile`?

commands/release.go
```diff
@@ -195,6 +202,45 @@ func showRelease(cmd *Command, args *Args) {
...
+
+	gh := github.NewClient(project.Host)
+
+	release, err := gh.FetchRelease(project, tagName)
+	utils.Check(err)
+
+	for _, asset := range release.Assets {
+		assetReader, apiError := gh.DownloadReleaseAsset(asset.ApiUrl)
+		utils.Check(apiError)
+		if assetReader == nil {
```
mislav: This conditional is superfluous because the line before it checks `apiError`?

commands/release.go
```diff
@@ -195,6 +202,45 @@ func showRelease(cmd *Command, args *Args) {
...
+
+		ui.Printf("Downloading %s ...\n", asset.Name)
+
+		assetFile, err := os.OpenFile(asset.Name, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0600)
+		utils.Check(err)
+
+		_, err = io.Copy(assetFile, assetReader)
+		utils.Check(err)
+
+		assetFile.Close()
```
mislav: Nitpick: maybe add `defer assetFile.Close()` right after opening the file so opening/closing logic is kept visually together?

mislav:
Few nits, but otherwise this is coming together really nicely! :+1: thanks for your patience

glennpratt:
Thanks for the comments. I'm poking at the cucumber test as I get free time - thought I don't have much the next few days.

commands/release.go
glennpratt: If you mean `os.Create`, it truncates an existing file. I fail if there is an existing file (`O_EXCL`). Ideally I would append (n) if the file already exists like a browser might do.

mislav: Gotcha, makes sense. How about using `0644` as default permissions?

glennpratt:
Added cucumber scenario. I'll address your comments as a get more free time.

commands/release.go
glennpratt: Oh, good catch. I saw this pattern somewhere else.

> When err is nil, resp always contains a non-nil resp.Body. Caller should close resp.Body when done reading from it.

https://golang.org/pkg/net/http/#Client.Do

glennpratt:
@mislav addressed your comments.

features/release.feature
```diff
@@ -294,3 +294,42 @@ MARKDOWN
...
+              ],
+            },
+          ]
+        }
+        get('/repos/mislav/will_paginate/assets/9876') {
+          status 302
+          headers['Location'] = 'https://github-cloud.s3.amazonaws.com/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz'
+          ""
+        }
+        get('/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz') {
```
mislav: Could you append the `:host_name => 'github-cloud.s3.amazonaws.com'` constraint to this route so it serves as an assertion that this route was called with the correct host name and not `api.github.com`? Thanks

glennpratt: @mislav good idea, just tried.

I get a 404. If I return the host header in the current method, it's 127.0.0.1. I'm not sure what's happening here but I suspect it's an artifact of the testing framework.

```
        get('/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz') {
          headers['Content-Type'] = 'application/octet-stream'
          request.host
        }
```
```
expected: "ASSET_TARBALL"
     got: "127.0.0.1"
```

mislav: Strange. That should work, since we're using `:host_name` constraint elsewhere in the test suite. When debugging, I think you should probably look at `request.headers['host']` or `request.env['HTTP_HOST']` since `request.host` is not guaranteed to be accurate.

glennpratt: Hrmm, maybe the Go client isn't redirecting correctly. I'll try changing the code to match the redirect tooling I see in `Get()` though it might be a few days before I do that.

tcpflow output:
```
127.000.000.001.53925-127.000.000.001.53924: GET /__identify__ HTTP/1.1
Accept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3
Accept: */*
User-Agent: Ruby
Host: 127.0.0.1:53924


127.000.000.001.53925-127.000.000.001.53924: GET /__identify__ HTTP/1.1
Accept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3
Accept: */*
User-Agent: Ruby
Host: 127.0.0.1:53924


127.000.000.001.53924-127.000.000.001.53925: HTTP/1.1 200 OK
Server: WEBrick/1.3.1 (Ruby/2.2.3/2015-08-18)
Date: Wed, 17 Feb 2016 15:13:53 GMT
Content-Length: 14
Connection: Keep-Alive


127.000.000.001.53924-127.000.000.001.53925: HTTP/1.1 200 OK
Server: WEBrick/1.3.1 (Ruby/2.2.3/2015-08-18)
Date: Wed, 17 Feb 2016 15:13:53 GMT
Content-Length: 14
Connection: Keep-Alive


127.000.000.001.53924-127.000.000.001.53925: 70296040667240
127.000.000.001.53924-127.000.000.001.53925: 70296040667240
127.000.000.001.53928-127.000.000.001.53924: GET /repos/mislav/will_paginate/releases HTTP/1.1
Host: api.github.com
User-Agent: Go-http-client/1.1
Authorization: token OTOKEN
X-Original-Port: 80
X-Original-Scheme: https
Accept-Encoding: gzip


127.000.000.001.53928-127.000.000.001.53924: GET /repos/mislav/will_paginate/releases HTTP/1.1
Host: api.github.com
User-Agent: Go-http-client/1.1
Authorization: token OTOKEN
X-Original-Port: 80
X-Original-Scheme: https
Accept-Encoding: gzip


127.000.000.001.53924-127.000.000.001.53928: HTTP/1.1 200 OK
Content-Type: application/json;charset=utf-8
Content-Length: 332
Server: WEBrick/1.3.1 (Ruby/2.2.3/2015-08-18)
Date: Wed, 17 Feb 2016 15:13:53 GMT
Connection: Keep-Alive


127.000.000.001.53924-127.000.000.001.53928: HTTP/1.1 200 OK
Content-Type: application/json;charset=utf-8
Content-Length: 332
Server: WEBrick/1.3.1 (Ruby/2.2.3/2015-08-18)
Date: Wed, 17 Feb 2016 15:13:53 GMT
Connection: Keep-Alive


127.000.000.001.53924-127.000.000.001.53928: [{"url":"https://api.github.com/repos/mislav/will_paginate/releases/123","upload_url":"https://api.github.com/uploads/assets{?name,label}","tag_name":"v1.2.0","name":"will_paginate 1.2.0","draft":true,"prerelease":false,"assets":[{"url":"https://api.github.com/repos/mislav/will_paginate/assets/9876","name":"hello-1.2.0.tar.gz"}]}]
127.000.000.001.53924-127.000.000.001.53928: [{"url":"https://api.github.com/repos/mislav/will_paginate/releases/123","upload_url":"https://api.github.com/uploads/assets{?name,label}","tag_name":"v1.2.0","name":"will_paginate 1.2.0","draft":true,"prerelease":false,"assets":[{"url":"https://api.github.com/repos/mislav/will_paginate/assets/9876","name":"hello-1.2.0.tar.gz"}]}]
127.000.000.001.53929-127.000.000.001.53924: GET /repos/mislav/will_paginate/assets/9876 HTTP/1.1
Host: api.github.com
User-Agent: Go-http-client/1.1
Authorization: token OTOKEN
Content-Type: application/octet-stream
X-Original-Port: 80
X-Original-Scheme: https
Accept-Encoding: gzip


127.000.000.001.53929-127.000.000.001.53924: GET /repos/mislav/will_paginate/assets/9876 HTTP/1.1
Host: api.github.com
User-Agent: Go-http-client/1.1
Authorization: token OTOKEN
Content-Type: application/octet-stream
X-Original-Port: 80
X-Original-Scheme: https
Accept-Encoding: gzip


127.000.000.001.53924-127.000.000.001.53929: HTTP/1.1 302 Found
Content-Type: application/json;charset=utf-8
Location: https://github-cloud.s3.amazonaws.com/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz
Content-Length: 2
Server: WEBrick/1.3.1 (Ruby/2.2.3/2015-08-18)
Date: Wed, 17 Feb 2016 15:13:53 GMT
Connection: Keep-Alive


127.000.000.001.53924-127.000.000.001.53929: HTTP/1.1 302 Found
Content-Type: application/json;charset=utf-8
Location: https://github-cloud.s3.amazonaws.com/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz
Content-Length: 2
Server: WEBrick/1.3.1 (Ruby/2.2.3/2015-08-18)
Date: Wed, 17 Feb 2016 15:13:53 GMT
Connection: Keep-Alive


127.000.000.001.53924-127.000.000.001.53929: {}
127.000.000.001.53924-127.000.000.001.53929: {}
127.000.000.001.53929-127.000.000.001.53924: GET /releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz HTTP/1.1
Host: 127.0.0.1:53924
User-Agent: Go-http-client/1.1
Referer: https://api.github.com/repos/mislav/will_paginate/assets/9876
X-Original-Port: 80
X-Original-Scheme: https
Accept-Encoding: gzip


127.000.000.001.53929-127.000.000.001.53924: GET /releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz HTTP/1.1
Host: 127.0.0.1:53924
User-Agent: Go-http-client/1.1
Referer: https://api.github.com/repos/mislav/will_paginate/assets/9876
X-Original-Port: 80
X-Original-Scheme: https
Accept-Encoding: gzip


127.000.000.001.53924-127.000.000.001.53929: HTTP/1.1 200 OK
Content-Type: application/octet-stream
Content-Length: 1238
Server: WEBrick/1.3.1 (Ruby/2.2.3/2015-08-18)
Date: Wed, 17 Feb 2016 15:13:53 GMT
Connection: Keep-Alive


127.000.000.001.53924-127.000.000.001.53929: HTTP/1.1 200 OK
Content-Type: application/octet-stream
Content-Length: 1238
Server: WEBrick/1.3.1 (Ruby/2.2.3/2015-08-18)
Date: Wed, 17 Feb 2016 15:13:53 GMT
Connection: Keep-Alive


127.000.000.001.53924-127.000.000.001.53929: {
  "GATEWAY_INTERFACE": "CGI/1.1",
  "PATH_INFO": "/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz",
  "QUERY_STRING": "",
  "REMOTE_ADDR": "127.0.0.1",
  "REMOTE_HOST": "localhost",
  "REQUEST_METHOD": "GET",
  "REQUEST_URI": "http://127.0.0.1:53924/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz",
  "SCRIPT_NAME": "",
  "SERVER_NAME": "127.0.0.1",
  "SERVER_PORT": "53924",
  "SERVER_PROTOCOL": "HTTP/1.1",
  "SERVER_SOFTWARE": "WEBrick/1.3.1 (Ruby/2.2.3/2015-08-18)",
  "HTTP_HOST": "127.0.0.1:53924",
  "HTTP_USER_AGENT": "Go-http-client/1.1",
  "HTTP_REFERER": "https://api.github.com/repos/mislav/will_paginate/assets/9876",
  "HTTP_X_ORIGINAL_PORT": "80",
  "HTTP_X_ORIGINAL_SCHEME": "https",
  "HTTP_ACCEPT_ENCODING": "gzip",
  "rack.version": [
    1,
    1
  ],
  "rack.input": "#<StringIO:0x007fde2505e988>",
  "rack.errors": "#<IO:0x007fde240c6520>",
  "rack.multithread": true,
  "rack.multiprocess": false,
  "rack.run_once": false,
  "rack.url_scheme": "http",
  "HTTP_VERSION": "HTTP/1.1",
  "REQUEST_PATH": "/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz",
  "rack.logger": "#<Rack::NullLogger:0x007fde2510f120>",
  "rack.request.query_string": "",
  "rack.request.query_hash": {
  }
}
127.000.000.001.53924-127.000.000.001.53929: {
  "GATEWAY_INTERFACE": "CGI/1.1",
  "PATH_INFO": "/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz",
  "QUERY_STRING": "",
  "REMOTE_ADDR": "127.0.0.1",
  "REMOTE_HOST": "localhost",
  "REQUEST_METHOD": "GET",
  "REQUEST_URI": "http://127.0.0.1:53924/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz",
  "SCRIPT_NAME": "",
  "SERVER_NAME": "127.0.0.1",
  "SERVER_PORT": "53924",
  "SERVER_PROTOCOL": "HTTP/1.1",
  "SERVER_SOFTWARE": "WEBrick/1.3.1 (Ruby/2.2.3/2015-08-18)",
  "HTTP_HOST": "127.0.0.1:53924",
  "HTTP_USER_AGENT": "Go-http-client/1.1",
  "HTTP_REFERER": "https://api.github.com/repos/mislav/will_paginate/assets/9876",
  "HTTP_X_ORIGINAL_PORT": "80",
  "HTTP_X_ORIGINAL_SCHEME": "https",
  "HTTP_ACCEPT_ENCODING": "gzip",
  "rack.version": [
    1,
    1
  ],
  "rack.input": "#<StringIO:0x007fde2505e988>",
  "rack.errors": "#<IO:0x007fde240c6520>",
  "rack.multithread": true,
  "rack.multiprocess": false,
  "rack.run_once": false,
  "rack.url_scheme": "http",
  "HTTP_VERSION": "HTTP/1.1",
  "REQUEST_PATH": "/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz",
  "rack.logger": "#<Rack::NullLogger:0x007fde2510f120>",
  "rack.request.query_string": "",
  "rack.request.query_hash": {
  }
}
```

mislav: Yeah seems that Go isn't handling 302 correctly. At first I thought that their HTTP client doesn't do redirects automatically at all, but now it turns out that it does, but messes it up. This might be a problem.

We should probably provide [our own `CheckRedirect` implementation](https://golang.org/pkg/net/http/#Client). Think you can take a stab at that?

glennpratt: @mislav I tried, but it doesn't seem to get called. afb17d8

Interestingly, this worked with real GH releases for a week or so after I opened this PR. Today, it no longer works - probably because of the host header, though I'm not totally sure. I verified cURL has a correct host header while following the redirects.

glennpratt: Ok, fixed! Last issues with S3 was pebkac. S3 doesn't seem to care about the host header so far (probably because they don't use SNI). I also implemented redirect header preservation to match cURL behavior. It's not strictly required if you are concerned about it.

github/http.go
```diff
@@ -135,7 +136,33 @@ func newHttpClient(testHost string, verbose bool) *http.Client {
...
+		// net/http doesn't send a Host header on redirect requests.
+		// TODO: Find or file a Go bug.
+		if req.Host == "" {
+			req.Host = req.URL.Host
+		}
+
+		// Maintain headers after redirect.
+		// https://github.com/golang/go/issues/4800
+		for key, val := range via[0].Header {
+			if key == "Authorization" {
```
mislav: We should do a case-insensitive comparison here. Also, could we maintain even this header if the redirect is to the same host?

features/release.feature
```diff
@@ -294,3 +294,50 @@ MARKDOWN
...
+          unless request.accept?('application/octet-stream')
+            status 415
+            return "Release API requires Accept: application/octet-stream to return file."
+          end
+          status 302
+          headers['Location'] = 'https://github-cloud.s3.amazonaws.com/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz'
+          ""
+        }
+        get('/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz', :host_name => 'github-cloud.s3.amazonaws.com') {
+          unless request.accept?('application/octet-stream')
```
mislav: Could we also assert that the `Authorization` header wasn't sent to the 3rd-party host?

mislav:
Awesome work on the redirect stuff! Still will come in handy for various different operations in the future :smile_cat:

github/http.go
glennpratt: Done.

features/release.feature
glennpratt: Done.

glennpratt:
@mislav addressed your comments.

commands/release.go
```diff
@@ -195,6 +205,41 @@ func showRelease(cmd *Command, args *Args) {
...
+
+	gh := github.NewClient(project.Host)
+
+	release, err := gh.FetchRelease(project, tagName)
+	utils.Check(err)
+
+	for _, asset := range release.Assets {
+		assetReader, apiError := gh.DownloadReleaseAsset(asset.ApiUrl)
+		utils.Check(apiError)
+		defer assetReader.Close()
```
jingweno: We could close the `assetReader` on each loop is finished instead of waiting for the whole program.

commands/release.go
```diff
@@ -195,6 +205,41 @@ func showRelease(cmd *Command, args *Args) {
...
+
+		ui.Printf("Downloading %s ...\n", asset.Name)
+
+		assetFile, err := os.OpenFile(asset.Name, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0644)
+		utils.Check(err)
+		defer assetFile.Close()
```
jingweno: Same as above. Alternative, we could extract the content of the for loop into a function then we could use `defer` there.

commands/release.go
```diff
@@ -195,6 +205,41 @@ func showRelease(cmd *Command, args *Args) {
...
+
+		_, err = io.Copy(assetFile, assetReader)
+		utils.Check(err)
+	}
+
+	os.Exit(0)
```
jingweno: Do we need the explicit exit here?

commands/release.go
glennpratt: Extracted to a function.

commands/release.go
glennpratt: Extracted to a function.

commands/release.go
glennpratt: I believe so. Without this line it seems to fall through to a missing argument handler and tests fail. I didn't look into this deeply, just a guess.

glennpratt:
@jingweno I addressed your comments.

mislav:
@glennpratt Thanks for all the hard work!

glennpratt:
Woohoo. Thank you for all the guidance.

PR #1110, opened 2016-02-12
Make tests more robust to various Locales.

pcorpet:
This shouldn't affect the CI, but it makes easier to test on my locale machine given that git runs with my current locale (which is not english) thus the git messages were different from what those tests expect.

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -46,6 +46,7 @@ Before do
   set_env 'GIT_EDITOR', 'false'
   # reset current localization settings
   set_env 'LANG', nil
+  set_env 'LANGUAGE', nil
   set_env 'LC_ALL', 'en_US.UTF-8'
   # ignore current user's token
   set_env 'GITHUB_TOKEN', nil


mislav:
Thanks for letting me know! Isn't this already covered by [the global test config](https://github.com/github/hub/blob/7d5a40b2719fa019be0d8a57358471f99af07244/features/support/env.rb#L47-L49)?

pcorpet:
Hi @mislav , thanks for the pointer. Apparently my git is also checking the LANGUAGE env var so I've added it as well in the global test config. Please Take Another Look.

mislav:
Thanks for the update! For posterity, what's your LANGUAGE setting and what's the error that you're getting?

pcorpet:
LANGUAGE="fr_FR"

I had many errors, here's one of them:
expected: "Hello\n\n\n# Requesting a pull to mislav:master from mislav:topic\n#\n# Write a message for this pull request. The first block\n# of text is the title and the rest is the description.\n#\n# Changes:\n#\n# SHA1SHA (Hub, 0 seconds ago)\n#    Two on topic\n#\n# SHA1SHA (Hub, 0 seconds ago)\n#    One on topic\n"
     got: "Hello\n\n\n# Requesting a pull to mislav:master from mislav:topic\n#\n# Write a message for this pull request. The first block\n# of text is the title and the rest is the description.\n#\n# Changes:\n#\n# 4d21e49 (Hub, il y a 0 seconde)\n#    Two on topic\n#\n# 7dc2ced (Hub, il y a 0 seconde)\n#    One on topic\n"

Diff:
-# SHA1SHA (Hub, 0 seconds ago)
+# 4d21e49 (Hub, il y a 0 seconde)

mislav:
Thanks for the fix!

Curious that LANGUAGE doesn't affect anything for me on OS X, git version 2.6.3.

Issue #1107, opened 2016-02-08
hub browse should directly open the pull request page


dufferzafar:
I am not sure if this is too far fetched or if it's even possible. So please bear with me :smile_cat: 

If I am on a feature branch that has an existing pull request on Github and I do a `hub browse`, I just want it to open the pull request page rather than the repository page.

So, let's say I am on branch `cool-feature`, and it is associated to pull request 90, then browse should just open `github.com/:user/:repo/pulls/90`

mislav:
That's an interesting suggestion. I definitely see a value in that feature, but what if a person expected to simply open the current branch online and didn't appreciate that this behavior will differ based on whether the current branch has a PR or not?

dufferzafar:
@mislav Um, we provide switches to handle the use cases?

mislav:
> we provide switches to handle the use cases?

That's an option of course. How would the switches look like?

dufferzafar:
If you want the default behavior of `hub browse` to stay unchanged, we could provide switches like `-p` for opening up the linked pull request, `-b` for the branch etc.

pcorpet:
This is a pretty easy follow up to #1109 . @mislav Assign it to me if you wish.

dufferzafar:
I needed this feature again today. Are there any plans on working on it? since people seem to agree with the feature request.

mislav:
@dufferzafar Nobody is working on this yet, afaik. It's up for grabs!

dufferzafar:
@mislav Okay. I might work on this, but I'm new to the project. Can you please tell where I'd start working on this? I saw PR #1109 being mentioned above but I had a look at it, and was quite overwhelmed by the level of discussion there. Just knowing which files I'd need to look into etc. would be a great help!

mislav:
@dufferzafar It can be tricky to navigate your way around a new codebase, but you shouldn't let it deter you.

Code-wise, you could definitely start at `browse.go`, see how that's implemented, and try to squeeze in a new option flag. But I wouldn't recommend it, since `browse` command is already too overloaded and messy and I'd like to make a new command to replace it.

So I'd rather start a new subcommand `hub open` and make it open a pull request page if the current branch corresponds to one. Then later we can extend the new command with abilities to browse other branches, commits, tags, and so on.

mislav:
I don't think we will be changing the `browse` command to add this feature (or any other features, for that matter), and instead we might deprecate it in favor of the upcoming `hub open` command that takes slightly different arguments.

[...]

PR #1112, opened 2016-02-13
Adds a flag to `hub browse` to choose whether to open the current or …

pcorpet:
…default branch.

This is a start for #1107 to let the user control a bit more which page  is open when running "hub browse" in a tracked branch.


mislav:
Looks good, but I'm reluctant to add a new flag that doesn't seem that it will get much use. What's the use-case for this?

pcorpet:
I agree, I don't find much value in it. I was merely following the suggestion in #1107.
"If you want the default behavior of hub browse to stay unchanged, we could provide switches like -p for opening up the linked pull request, -b for the branch etc."

I'm just trying to help out here, so if you don't like it, just push back in #1107 and let's drop this PR.

npcode:
I want this feature to see the code reviews on the pull request related with the branch I'm working on.

SamirBoulil:
@npcode That's why I ended up doing in bash .aliases:
`echo $(hub browse -u)| sed "s/tree/pulls/g" | xargs open`

I didn't find any way to use the `hub browse -- [sub-page]` to replace the "tree" segment of the URL for "pulls" given the current directory.

PR #1114, opened 2016-02-14
Fix git remote when using -C:

### LINKED ISSUE #1105: hub breaks git -C for the `git remote` command
```
Josephs-MacBook-Pro:~ joe$ /usr/bin/git --version
git version 1.8.5.2 (Apple Git-48)
Josephs-MacBook-Pro:~ joe$ /usr/local/bin/git --version
git version 2.7.1
Josephs-MacBook-Pro:~ joe$ alias | grep hub
alias git='hub'
Josephs-MacBook-Pro:~ joe$ git --version
git version 2.7.1
hub version 2.2.3
Josephs-MacBook-Pro:~ joe$ hub --version
git version 2.7.1
hub version 2.2.3
Josephs-MacBook-Pro:~ joe$ /usr/bin/git -C /usr/local remote
ilovezfs
ilovezfs_local
origin
this
Josephs-MacBook-Pro:~ joe$ /usr/local/bin/git -C /usr/local remote
ilovezfs
ilovezfs_local
origin
this
Josephs-MacBook-Pro:~ joe$ git -C /usr/local remote
fatal: Not a git repository (or any of the parent directories): .git
Josephs-MacBook-Pro:~ joe$ hub -C /usr/local remote
fatal: Not a git repository (or any of the parent directories): .git

pcorpet:
- use Git root dir instead of working dir to pick a repo name.

Fixes #1105.

### DIFF
--- a/commands/create.go
+++ b/commands/create.go
@@ -69,7 +69,7 @@ func create(command *Command, args *Args) {
 
 	var newRepoName string
 	if args.IsParamsEmpty() {
-		newRepoName, err = utils.DirName()
+		newRepoName, err = git.RootDirName()
 		utils.Check(err)
 	} else {
 		reg := regexp.MustCompile("^[^-]")
--- a/commands/remote.go
+++ b/commands/remote.go
@@ -5,6 +5,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/github/hub/git"
 	"github.com/github/hub/github"
 	"github.com/github/hub/utils"
 )
@@ -68,7 +69,7 @@ func transformRemoteArgs(args *Args) {
 			repoName = project.Name
 			host = project.Host
 		} else {
-			repoName, err = utils.DirName()
+			repoName, err = git.RootDirName()
 			utils.Check(err)
 		}
 
--- a/features/remote_add.feature
+++ b/features/remote_add.feature
@@ -9,6 +9,14 @@ Feature: hub remote add
     Then the url for "origin" should be "git@github.com:EvilChelu/dotfiles.git"
     And there should be no output
 
+  Scenario: Add origin remote for my own repo using -C
+    Given there are no remotes
+    And I cd to ".."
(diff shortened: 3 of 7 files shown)


mislav:
Hi thanks for the contribution but your diff was pretty big, and I was sure that the fix for the original issue should be pretty small so I've attempted it myself: 40e0cd22fd141d9b98bb69626f338f4fd9d581af

In the future, could you split different fixes/features into separate PRs?

What does the git root dir change fix and could you open a new PR which only addresses that? Thanks!

pcorpet:
I had almost the same code as you do (with the same little change), and then start testing with a real case scenario (instead of unit test as you do), using the scenario described in #1105. I quickly realized that it was then taking the wrong name for the repo, and used the name where the user was working (aruba.git in my case) instead of the one that was specified by -C (dotfiles in my case).

mislav:
Could you now extract your git root dir name fix as a separate PR please?

pcorpet:
> Could you now extract your git root dir name fix as a separate PR please?
Done.

pcorpet:
A friendly ping.

mislav:
Thanks for your patience. I wasn't completely on board with your refactoring of helper functions. For instance, the `RootDirName()` function seemed to be a misnomer since it didn't really return the actual git root directory. All it did is resolve the result of `-C` global flags, but if those were missing, the function returned a blank string. Also, if you were deeply nested within a project's subdirectories, this method wouldn't return the root directory for the project.

Instead, I've adopted your test and taken my own approach. First, I removed `utils.DirName()` dependency from some commands. Then, I've replaced remaining `DirName()` invocations with a combination of two new commands: `git.WorkdirName()` & `github.SanitizeProjectName()`. The resulting invocation are a bit more explicit and thus more clear in what they do, the individual functions are simpler, and I avoided having to change `git.Dir()` function.

pcorpet:
Great, I like it.

PR #1116, opened 2016-02-15
Don't use octokit in GhLatestTagName.

pcorpet:
This introduces no change at all, but start cleaning up the usage of octokit.


mislav:
Thanks! I think this code used for the auto-updater is dead (since it points to the old repo `jingweno/gh`), but a go-octokit cleanup is welcome nevertheless. We can revisit the auto-updater code later.

PR #1120, opened 2016-02-18
Restore `go get` installation instructions

tSU-RooT:
According to [Go 1.6 Release Note](https://blog.golang.org/go1.6), 
vendor experiment is enabled by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,13 @@ $ cd hub
 $ ./script/build -o ~/bin/hub
 ~~~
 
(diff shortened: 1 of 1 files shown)


mislav:
Thanks for the ping, but we still haven't migrated the project to require Go 1.6 at minimum, and I'd hate to maintain different sets of installation instructions for different Go versions. Is `go get` in Go 1.6 even aware of vendor directories?

mislav:
The instructions are now back in ddd5b167d5cc8166eb0b105cd65e7de51998efb5!

PR #1126, opened 2016-02-27
Fixes #1125

### LINKED ISSUE #1125: build on netbsd
I needed to comment out `"github.com/mattn/go-isatty"` and `return isatty.IsTerminal(f.Fd())` on `ui/ui.go` to make it build.

It works great now.

aerth:
Is it assumed anyway that if you run hub you are on a terminal?
Or does that function prevent something like `cat photo.jpg | hub`, which in that case hub would not be running in a terminal.

Anyways, it runs on NetBSD now.

### DIFF
--- a/ui/ui.go
+++ b/ui/ui.go
@@ -6,7 +6,7 @@ import (
 	"os"
 
 	"github.com/mattn/go-colorable"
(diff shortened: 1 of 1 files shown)


aerth:
Breaks everything else

PR #1136, opened 2016-03-05
bashcomp: try to load git comp first (fix #592)

### LINKED ISSUE #592: hub bash completions don't work by default when git's bash completions are loaded dynamically

junghans:
Not super portable, but a step forward.

### DIFF
--- a/etc/hub.bash_completion.sh
+++ b/etc/hub.bash_completion.sh
@@ -1,6 +1,11 @@
 # hub tab-completion script for bash.
 # This script complements the completion script that ships with git.
 
+# If there is no git tab completion, but we have the _completion loader try to load it
+if ! declare -F _git > /dev/null && declare -F _completion_loader > /dev/null; then
+  _completion_loader git
+fi
+
 # Check that git tab completion is available
 if declare -F _git > /dev/null; then
   # Duplicate and rename the 'list_all_commands' function


junghans:
In testing: gentoo/gentoo@a8181ee921c13ac81be1a6a5beb8b20a62ff7b83

junghans:
@mislav

mislav:
Sorry for late reply. This looks innocent enough to add, and I'm happy that you found a workaround for dynamically loaded completion scripts!

PR #1137, opened 2016-03-07
Update README.md

gleber:
Fix installation command for zsh


gleber:
Ah, I can not accept the CLA right away due to other contractual obligations, chence losing the request.

PR #1139, opened 2016-03-08
Fix #1100: Append /api/v3/ to request when using GHE

### LINKED ISSUE #1100: ci-status: invalid character '<' looking for beginning of value
```
hub ci-status -v b116a403b9cbb95af90a809b65b36098d5c9c397
invalid character '<' looking for beginning of value
```

I am trying to use it against a GitHub Enterprise instance.

queeno:
Git Hub Enterprise requests should always include /api/v3 in the URL in order to hit the correct API endpoint. The ci-status and release commands were producing the wrong request URL.

```
> GET https://[GHE URL]/repos/[ORG]/[REPO]/commits/[SHA]/status
```
rather than:
```
> GET https://[GHE URL]/api/v3/repos/[ORG]/[REPO]/commits/[SHA]/status
```

As result, an LDAP authentication page was returned rather than the expected JSON. A parsing error could be seen:

```
invalid character '<' looking for beginning of value
```

This commit changes the construction of the request URL for the ci-status and release commands, making sure the existing client.requestURL function is called. This makes sure the API request always hits the right endpoint in Git Hub Enterprise.

### DIFF
--- a/github/client.go
+++ b/github/client.go
@@ -274,7 +274,12 @@ func (client *Client) FetchReleases(project *Project) (response []Release, err e
 		return
 	}
 
-	res, err := api.Get(fmt.Sprintf("repos/%s/%s/releases", project.Owner, project.Name))
+	request_url, err := url.Parse(fmt.Sprintf("repos/%s/%s/releases", project.Owner, project.Name))
+	if err != nil {
+		return
+	}
+
+	res, err := api.Get(client.requestURL(request_url).String())
 	if err = checkStatus(200, "fetching releases", res, err); err != nil {
 		return
 	}
@@ -402,7 +407,12 @@ func (client *Client) FetchCIStatus(project *Project, sha string) (status *CISta
 		return
 	}
 
-	res, err := api.Get(fmt.Sprintf("repos/%s/%s/commits/%s/status", project.Owner, project.Name, sha))
+	request_url, err := url.Parse(fmt.Sprintf("repos/%s/%s/commits/%s/status", project.Owner, project.Name, sha))
+	if err != nil {
+		return
+	}
+
+	res, err := api.Get(client.requestURL(request_url).String())
 	if err = checkStatus(200, "fetching statuses", res, err); err != nil {
 		return
 	}
--- a/github/client.go
+++ b/github/client.go
@@ -274,12 +274,12 @@ func (client *Client) FetchReleases(project *Project) (response []Release, err e
 		return
 	}
 
-	request_url, err := url.Parse(fmt.Sprintf("repos/%s/%s/releases", project.Owner, project.Name))
+	requestUrl, err := url.Parse(fmt.Sprintf("repos/%s/%s/releases", project.Owner, project.Name))
 	if err != nil {
 		return
 	}
 
-	res, err := api.Get(client.requestURL(request_url).String())
+	res, err := api.Get(client.requestURL(requestUrl).String())
 	if err = checkStatus(200, "fetching releases", res, err); err != nil {
 		return
 	}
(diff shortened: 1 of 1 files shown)


zhongjiewu:
This sounds great. Really looking forward to get this fix in the next release

buehmann:
Works great for me. Thanks!

mislav:
The fix looks great. Sorry that I haven't seen this before.

github/client.go
```diff
@@ -274,7 +274,12 @@ func (client *Client) FetchReleases(project *Project) (response []Release, err e
 		return
 	}
 
-	res, err := api.Get(fmt.Sprintf("repos/%s/%s/releases", project.Owner, project.Name))
+	request_url, err := url.Parse(fmt.Sprintf("repos/%s/%s/releases", project.Owner, project.Name))
```
mislav: Two problems here:

1. We use camelCase for variable naming convention, so `request_url` should be `requestUrl`.

2. I dislike the fact that we construct a URL object out of this string, pass to `client.requestURL()` which takes a URL, and then extract the `.String()` value from the resulting URL. The whole String → URL → String dance now feels unnecessary. How about extracting the part of the logic from `client.requestURL()` into a new method that only prepends `/api/v3` to a string, and calling it something like `normalizeRequestPath()`?

mislav:
@queeno Let me know if you have the time to perform the tweaks necessary before merging! Otherwise let me know and I can do the tweaks myself. Let's get this fix in soon 😉

queeno:
@mislav Hey, sorry I've just seen this. Will perform the requested tweaks as soon as possible! 😄

queeno:
@mislav I have committed a fix about point 1.

I very much agree with you about point 2, it's unnecessary to convert a string into url, normalise it and convert it back to string. That said, `client.RequestURL()` is already used elsewhere in the code to perform our desired normalisation.

We have two options here: duplicating the functionality of `client.RequestURL()` by creating a separate function that takes strings rather than urls - as you suggest - or use the existing `client.RequestURL()` and converting parameters in and out.

Since this is a massive community project, I have preferred the latter approach to the former to avoid duplication, keep my change simple and preserve consistency.
For example, should the non-GHE URL ever change in the future, we would need to amend the URL in two separate locations rather than one. This could lead to potential future bugs. Also, only two of the hub would make use of our new method. Are we sure we want to introduce some potential complexity to be extremely efficient in a couple of places only?

If you believe we shouldn't be concerned about some trivial duplication and we should rather concentrate on efficiency, I'm more than happy to introduce the changes you suggested.

github/client.go
```diff
@@ -402,7 +407,12 @@ func (client *Client) FetchCIStatus(project *Project, sha string) (status *CISta
 		return
 	}
 
-	res, err := api.Get(fmt.Sprintf("repos/%s/%s/commits/%s/status", project.Owner, project.Name, sha))
+	request_url, err := url.Parse(fmt.Sprintf("repos/%s/%s/commits/%s/status", project.Owner, project.Name, sha))
```
mislav: This still uses underscores in `request_url` 🙈

mislav:
@queeno Thanks for the update. Yeah, don't worry about changing `RequestURL()` for now. I'll look at the code and see how we can dry this up without duplicating logic when I merge your PR. I have an idea.

queeno:
Ops 😓  Just fixed!

Thanks @mislav!

mislav:
I've noticed a lot more invocations of `simpleApi` where we would have to keep applying the prefix manually by invoking `RequestURL()`. Instead, I've moved the logic of the prefix into a one-time operation when constructing the root URL that the API client will use b346fe0cf8910d4856d476871c62068869584ec7

queeno:
Thanks @mislav

Issue #1118, opened 2016-02-18
Add support to pull-request template like web interface


rickmak:
Github web interface now will pre-fill the pull-request description with the template named `PULL_REQUEST_TEMPLATE`. `hub pull-request` can do this too if no `-m` or `-F` is supplied.

refs: https://github.com/blog/2111-issue-and-pull-request-templates

chris-morgan:
`hub issue create` also for `ISSUE_TEMPLATE`.

mislav:
Good idea to bake this into hub. I could look into whether that would be hard to add to `hub issue` revamp in #1099. Then, the same approach could be adopted for `hub pull-request`.

maxmanders:
Just wanted to express my interest in this issue too - would be a great feature!

pwaller:
:+1: same idea over here too.

ptarjan:
:+1:

alexturek:
Maintainers: https://github.com/github/hub/issues/937 and https://github.com/github/hub/issues/446 are duplicates

maxmanders:
:+1:

tconroy:
Apologies for the old bump, but has this been implemented? Similar to @ptarjan my org has implemented a template policy, but I would prefer to stay in the CLI. :)

PR #1145, opened 2016-03-14
Support Pull Request and issue template

ganmacs:
Ref #1118 

Support Pull  request and issues template.

### DIFF
--- a/github/template.go
+++ b/github/template.go
@@ -10,13 +10,13 @@ import (
 )
 
 const (
-	pullRequestTempalte = "PULL_REQUEST_TEMPLATE"
+	pullRequestTemplate = "PULL_REQUEST_TEMPLATE"
 	issueTemplate       = "ISSUE_TEMPLATE"
 	githubTemplateDir   = ".github"
 )
 
 func GetPullRequestTemplate() string {
-	return getGithubTemplate(pullRequestTempalte)
+	return getGithubTemplate(pullRequestTemplate)
 }
 
 func GetIssueTemplate() string {
--- a/github/template.go
+++ b/github/template.go
@@ -15,11 +15,11 @@ const (
 	githubTemplateDir   = ".github"
 )
 
-func GetPullRequestTempalte() string {
+func GetPullRequestTemplate() string {
 	return getGithubTemplate(pullRequestTempalte)
 }
 
-func GetIssueTempalte() string {
+func GetIssueTemplate() string {
 	return getGithubTemplate(issueTempalte)
 }
 
--- a/github/template.go
+++ b/github/template.go
@@ -1,0 +1,77 @@
+package github
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/github/hub/utils"
+)
+
+const (
+	pullRequestTempalte = "PULL_REQUEST_TEMPLATE"
+	issueTempalte       = "ISSUE_TEMPLATE"
+	githubTemplateDir   = ".github"
+)
+
+func GetPullRequestTempalte() string {
+	return getGithubTemplate(pullRequestTempalte)
+}
+
+func GetIssueTempalte() string {
+	return getGithubTemplate(issueTempalte)
+}
+
+func getGithubTemplate(pat string) (body string) {
+	var path string
+
+	if _, err := os.Stat(githubTemplateDir); err == nil {
+		if p := getFilePath(githubTemplateDir, pat); p != "" {
+			path = p
+		}
+	}
+
+	if path == "" {
+		if p := getFilePath(".", pat); p != "" {
+			path = p
+		}
+	}
+
+	if path == "" {
+		return
+	}
+
+	body, err := readContentsFromFile(path)
+	utils.Check(err)
+	return
+}
+
+func getFilePath(dir, pattern string) string {
+	files, err := ioutil.ReadDir(dir)
+	utils.Check(err)
+
+	for _, file := range files {
+		fileName := file.Name()
+		path := fileName
+
+		if ext := filepath.Ext(fileName); ext == ".md" {
+			path = strings.TrimRight(fileName, ".md")
+		}
+
+		if ok, _ := filepath.Match(pattern, path); ok {
+			return filepath.Join(dir, fileName)
+		}
+	}
+	return ""
+}
+
+func readContentsFromFile(filename string) (contents string, err error) {
+	content, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return
+	}
+
+	contents = strings.Replace(string(content), "\r\n", "\n", -1)
+	return
+}
--- a/commands/issue.go
+++ b/commands/issue.go
@@ -108,13 +108,18 @@ func createIssue(cmd *Command, args *Args) {
 }
 
 func writeIssueTitleAndBody(project *github.Project) (string, string, error) {
-	message := `
+	message := `%s
 # Creating issue for %s.
 #
 # Write a message for this issue. The first block of
 # text is the title and the rest is the description.
 `
-	message = fmt.Sprintf(message, project.Name)
+
+	if template := github.GetIssueTempalte(); template == "" {
+		message = fmt.Sprintf(message, "", project.Name)
+	} else {
+		message = fmt.Sprintf(message, template, project.Name)
+	}
 
 	editor, err := github.NewEditor("ISSUE", "issue", message)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


github/template.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/github/hub/utils"
+)
+
+const (
+	pullRequestTempalte = "PULL_REQUEST_TEMPLATE"
+	issueTempalte       = "ISSUE_TEMPLATE"
+	githubTemplateDir   = ".github"
+)
+
+func GetPullRequestTempalte() string {
```
scompt: There's a typo here. Should be `GetPullRequestTemplate `

github/template.go
```diff
@@ -0,0 +1,77 @@
...
+	return getGithubTemplate(pullRequestTempalte)
+}
+
+func GetIssueTempalte() string {
```
scompt: There's a typo here. Should be `GetIssueTemplate `

dawehner:
It would be quite nice to have that

jeroenvisser101:
@ganmacs is there anything keeping this from being merged?

NickLaMuro:
@jeroenvisser101 So I have used the custom homebrew recipe below to install this from @ganmacs branch:

```ruby
class Hub < Formula
  desc "Add GitHub support to git on the command-line"
  homepage "https://hub.github.com/"
  url 'https://github.com/ganmacs/hub.git', :using => :git, :branch => 'support-github-template'
  version '2'

  option "without-completions", "Disable bash/zsh completions"

  depends_on "go" => :build

  def install
    system "script/build", "-o", "hub"
    bin.install "hub"
    man1.install Dir["man/*"]

    if build.with? "completions"
      bash_completion.install "etc/hub.bash_completion.sh"
      zsh_completion.install "etc/hub.zsh_completion" => "_hub"
    end
  end

  test do
    HOMEBREW_REPOSITORY.cd do
      assert_equal "bin/brew", shell_output("#{bin}/hub ls-files -- bin").strip
    end
  end
end
```

Pretty much the same as the [original formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/hub.rb), so if you use homebrew and are on a mac, you can install it by saving the file above and just running `brew install name_of_this_file.rb`.

That said, there are a few issues that I have noticed since using this that would probably make it not mergable:

* The PR template is meant to be just the description section of the PR, but when it isn't a single commit PR, the template is added to the editor as is, which makes the first line of the template the title instead of the start of the body.
* With a single commit PR, the following happens:
  - If it is just a single line commit, the commit messages is appended to the first line of the template instead of making it the first line and adding two new lines in between
  - If it is a multi-line commit message, same problem, but it spans multiple lines until the last line of the commit message, where the last line of the commit message and the first line of the template are the same.
* Possible typo [here](https://github.com/github/hub/pull/1145/files#diff-319d83eedfd5d5f46fead682343297f7R281) (pedantic, so I would say it doesn't need to be address)
* No tests added for this (arguably optional, but based on the current verbosity of the test suite, I doubt it)

jeroenvisser101:
@NickLaMuro thanks – the problems you mention I can see as well. I'll try using the homebrew recipe you sent, thanks!

NickLaMuro:
To anyone hoping to use this before it is merged, I have made some updates to @ganmacs's PR in the following:

https://github.com/ganmacs/hub/pull/1

I have no expectations that this will or will not get merged in, but I am providing it as a stop gap for those that would like to use this feature (as is) before it is merged and released (or redone entirely for all I know).

For those on OSX, the above homebrew recipe will work, but requires a few updates:

* Make sure the file is named `hub.rb`.  I failed to mention that before (because I forgot), but Homebrew will barf if you don't have that same name as the class name of the file.
* Update the url in the formula from "https://github.com/ganmacs/hub.git" to "https://github.com/NickLaMuro/hub.git" to get my updates.

jeroenvisser101:
@NickLaMuro that's great! You might want to consider opening up a PR from `NickLaMuro:support-github-template` to `github:master` yourself, as it'd probably get merged sooner. Thanks for your effort, it's greatly appreciated!

NickLaMuro:
@ganmacs These changes are excellent.  I am humbled, and my lame attempt at `go` has been out-done by what you have done here!

Thanks for this.  I learned at lot!

jeroenvisser101:
@NickLaMuro I really appreciate your attempt, and since you started all this, I thank you a lot! I don't think it was a lame attempt at all :)

jeroenvisser101:
@mislav is this something that we can get into master?

mislav:
This looks great. Thanks @ganmacs for doing the hard work, @NickLaMuro for in-depth review and contribution, and @jeroenvisser101 for 👀 and pings. I'll take this out for a spin and see how it feels in the text editor, and then merge if I don't find any problems. 👍

jez:
Hey @mislav, have you had some time to poke around with it? What are your thoughts?

I've been poking around myself, and here are some things that I noticed:

- [ ] It doesn't handle case insensitivity ([relevant docs][1])
- [ ] It doesn't handle `.txt` extensions ([relevant docs][1])

It's a simple change to fix these; in fact I've already implemented it here: [jez/hub@2b777f1][2].

As far as I can tell though, it works Well Enough™, so we could probably merge the current version and address these and those that @NickLaMuro mentioned in PR's to follow.

[1]: https://help.github.com/articles/creating-an-issue-template-for-your-repository/
[2]: https://github.com/jez/hub/commit/2b777f1bd6718b31c0c18706b82973557acfb788

jeroenvisser101:
@jez great points, and I agree, if nothing serious comes up, merge and follow up with improvements/fixes :)

jez:
I found and fixed another issue: there was an error with the testing fixture code which was causing a test that should have failed to not fail. I've fixed the failing test, and the faulty fixture code it relied on in [jez/hub@499bfa8][1].

[1]: https://github.com/jez/hub/commit/499bfa821a88931ea9c96bce3c7c8b8005ecf6a4

mislav:
@jez: Thanks for testing and fixes! Feel free to create a branch that adds your fixes on top of `ganmacs:support-github-template`. I can that pull in both this PR and your fixes in the same go.

jez:
I already have a branch on my fork: [`case-insensitive-matching`][1]. Is that what you wanted, or do you want a separate PR?

[1]: https://github.com/jez/hub/tree/case-insensitive-matching

jez:
Bump :D

rpechayr:
@jez I guess you work (on top of  anmacs:support-github-template) need to be in a PR of this repo so that everything can be reviewed and merged altogether.  

This work is actually quite important for my team because we want to enforce a Definition of Done for every PR thought a PR template and we want to be able to do that using hub :)

jez:
@rpechayr, @mislav: I've created a separate PR (see reference above). It includes all the changes here as well as my changes. It's ready to be merged. Any further conversation can probably happen in that PR.

extemporalgenome:
I've been using this patch for a while now, and it does have one issue: if PULL_REQUEST_TEMPLATE.md contains markdown headers (starting with `#`), those lines get stripped out when creating the PR. Similar behavior will occur for any markup language that uses lines starting with the git comment char. To properly handle that, the pull-request command would likely need to suppress insertion of typical git instruction lines, while also suppressing comment-stripping.

jez:
__TL;DR__: That's a feature not a bug. Here's what I mean, and what you can do to use the feature as intended.

@extemporalgenome This is actually expected behavior, and hub has a built-in way of handling it using git's `core.commentChar`. From the manual:

```
core.commentChar
    Commands such as commit and tag that lets you edit messages
    consider a line that begins with this character commented,
    and removes them after the editor returns (default #).
```

If you set it to something like `%`, this will free up `#` to be used as a Markdown ATX header. You can set this option globally (`git config --global`), local to a specific repo (`git config`), or per-command (`git -c core.commentChar='%' ...`). Because most Git syntax highlighters only consider `#` to be a comment character (and I don't want to lose that visual feedback when making commits), I've opted for the 3rd invocation, using these aliases:

```shell
# We want to use '#' as a markdown character, so let's use '%' for comments
alias hubmdpr="hub -c core.commentChar='%' pull-request"
alias hubmdis="hub -c core.commentChar='%' issue create"
```

Additionally, I wrote a Vim plugin that plays nicely with custom comment characters in issue and pull request buffers: [jez/vim-github-hub][1].

[1]: https://github.com/jez/vim-github-hub

github/template.go
dharmik0070: Feature

github/template.go
dharmik0070: Feature

dharmik0070: hhhh
Feature

PR #1149, opened 2016-03-28
Updated hub alias for Fish to suggest defining a function instead of an alias

frederickjh:
I also updated the test but am not sure how to run them.  The formating looks wrong but outputs correctly when the commands are run. 

This is a response to create a pull request in comments on commit  [ fix hub alias for fish](https://github.com/github/hub/commit/df0e8293f1c2f5ee4f3be70462c69d65eb47f39c)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -52,3 +52,6 @@ DEPENDENCIES
   cucumber (~> 1.3.9)
   ronn
   sinatra
+
+BUNDLED WITH
+   1.11.2
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -52,6 +52,3 @@ DEPENDENCIES
   cucumber (~> 1.3.9)
   ronn
   sinatra
-
-BUNDLED WITH
-   1.11.2
--- a/commands/alias.go
+++ b/commands/alias.go
@@ -86,7 +86,7 @@ func alias(command *Command, args *Args) {
 		case "ksh":
 			profile = "~/.profile"
 		case "fish":
-			profile = "~/.config/fish/config.fish"
+			profile = "~/.config/fish/features/git.fish"
 		case "csh":
 			profile = "~/.cshrc"
 		case "tcsh":
@@ -101,7 +101,9 @@ func alias(command *Command, args *Args) {
 		var eval string
 		switch shell {
 		case "fish":
-			eval = `eval (hub alias -s)`
+			eval = `function git --description 'Alias for Github wrapper for git called hub, that adds some Github sugar to git.'
+	hub $argv
+end`
 		case "csh", "tcsh":
 			eval = "eval \"`hub alias -s`\""
 		default:
--- a/commands/alias.go
+++ b/commands/alias.go
@@ -101,7 +101,7 @@ func alias(command *Command, args *Args) {
 		var eval string
 		switch shell {
 		case "fish":
-			eval = `function git --description 'Alias for Github wrapper for git called hub, that adds some Github sugar to git.'
+			eval = `function git --description 'Alias for hub, which wraps git to provide extra functionality with GitHub.'
 	hub $argv
 end`
 		case "csh", "tcsh":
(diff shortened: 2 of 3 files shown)


features/alias.feature
```diff
@@ -15,9 +15,11 @@ Feature: hub alias
     When I successfully run `hub alias`
     Then the output should contain exactly:
       """
-      # Wrap git automatically by adding the following to ~/.config/fish/config.fish:
+      # Wrap git automatically by adding the following to ~/.config/fish/features/git.fish:
 
-      eval (hub alias -s)\n
+function git --description 'Alias for Github wrapper for git called hub, that adds some Github sugar to git.'
+	hub $argv
+end#
```
mislav: Could you change the `#` character to `\n`? Thanks :)

features/alias.feature
```diff
@@ -15,9 +15,11 @@ Feature: hub alias
     When I successfully run `hub alias`
     Then the output should contain exactly:
       """
-      # Wrap git automatically by adding the following to ~/.config/fish/config.fish:
+      # Wrap git automatically by adding the following to ~/.config/fish/features/git.fish:
 
-      eval (hub alias -s)\n
+function git --description 'Alias for Github wrapper for git called hub, that adds some Github sugar to git.'
```
mislav: Could we have a shorter description like:

> Alias for hub, which wraps git to provide extra functionality with GitHub.

mislav:
Thanks, we're close! You'll also have to accept the CLA (find the link in the check failures on this page) before I can accept this 😉

frederickjh:
Hi @mislav !

I made the changes and accepted the CLA, so we should be good to go.

God Bless!
Frederick

commands/alias.go
```diff
@@ -101,7 +101,9 @@ func alias(command *Command, args *Args) {
 		var eval string
 		switch shell {
 		case "fish":
-			eval = `eval (hub alias -s)`
+			eval = `function git --description 'Alias for Github wrapper for git called hub, that adds some Github sugar to git.'
```
mislav: You seem to have only updated the description text in the test but not here as well. Now they don't match and the tests fail.

frederickjh: Sorry about that! I have added a new commit that should fix this.

features/alias.feature
```diff
@@ -15,9 +15,11 @@ Feature: hub alias
     When I successfully run `hub alias`
     Then the output should contain exactly:
       """
-      # Wrap git automatically by adding the following to ~/.config/fish/config.fish:
+      # Wrap git automatically by adding the following to ~/.config/fish/features/git.fish:
 
-      eval (hub alias -s)\n
+function git --description 'Alias for hub, which wraps git to provide extra functionality with GitHub.'
+	hub $argv
```
mislav: This line in Go is now indented with a tab character, while the same line here in the test is indented using spaces, so the test fails because of this mismatch. It's not fault, because it was probably hard to predict that this is going to happen, but it makes sense since we indent Go with tabs and Cucumber stories (`*.features`) with spaces.

Maybe try replacing this indentation with a `\t` string here?

mislav:
You can run tests locally with:

```
script/bootstrap
bin/cucumber features/alias.feature
```

frederickjh:
Thanks for the tip on running the tests locally.  I never did figure out how to have the proper indent in front of 'hub $argv'.

Gemfile.lock
```diff
@@ -52,6 +52,3 @@ DEPENDENCIES
   cucumber (~> 1.3.9)
   ronn
   sinatra
-
-BUNDLED WITH
```
mislav: Could you not include this change? It's not relevant to the topic of your PR.

mislav:
Don't worry about the indent then, it's not your fault. I'll try to solve the indent problem once this PR is ready to 🚢 .

Gemfile.lock
frederickjh: sorry that what I get for doing `git commit -am' commit message'` without doing `git status` and `git dif` first!

mislav: I like to commit with `git commit -vp`. It first asks you for each hunk do you want to stage it (like `git add -p`), and then in renders the full diff of the commit when you're writing the commit message in the text editor. That way you review all that you stage into a commit twice.

frederickjh: Thanks for the tip @mislav! I guess we are always learning new and better ways to do things.

PR #1181, opened 2016-05-16
Add support for pull request templates

Manishearth:
Projects using pull request templates probably want them to get used in `hub` workflow too

The go template could be tweaked a bit

### DIFF
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -1,7 +1,9 @@
 package commands
 
 import (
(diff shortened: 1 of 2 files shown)


mislav:
Thanks for the contribution, but this looks like a duplicate of #1145, which might be a bigger diff but it does include support for issue templates and has tests.

Manishearth:
Oh, thanks.

PR #1189, opened 2016-06-01
Uptade rack to 1.4.6

ecneladis:
Rack `1.4.1` is vulnerable to `CVE-2015-3225`, `CVE-2012-6109`, `CVE-2013-0183`, `CVE-2013-0184`, `CVE-2013-0262`, `CVE-2013-0263`.


mislav:
Thanks for letting us know, but rack is only used in testing when spinning up a temporary server bound to the local interface. It can't be an attack vector, unless someone wanted to hack into their own machine.

PR #1190, opened 2016-06-02
Added release flag to exclude prereleases.

jamesiarmes:
`hub release` currently list all non-draft releases, with an optional flag to include drafts. However, there is no way to list only full releases. This adds an `--exclude-prerelease` flag which omits all prerelease release from the output.

### DIFF
--- a/commands/release.go
+++ b/commands/release.go
@@ -17,7 +17,7 @@ var (
 	cmdRelease = &Command{
 		Run: listReleases,
 		Usage: `
-release [--include-drafts]
+release [--include-drafts, --exclude-prerelease]
 release show <TAG>
 release create [-dp] [-a <FILE>] [-m <MESSAGE>|-F <FILE>] [-c <TARGET>] <TAG>
 release edit [<options>] <TAG>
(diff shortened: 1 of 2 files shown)


mislav:
Thanks for the suggestion and the code! It looks good, but I'm not yet sold on the feature. It feels weird to have one `--include-*` flag and one `--exclude-*` flag. What's the use case for this?

jamesiarmes:
I agree, I'd rather see both of them be `--include-*` flags, but I'm not sure that's possible without altering existing behavior and excluding prereleases by default.

This feature allows a user to list only stable releases, excluding nightlies/alphas/betas/etc. (assuming they have been tagged as prerelease).

glennpratt:
👍 

@mislav any thoughts on what could be changed to make this PR better?

Is `exclude` the issue?  Would `no-include` be better with a plan to make it the default on the next major version?

mislav:
The PR was fine! I might have been too picky earlier. I've changed the flag to `--exclude-prereleases` (plural) for consistency. Thank you @jamesiarmes for the contribution!

PR #1191, opened 2016-06-06
Added command to show the latest release.

jamesiarmes:
This adds a `release latest` command that uses the [`releases/latest`](https://developer.github.com/v3/repos/releases/#get-the-latest-release) endpoint to retrieve the latest full release and output the results using the same format as `release show`.


glennpratt:
Bump

mislav:
I've just pulled a change that adds the `hub release --exclude-prereleases` flag. This way, you get an ordered list of all stable releases, newest first. Then, showing the details of the latest release can be accomplished simply as:

```
hub release show `hub release --exclude-prereleases | head -1`
```

I'm inclined to close this PR in favor of the existing solution. What do others think?

jamesiarmes:
This command acts differently in that it uses the actual `releases/latest` API endpoint, which is like calling `hub release show` on the release and is analogous to the [releases/latest](https://github.com/github/hub/releases/latest) url on repos. The alternative, assuming you don't know the name of the latest release, is to run two commands (`hub release` followed by `hub show`). While certainly not the end of the world, this combines those two steps in to one which is especially useful for automated tools.

This should also be more efficient as it only ever returns the single release, while `hub release --exclude-prereleases` will return a larger list of commits and then filter them locally.

PR #1204, opened 2016-07-01
Trivial refactoring of checkout command

huydx:
Just a very trivial refactoring of checkout command logic


mislav:
Good catch; thanks!

PR #1207, opened 2016-07-07
Explicitly specify binstub path in scripts/bootstrap

voidus:
There might be a bundle config that specifies a different path, which
breaks the build. Explicitly specifying the path changes that.


mislav:
Makes sense! Thanks

PR #1208, opened 2016-07-11
README.md: update of git and hub versions

thedanotto:
Simply updated the git and hub versions in the `README.md`.


mislav:
No sorry! It's not strictly necessary to keep these versions up to date, otherwise we would have to edit the README after every release of hub *and* git.

PR #1229, opened 2016-08-12
Fix CreatePullRequest to expect 201 over 200

### LINKED ISSUE #1226: HTTP 201 when running `hub pull-request --browse`
Seems to happen every time (the pull request itself does go through):
```
Error creating pull request: Created (HTTP 201)
Error: Failure while executing: hub pull-request --browse -m clipper\ 0.3
```
This is a regression since hub 2.2.3.

arbourd:
According to Github's API docs, creating a pull request should yield a
status code of 201 and not 200.

Closes #1226

### DIFF
--- a/features/pull_request.feature
+++ b/features/pull_request.feature
@@ -25,6 +25,7 @@ Feature: hub pull-request
         assert :base  => 'master',
                :head  => 'Manganeez:master',
                :title => 'here we go'
+        status 201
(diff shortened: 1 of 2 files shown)


mislav:
Thanks! Will make a patch release in a few minutes.

mislav:
This is now v2.2.5. Thanks for catching this nasty bug!

PR #1234, opened 2016-08-18
Add missing filters when listing issues.

pcorpet:
Useful for #1099

### DIFF
--- a/commands/issue.go
+++ b/commands/issue.go
@@ -47,7 +47,7 @@ With no arguments, show a list of open issues.
 
 	-f, --format <FORMAT>
 		Pretty print the contents of the issues using format <FORMAT> (default:
-		"%sC%>(8)%i%Creset	%t%	l%n"). See the "PRETTY FORMATS" section of the
+		"%sC%>(8)%i%Creset  %t%  l%n"). See the "PRETTY FORMATS" section of the
 		git-log manual for some additional details on how placeholders are used in
 		format. The available placeholders for issues are:
 
@@ -196,7 +196,7 @@ func listIssues(cmd *Command, args *Args) {
 			}
 		}
 		// Unfortunately hub does not use the same flag ("label") than the GitHub
-		// API ("labels")
+		// API ("labels").
 		if cmd.FlagPassed("label") {
 			filters["labels"] = flagIssueLabelsFilter
 		}
--- a/commands/issue.go
+++ b/commands/issue.go
@@ -17,7 +17,7 @@ var (
 	cmdIssue = &Command{
 		Run: listIssues,
 		Usage: `
-issue [-a <ASSIGNEE>] [-c <CREATOR>] [-@ <USER] [-s <STATE>] [-f <FORMAT>] [-M <MILESTONE>] [-l <LABELS>]
+issue [-a <ASSIGNEE>] [-c <CREATOR>] [-@ <USER] [-s <STATE>] [-f <FORMAT>] [-M <MILESTONE>] [-l <LABELS>] [-t <TIME>]
 issue create [-o] [-m <MESSAGE>|-F <FILE>] [-a <USERS>] [-M <MILESTONE>] [-l <LABELS>]
 `,
 		Long: `Manage GitHub issues for the current project.
@@ -115,6 +115,10 @@ With no arguments, show a list of open issues.
 		Display only issues with certain labels.
 
 		When opening an issue, add a comma-separated list of labels to this issue.
+
+	-t, --since <TIME>
+		Display only issues updated at or after a certain time. The time is a
+		timestamp in ISO-8601 format: YYYY-MM-DDTHH:MM:SSZ.
 `,
 	}
 
@@ -133,6 +137,7 @@ With no arguments, show a list of open issues.
 	flagIssueCreator,
 	flagIssueMentioned,
 	flagIssueLabelsFilter,
+	flagIssueSince,
 	flagIssueFile string
 
 	flagIssueBrowse bool
@@ -158,6 +163,7 @@ func init() {
 	cmdIssue.Flag.StringVarP(&flagIssueCreator, "creator", "c", "", "CREATOR")
 	cmdIssue.Flag.StringVarP(&flagIssueMentioned, "mentioned", "@", "", "USER")
 	cmdIssue.Flag.StringVarP(&flagIssueLabelsFilter, "label", "l", "", "LABELS")
+	cmdIssue.Flag.StringVarP(&flagIssueSince, "since", "t", "", "TIME")
 
 	cmdIssue.Use(cmdCreateIssue)
 	CmdRunner.Use(cmdIssue)
@@ -181,6 +187,7 @@ func listIssues(cmd *Command, args *Args) {
 			"milestone": flagIssueMilestoneFilter,
 			"creator":   flagIssueCreator,
 			"mentioned": flagIssueMentioned,
+			"since":     flagIssueSince,
 		}
 		filters := map[string]interface{}{}
 		for flag, filter := range flagFilters {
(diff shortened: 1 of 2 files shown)


commands/issue.go
```diff
@@ -101,10 +107,18 @@ With no arguments, show a list of open issues.
...
+		When opening an issue, add this issue to a GitHub milestone with id <ID>.
 
 	-l, --labels <LABELS>
-		Add a comma-separated list of labels to this issue.
+		Display only issues with certain labels.
+
+		When opening an issue, add a comma-separated list of labels to this issue.
+
+	-t, --since <TIME>
+		Display only issues updated at or after a certain time. The time is a
```
mislav: Maybe instead of "time" we can say "date" and allow a date input without a time?

commands/issue.go
```diff
@@ -101,10 +107,18 @@ With no arguments, show a list of open issues.
...
+		timestamp in ISO-8601 format: YYYY-MM-DDTHH:MM:SSZ.
```
mislav: We could also allow date input in format of "YYYY-MM-DD" and default the hour to `00:00:00`.

commands/issue.go
```diff
@@ -157,12 +181,24 @@ func listIssues(cmd *Command, args *Args) {
...
+			"creator":   flagIssueCreator,
+			"mentioned": flagIssueMentioned,
+			"since":     flagIssueSince,
+		}
 		filters := map[string]interface{}{}
-		if cmd.FlagPassed("state") {
-			filters["state"] = flagIssueState
+		for flag, filter := range flagFilters {
+			if cmd.FlagPassed(flag) {
+				filters[flag] = filter
```
mislav: Looks good! 👍

commands/issue.go
```diff
@@ -140,6 +159,11 @@ func init() {
 	cmdIssue.Flag.StringVarP(&flagIssueAssignee, "assignee", "a", "", "ASSIGNEE")
 	cmdIssue.Flag.StringVarP(&flagIssueState, "state", "s", "", "STATE")
 	cmdIssue.Flag.StringVarP(&flagIssueFormat, "format", "f", "%sC%>(8)%i%Creset  %t%  l%n", "FORMAT")
+	cmdIssue.Flag.StringVarP(&flagIssueMilestoneFilter, "milestone", "M", "", "MILESTONE")
+	cmdIssue.Flag.StringVarP(&flagIssueCreator, "creator", "c", "", "CREATOR")
+	cmdIssue.Flag.StringVarP(&flagIssueMentioned, "mentioned", "@", "", "USER")
+	cmdIssue.Flag.StringVarP(&flagIssueLabelsFilter, "label", "l", "", "LABELS")
```
mislav: Does the filtering option support multiple labels? Then this flag should be `--labels` in long format.

commands/issue.go
```diff
@@ -157,12 +181,24 @@ func listIssues(cmd *Command, args *Args) {
...
-		if cmd.FlagPassed("state") {
-			filters["state"] = flagIssueState
+		for flag, filter := range flagFilters {
+			if cmd.FlagPassed(flag) {
+				filters[flag] = filter
+			}
 		}
-		if cmd.FlagPassed("assignee") {
-			filters["assignee"] = flagIssueAssignee
+		// Unfortunately hub does not use the same flag ("label") than the GitHub
```
mislav: I don't understand what this comment is saying

features/issue.feature
```diff
@@ -31,6 +31,116 @@ Feature: hub issue
...
     And the exit status should be 0
 
+  Scenario: Fetch issues not assigned to any milestone
+    Given the GitHub API server:
+    """
+    get('/repos/github/hub/issues') {
+      assert :milestone => "none"
+
+      json [
+        { :number => 102,
```
mislav: For the tests that assert a specific filter parameter has been passed, you don't actually have to return any results. You can return an empty array.

features/issue.feature
```diff
@@ -31,6 +31,116 @@ Feature: hub issue
...
+          :title => "First issue",
+          :state => "open",
+          :user => { :login => "octocat" },
+        },
+      ]
+    }
+    """
+    When I run `hub issue -M none`
```
mislav: You can change this step to ``When I successfully run `hub issue -M none` `` and you can not assert any output at all. We are just testing that the API endpoint received the correct filtering parameters. If it didn't, the fake API endpoint will return HTTP 400, and the test will fail. We don't need to actually return any records or assert any output here, or in any similar tests.

mislav:
This looks great! Thanks for working on this.

commands/issue.go
pcorpet: I agree, however, I preferred to stay consistant with the label flag above. If you think it's OK not to be backward compatible we could fix both to `labels`. If you think it's OK not to be consistent we could have both: `label` when creating an issue, `labels` when filtering.

mislav: We don't have to care about backwards compatibility since the feature was in beta before. It's fine and easier for everyone that both are `--labels`.

features/issue.feature
pcorpet: Done.

commands/issue.go
pcorpet: Done.

commands/issue.go
pcorpet: The GitHub API already supports the YYYY-MM-DD format, so I've updated the doc, but no need to write more code.

commands/issue.go
pcorpet: Gone with the wind.

features/issue.feature
pcorpet: Done. However, we're not sure, by this test alone, that the assert is actually called.

commands/issue.go
pcorpet: Yes, I like `labels` much better. I forgot it was beta before. One less technical debt.

commands/issue.go
```diff
@@ -157,12 +182,20 @@ func listIssues(cmd *Command, args *Args) {
...
-		if cmd.FlagPassed("state") {
-			filters["state"] = flagIssueState
+		flagFilters := map[string]string{
+			"state":     flagIssueState,
+			"assignee":  flagIssueAssignee,
+			"milestone": flagIssueMilestoneFilter,
+			"creator":   flagIssueCreator,
+			"mentioned": flagIssueMentioned,
+      "labels":    flagIssueLabelsFilter,
+			"since":     flagIssueSince,
```
mislav: I still think we need to process `flagIssueSince` a bit before passing it on to the API. The reason is timezones.

A person in New Zealand might use `2016-08-19` to get all isues on or after Friday. However, when the API interprets that date, most likely in the PST timezone, the results returned from start of Friday will be very different, ignoring most issues opened on Friday in New Zealand.

We have to assume that when a person doesn't specify a time or timezone offset, `2016-08-19` means “the start of the 19th of August for me in my local timezone“. So, we should parse the date in the local timezone, have it default to `00:00:00` for time, and serialize it into ISO 8601 format when passing to API so the time is absolutely precise and unambiguous.

features/issue.feature
```diff
@@ -31,6 +31,66 @@ Feature: hub issue
...
+    Given the GitHub API server:
+    """
+    get('/repos/github/hub/issues') {
+      assert :milestone => "none"
+
+      json []
+    }
+    """
+    When I run `hub issue -M none`
+    Then the exit status should be 0
```
mislav: Instead of checking if exit status was 0, the previous step could be "When **successfully** I run"

features/issue.feature
```diff
@@ -31,6 +31,66 @@ Feature: hub issue
...
+
+  Scenario: Fetch issues created by a given user
+    Given the GitHub API server:
+    """
+    get('/repos/github/hub/issues') {
+      assert :creator => "octocat"
+
```
mislav: No need for empty line here, since the API specification is so short anyway

PR #1235, opened 2016-08-18
Reimplement hub checkout to avoid creating unnecessary remotes

pcorpet:
Update hub checkout not to create any new remotes thus leaving less tracks behind.

### DIFF
--- a/commands/checkout.go
+++ b/commands/checkout.go
@@ -94,16 +94,20 @@ func transformCheckoutArgs(args *Args) error {
 
 	_, err = repo.RemoteByName(user)
 	if err == nil {
+		// The remote for the head of the PR is already tracked by the current git workdir.
 		args.Before("git", "remote", "set-branches", "--add", user, branch)
 		remoteURL := fmt.Sprintf("+refs/heads/%s:refs/remotes/%s/%s", branch, user, branch)
 		args.Before("git", "fetch", user, remoteURL)
-	} else {
-		u := url.Project.GitURL(headRepo.Name, user, headRepo.Private)
-		args.Before("git", "remote", "add", "-f", "--no-tags", "-t", branch, user, u)
-	}
 
-	remoteName := fmt.Sprintf("%s/%s", user, branch)
-	replaceCheckoutParam(args, checkoutURL, newBranchName, remoteName)
+		remoteName := fmt.Sprintf("%s/%s", user, branch)
+		replaceCheckoutParam(args, checkoutURL, "--track", "-B", newBranchName, remoteName)
+	} else {
+		// Let's fetch the head directly without adding a remote nor tracking.
+		branchMapping := fmt.Sprintf("pull/%s/head:%s", id, newBranchName)
+		args.Before("git", "fetch", url.GitURL("", "", false), branchMapping)
+
+		replaceCheckoutParam(args, checkoutURL, newBranchName)
+	}
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


commands/checkout.go
```diff
@@ -94,16 +94,20 @@ func transformCheckoutArgs(args *Args) error {
 
 	_, err = repo.RemoteByName(user)
 	if err == nil {
+		// The remote for the head of the PR is already tracked by the current git workdir.
 		args.Before("git", "remote", "set-branches", "--add", user, branch)
```
mislav: Let's remove this `git remote set-branches` code path completely too.

commands/checkout.go
```diff
@@ -94,16 +94,20 @@ func transformCheckoutArgs(args *Args) error {
...
+
+		remoteName := fmt.Sprintf("%s/%s", user, branch)
+		replaceCheckoutParam(args, checkoutURL, "--track", "-B", newBranchName, remoteName)
 	} else {
-		u := url.Project.GitURL(headRepo.Name, user, headRepo.Private)
-		args.Before("git", "remote", "add", "-f", "--no-tags", "-t", branch, user, u)
-	}
+		// Let's fetch the head directly without adding a remote nor tracking.
+		branchMapping := fmt.Sprintf("pull/%s/head:%s", id, newBranchName)
+		args.Before("git", "fetch", url.GitURL("", "", false), branchMapping)
```
mislav: Instead of git URL, can we pull from existing remote name?

features/checkout.feature
```diff
@@ -77,26 +77,8 @@ Feature: hub checkout <PULLREQ-URL>
       }
       """
     When I run `hub checkout -f https://github.com/mojombo/jekyll/pull/77 -q`
-    Then "git remote add -f --no-tags -t fixes mislav git@github.com:mislav/jekyll.git" should be run
-    And "git checkout -f --track -B mislav-fixes mislav/fixes -q" should be run
-
-  Scenario: Custom name for new branch
```
mislav: Oops I didn't know we had 2 scenarios with the same name!

mislav:
Nice work! I have only a few nitpicks, but this feels great in general.

commands/checkout.go
pcorpet: Done.

commands/checkout.go
pcorpet: Done.

commands/checkout.go
```diff
@@ -92,18 +92,18 @@ func transformCheckoutArgs(args *Args) error {
...
 	utils.Check(err)
 
+	var remoteRepo string
+
 	_, err = repo.RemoteByName(user)
 	if err == nil {
-		args.Before("git", "remote", "set-branches", "--add", user, branch)
-		remoteURL := fmt.Sprintf("+refs/heads/%s:refs/remotes/%s/%s", branch, user, branch)
-		args.Before("git", "fetch", user, remoteURL)
+		remoteRepo = user
```
mislav: I don't think this is accurate anymore. We don't want to fetch from a remote that points to the head repo of the PR. We always want to fetch from the origin remote that represents the "main" project.

Thus, the process should be:

1. Get the **base** repo of the PR
2. Find the remote that points to that repo (will usually resolve to "origin" or "upstream")
3. If the remote can't be found, abort the action.

mislav:
@pcorpet Thanks for starting this! I've finalized the implementation in https://github.com/github/hub/compare/2157ad1...58d332d.

Issue #877, opened 2015-04-23
pull-request for unpushed branch opens message editor with unrelated remote


betelgeuse:
Steps to reproduce:
 1. git checkout -b feature_branch org/master
 2. git commit
 3. hub pull-request -b org:master

This results in the following prepopulated message:

```
  1 
  2 # Requesting a pull to org:master from unrelated:feature_branch
  3 #
  4 # Write a message for this pull request. The first block
  5 # of text is the title and the rest is description.
```

The .git/config for the feature_branch is

```
[branch "feature_branch"]
	remote = org
	merge = refs/heads/master
```

My remotes are listed as follows on git remote -v 

```
org	git@github.com:xxx (fetch)
org	git@github.com:xxx (push)
unrelated	git@github.com:xxx (fetch)
unrelated	git@github.com:xxx (push)
unrelated2	git@github.com:xxx (push)
unrelated2	git@github.com:xxx (fetch)
myfork	git@github.com:xxx (fetch)
myfork	git@github.com:xxx (push)
unrelated3	git@github.com:xxx (fetch)
unrelated3	git@github.com:xxx (push)
unrelated4	git@github.com:xxx (push)
unrelated4	git@github.com:xxx(fetch)
unrelated5	git@github.com:xxx (fetch)
unrelated5	git@github.com:xxx (push)
```

I am guessing it picks up first that does not match base from the list. I think it should rather fail if it finds there's more than one potential remote.

danmbox:
... and org was what the world calls "origin", which you for some reason renamed to make things interesting? :)

betelgeuse:
@danmbox because I keep it named after organization which I abbreviated org. I think origin is misleading when you have many remotes. Might be just me though.

danmbox:
Looking at the report I wasn't at first sure whether "org" was just origin, the origin's github username, or the fork author's username. But thankfully you have the remote list at the end :)

mislav:
What is your `git config push.default` and `hub version`? Thanks

betelgeuse:
```
 $ git config push.default
 $ hub version
git version 2.3.2
hub version 2.2.1
```

mislav:
[...]

PR #1239, opened 2016-08-22
Feature to exercise #877

tiennou:
This is a simplification of my failing #877 workflow, so at least I can be sure that it's not some unrelated, local-only occurrence.

### DIFF
--- a/features/pull_request.feature
+++ b/features/pull_request.feature
@@ -646,6 +646,42 @@ BODY
     When I successfully run `hub pull-request -m hereyougo`
     Then the output should contain exactly "the://url\n"
 
+  Scenario: Create pull request to non-standard remote
+    Given there are no "origin" remote
+    And the "github" remote has url "git://github.com/github/coral.git"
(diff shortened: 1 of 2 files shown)


tiennou:
At least here it seems to fail with what I'm seeing when I'm "workflowing", eg the automatic remote selection gets confused and uses whatever remote is the first one.

mislav:
Thanks for the failing test! The head/base resolution for PR is broken is all sorts of ways right now and is high on my list to fix next. There are so many edge-case problems with it, that I'll probably end up rewriting the whole head/base resolution logic instead of tweaking existing code. However,  I won't be able to get to this before 2 weeks form now, because I'm offline on vacation.

Until then, whenever you have a problem, simply specify head (and/or base) explicitly.

tiennou:
No problem, it can wait. I may try to take a look, but I'm not a Go-er, so that might be out of my reach, we'll see. Have a nice vacation, sir !

mislav:
I was not a Go-er until somone entirely rewrote hub from Ruby into Go and it was faster ;)

PR #1245, opened 2016-08-27
Add sorting options to `hub issue`.

pcorpet:


commands/issue.go
```diff
@@ -121,6 +121,12 @@ With no arguments, show a list of open issues.
 
 	-d, --since <DATE>
 		Display only issues updated on or after <DATE> in ISO 8601 format.
+
+	-o, --sort <SORT_KEY>
+		Sort displayed issues by "created" (default), "updated" or "comments".
+
+	-^ --sort-ascending
```
mislav: Interesting choice for a shorthand! What made you think of `-^`?

features/issue.feature
```diff
@@ -80,6 +80,34 @@ Feature: hub issue
     """
     When I successfully run `hub issue -d 2016-08-18T09:11:32Z`
 
+  Scenario: Fetch issues
```
mislav: The scenario name needs to be more specific, like "Fetch issues sorted by number of comments ascending"

features/issue.feature
```diff
@@ -80,6 +80,34 @@ Feature: hub issue
...
+    Given the GitHub API server:
+    """
+    get('/repos/github/hub/issues') {
+      assert :sort => "comments"
+      assert :direction => "asc"
+
+      json [
```
mislav: You don't have to return any issues here. You can return a blank array.

commands/issue.go
pcorpet: this is for sorting, I thought about Sorted (s is taken), then sOrted (o is taken), then sort Ascending (a is taken), and then tried to thinkg outside of the box and thought about the little chevron shown in lists when you want to change ascending/descending sort. I'll take any better idea you can come up with.

features/issue.feature
```diff
@@ -80,6 +80,34 @@ Feature: hub issue
...
+        { :number => 102,
+          :title => "First issue",
+          :state => "open",
+          :user => { :login => "octocat" },
+        },
+      ]
+    }
+    """
+    When I successfully run `hub issue -o comments -^`
+    Then the output should contain exactly:
```
mislav: You don't have to assert any output here. The `When I successfully run…` line already verified that the sort critera was passed correctly. Otherwise, the fake API server would have returned a HTTP 400, and the command would have failed.

features/issue.feature
pcorpet: I prefer returning an array to make sure that the sort is done by GitHub and that the order is kept exactly as such by hub.

commands/issue.go
```diff
@@ -192,6 +202,7 @@ func listIssues(cmd *Command, args *Args) {
 			"creator":   flagIssueCreator,
 			"mentioned": flagIssueMentioned,
 			"labels":    flagIssueLabelsFilter,
+			"sort":      flagIssueSort,
```
mislav: This always adds `sort: created` as the default parameter. We should verify this in one of the existing cukes by adding:

```
assert :sort => "created"
assert :direction => nil
```

features/issue.feature
mislav: I don't think that's necessary. We always display issues in the order returned by the server. Other cukes verify that.

mislav:
@pcorpet Do you have the time to finalize this PR in the next days? At any point, just give me the signal that I should take over. This is good stuff in general, but there are points that I wanted you to address.

pcorpet:
Yep, I will address those comments.

mislav:
@pcorpet Thanks; greatly appreciated!

BTW, completely unrelated: I landed [copy to clipboard](https://github.com/github/hub/pull/1264) and [`hub sync`](https://github.com/github/hub/pull/1265) over the weekend. The next release is going to be massive! 🚀

commands/issue.go
pcorpet: It doesn't: line 209 I check if the flag was passed before adding the sort parameter. But I'm going to add the check that sort and direction params are nil.

features/issue.feature
pcorpet: Done.

features/issue.feature
pcorpet: OK, done. (my initial thought was that a naive implementation could have tried to sort the results locally, and I wanted to test against that)

features/issue.feature
pcorpet: Done.

commands/issue.go
pcorpet: Done.

pcorpet:
Congrats on copy to clipboard and sync. Let's make it happen!

mislav:
Thank you!!

Issue #1206, opened 2016-07-07
Pull request checkout with no 'leftovers'


rcorre:
I often checkout a pull request as a one-time thing. I want to take a look at the code and run it, but afterwards I'll switch back to my own branches and don't want a new remote or branch hanging around. It would be nice to have `--no-remote` and `--no-branch` flags for `hub checkout`.

I would expect `--no-remote` to do:

    git fetch <remote-url>
    git branch user-branchname FETCH_HEAD
    git checkout user-branchname
    
With `--no-branch` it would be:

    git fetch <remote-url>
    git checkout FETCH_HEAD

Does this sound useful? I'd be happy to take a stab at it (I'd like to learn some go).

mislav:
That's not a bad idea. I definitely already have a plan to change the `checkout` command to never add a remote, since those can pile up over time and are not very useful. The branch is another topic, though. By default a branch will definitely be created, but I'll consider your `--no-branch` option to avoid creating a branch and put you to detached head instead.

Since I plan to refactor `checkout` significantly, you don't have to make a contribution right now, since it will likely conflict with the other changes. So don't worry about making the PR right now! But If you'd like to learn some Go, there are definitely other issues in the projects that need attention… 😉

rcorre:
Thanks, sounds good to me.

mcornella:
+1.

The [official github docs](https://help.github.com/articles/checking-out-pull-requests-locally/#modifying-an-inactive-pull-request-locally) uses the command 
```
git fetch origin pull/ID/head:BRANCHNAME
```

which avoids the `git branch` call.

Thanks for submitting this!

rcorre:
Oh, that's clever, and a good workkaround in the meanwhile. Thanks!

mcornella:
Yay thanks! :tada:

mislav:
Mind that the new implementation still leaves a branch around. But, most importantly, it doesn't create any git remotes anymore.

mcornella:
That's exactly what I'd expect as a user so thanks!

rcorre:
Thanks! Not having a remote is nice, but is there any chance of adding a `no-branch` option to fetch to `FETCH_HEAD`?

mislav:
@rcorre That shouldn't be too hard to add. Please take a look at the current `checkout` implementation and see if you can find your way around adding such a feature. If it's too confusing, let me know!

rcorre:
I'll give it a shot, thanks!

[...]

PR #1250, opened 2016-08-31
Checkout no branch

rcorre:
Initial draft for the `checkout --no-branch <PULLREQUEST>` as discussed in #1206

### DIFF
--- a/commands/checkout.go
+++ b/commands/checkout.go
@@ -11,8 +11,12 @@ import (
 var cmdCheckout = &Command{
 	Run:          checkout,
 	GitExtension: true,
-	Usage:        "checkout <PULLREQ-URL> [<BRANCH>]",
+	Usage:        "checkout [--no-branch] <PULLREQ-URL> [<BRANCH>]",
 	Long: `Check out the head of a pull request as a local branch.
+
+## Options:
+	--no-branch
+		Use FETCH_HEAD instead of creating a local branch.
 
 ## Examples:
 		$ hub checkout https://github.com/jingweno/gh/pull/73
@@ -25,7 +29,11 @@ hub-merge(1), hub-am(1), hub(1), git-checkout(1)
 `,
 }
 
+var flagCheckoutNoBranch bool
+
 func init() {
+	cmdCheckout.Flag.BoolVar(&flagCheckoutNoBranch, "no-branch", false, "")
+
 	CmdRunner.Use(cmdCheckout)
 }
 
@@ -74,6 +82,11 @@ func transformCheckoutArgs(args *Args) error {
 		return err
 	}
 
+	if idx := args.IndexOfParam("--no-branch"); idx >= 0 {
+		args.RemoveParam(idx)
+		flagCheckoutNoBranch = true
+	}
+
 	if idx := args.IndexOfParam(newBranchName); idx >= 0 {
 		args.RemoveParam(idx)
 	}
@@ -102,8 +115,13 @@ func transformCheckoutArgs(args *Args) error {
 		if newBranchName == "" {
 			newBranchName = fmt.Sprintf("%s-%s", pullRequest.Head.Repo.Owner.Login, pullRequest.Head.Ref)
 		}
-		refSpec = fmt.Sprintf("pull/%s/head:%s", id, newBranchName)
-		newArgs = append(newArgs, newBranchName)
+		if flagCheckoutNoBranch {
+			refSpec = fmt.Sprintf("pull/%s/head", id)
+			newArgs = append(newArgs, "FETCH_HEAD")
+		} else {
+			refSpec = fmt.Sprintf("pull/%s/head:%s", id, newBranchName)
+			newArgs = append(newArgs, newBranchName)
+		}
 	}
 
 	args.Before("git", "fetch", remote.Name, refSpec)
--- a/commands/checkout.go
+++ b/commands/checkout.go
@@ -114,6 +114,8 @@ func transformCheckoutArgs(args *Args) error {
 	} else {
 		if newBranchName == "" {
 			newBranchName = fmt.Sprintf("%s-%s", pullRequest.Head.Repo.Owner.Login, pullRequest.Head.Ref)
+		} else if flagCheckoutNoBranch {
+			return fmt.Errorf("Cannot specify a branch name and --no-branch flag")
 		}
 		if flagCheckoutNoBranch {
 			refSpec = fmt.Sprintf("pull/%s/head", id)
--- a/commands/checkout.go
+++ b/commands/checkout.go
@@ -107,10 +107,17 @@ func transformCheckoutArgs(args *Args) error {
 	if pullRequest.IsSameRepo() {
 		if newBranchName == "" {
 			newBranchName = pullRequest.Head.Ref
+		} else if flagCheckoutNoBranch {
+			return fmt.Errorf("Cannot specify a branch name and --no-branch flag")
 		}
 		remoteBranch := fmt.Sprintf("%s/%s", remote.Name, pullRequest.Head.Ref)
-		refSpec = fmt.Sprintf("+refs/heads/%s:refs/remotes/%s", pullRequest.Head.Ref, remoteBranch)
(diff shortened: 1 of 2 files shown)


commands/checkout.go
```diff
@@ -74,6 +82,11 @@ func transformCheckoutArgs(args *Args) error {
 		return err
 	}
 
+	if idx := args.IndexOfParam("--no-branch"); idx >= 0 {
```
rcorre: I feel like I shouldn't have to do this. I registered the flag as a `BoolVar` like other commands do, but it wasn't parsing out `--no-branch` automatically.

rcorre:
I have a few questions on the current implementation:

1. Why handle the 'same repo' case differently? It seems like needless complication.

2. Why error out if you haven't already added the remote? It seems like you could just use `fetch <remoteurl>` and not worry about matching the url to an existing remote.

3. Why not have `hub checkout` work more like `git checkout`? 

| command                    | creates a new ref|
|----------------------------|------------------|
| `git checkout`             | no               |
| `git checkout -b`          | yes              |
| `hub checkout`             | yes              |
| `hub checkout --no-branch` | no               |
| `hub checkout -b`          | disallowed       |

It seems a little backwards and unintuitive to me. What if `hub checkout` just checked out a freshly fetched PR, and `hub checkout -b` checked it out into a new local branch? It seems like it would parallel git more closely and would actually require less work (you could pass `-b` right through to `git`).

For 1 & 2 I seem to remember `hub checkout` working like this before, so maybe I missed a discussion on why it was changed.

commands/checkout.go
mislav: hub preserves all given flags by default, even non-standard ones. Because git doesn't recognize this flag, we have to remove it. You're doing all right here.

commands/checkout.go
```diff
@@ -74,6 +82,11 @@ func transformCheckoutArgs(args *Args) error {
...
+		args.RemoveParam(idx)
+		flagCheckoutNoBranch = true
```
mislav: This line should not be needed, since `flagCheckoutNoBranch` value should be already updated by the flag parsing mechanism.

features/checkout.feature
```diff
@@ -120,3 +120,104 @@ Feature: hub checkout <PULLREQ-URL>
     When I run `hub checkout https://github.com/mojombo/jekyll/pull/77 mycustombranch`
     Then "git fetch origin +refs/heads/fixes:refs/remotes/origin/fixes" should be run
     And "git checkout -b mycustombranch --track origin/fixes" should be run
+
+  Scenario: Checkout a pull request with --no-branch
+    Given the GitHub API server:
+      """
+      get('/repos/mojombo/jekyll/pulls/77') {
+        halt 406 unless request.env['HTTP_ACCEPT'] == 'application/vnd.github.v3+json;charset=utf-8'
```
mislav: You don't need to repeat this assertion in this cuke.

features/checkout.feature
```diff
@@ -120,3 +120,104 @@ Feature: hub checkout <PULLREQ-URL>
...
+      """
+    When I run `hub checkout --no-branch https://github.com/mojombo/jekyll/pull/77 -q`
+    Then "git fetch origin pull/77/head" should be run
+    And "git checkout FETCH_HEAD -q" should be run
+
+  Scenario: Checkout a pull request with --no-branch and a branch name
+    Given the GitHub API server:
+      """
+      get('/repos/mojombo/jekyll/pulls/77') {
+        halt 406 unless request.env['HTTP_ACCEPT'] == 'application/vnd.github.v3+json;charset=utf-8'
```
mislav: Don't need this either.

mislav: Actually, the whole `Given the GitHub API server` block should not be needed at all in this scenario, since the flag parsing should abort the command early and no API communication should take place. I think.

features/checkout.feature
```diff
@@ -120,3 +120,104 @@ Feature: hub checkout <PULLREQ-URL>
...
+          }
+        }
+      }
+      """
+    When I run `hub checkout --no-branch https://github.com/mojombo/jekyll/pull/77`
+    Then "git fetch origin refs/heads/fixes" should be run
+    And "git checkout FETCH_HEAD" should be run
+
+  Scenario: Same-repo with --no-branch and a branch name
+    Given the GitHub API server:
```
mislav: Ditto about the fake API server. It might not be needed.

mislav:
> It seems a little backwards and unintuitive to me.

You raise some valid points. `hub checkout <PR-URL>` is an old command that only served as a quick and dirty shortcut. Over time, more features and edge cases got added to it. Now, it might be a little counter-intuitive.

What would you suggest as a new API? Keep in mind that we need to both cover cases where you choose to checkout a branch that's automatically named, and the case where you want to specify the name of the new branch specifically.

features/checkout.feature
rcorre: I would expect that too ... but it seems to go through with fetching: `Error getting pull request: Get https://api.github.com/repos/mojombo/jekyll/pulls/77: dial tcp 127.0.0.1:0: getsockopt: connection refused`

commands/checkout.go
rcorre: That's what I meant above ... for some reason it isn't getting set, but it seems to work in other commands. I must be missing something...

rcorre:
> What would you suggest as a new API? 

I'm speculating that there are two use cases:

`hub checkout <URL>` -> `git fetch URL BRANCH && git checkout FETCH_HEAD`

This lets me take a quick look at a PR. When I'm done, I can run (say) `git checkout -` to return to my previous branch. There are no remotes or branches left over (even under `refs/remotes`). If I decide I want to do work on it, I can always run `checkout -b` later.

`hub checkout <URL> -b mybranch` -> `git fetch URL BRANCH && git checkout FETCH_HEAD -b mybranch`

This is what I'd run when I want to 'work' on a PR by creating a local branch. The beauty of this is that `hub` needs no special handling. It can just pass the `-b` along to `git`.

> Keep in mind that we need to both cover cases where you choose to checkout a branch that's automatically named, and the case where you want to specify the name of the new branch specifically.

At least for me, an auto-named local branch is more likely to create clutter than serve any useful purpose.
I think the above two forms would cover most use cases.

mislav:
> At least for me, an auto-named local branch is more likely to create clutter than serve any useful purpose.

I can see your argument. But, there is the case where you are checking out a PR coming from the local repo. In that case, the branch is automatically named the same as the `head` branch and set up to track `refs/remotes/origin/BRANCH` so that you can update same branch using `git pull/push`. So, this use-case is a strong argument for the auto-naming case.

rcorre:
Ah, that explains the same-repo logic, I wasn't sure why that was special-cased. I don't have a problem with it creating entries under `refs/remotes`, I'm mainly worried about `refs/heads`. If we keep the current behavior of creating remote branches, then `checkout -b mybranch` will set up tracking info if you have `autosetupmerge=true`.

dwijnand:
It's a shame this never merged.

rcorre:
I haven't been using `hub` lately, but someone else is welcome to take this up if there is still demand for the feature.

dwijnand:
Would `git pr fetch 1250` (or `git fetch https://github.com/owner/repo/pull/1250`) be acceptable?

It would save me (and others) from remembering it's `refs/pull/1250/head` which is relevant for:
* this use case (as you mentioned), and
* being able to fetch the branch of a PR that was squash-merged to be able to blame it

mislav:
I could be down with `hub pr fetch 1250`, but I'm reluctant to maintain a new subcommand just to save someone a few keystrokes. Btw, if the intended use for this is scripting, then wouldn't it be just as easy to write `git fetch origin refs/pull/1250/head` one time as it is to write `hub pr fetch 1250`?

You're welcome to add a new `pr fetch` subcommand and send us a PR for review. 👍

dwijnand:
I haven't considered how I'd wrap the behaviour in a script, so it would be manual use.

OK, thanks, I'll come back if I get angry enough about it missing 😄

PR #1252, opened 2016-08-31
Read templates from ~/.github

sorenmat:
First try to locate a template in the repository .github, if
that fails try to find the templates in $home/.github

### DIFF
--- a/github/template.go
+++ b/github/template.go
@@ -4,6 +4,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


mislav:
No, sorry. We don't want to support system-wide templates. There is no use case in my mind where one should want to use a single issue/PR template for all projects on their system. The templates should only be per-project.

You can always use `hub pull-request -F /path/to/template -e` to edit a fixed template from file. A global configuration mechanism is in the works, after which you'll be able to configure this by default for every `pull-request` invocation.

sorenmat:
Fair enough ;)

PR #1263, opened 2016-09-11
Add a flag `-s|--stat` to append a diffstat to pull request message.

weakish:
Add a flag `-s|--stat`
to append a diffstat to pull request message.

For MarkDown compatibility, seperate `---` with an empty line.
This avoids `---` being recognized as heading mark.

---

 Gemfile.lock             |  3 ---
 commands/pull_request.go | 15 ++++++++++++++-
 git/git.go               | 16 ++++++++++++++++
 git/git_test.go          | 12 ++++++++++++
 4 files changed, 42 insertions(+), 4 deletions(-)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -52,3 +52,6 @@ DEPENDENCIES
   cucumber (~> 1.3.9)
   ronn
   sinatra
+
+BUNDLED WITH
+   1.11.2
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -52,6 +52,3 @@ DEPENDENCIES
   cucumber (~> 1.3.9)
   ronn
   sinatra
-
-BUNDLED WITH
-   1.11.2
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -15,7 +15,7 @@ import (
 var cmdPullRequest = &Command{
 	Run: pullRequest,
 	Usage: `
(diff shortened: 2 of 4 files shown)


mislav:
Hi, thanks for the nice contribution! The code looks good. However, I don't feel that the feature is useful. Why would people attach diffstat information to the PR, when that information is readily available in the “Files” tab of the pull request in the web interface?

Also, new features available in 2.3.0-pre releases allow people to customize further what goes in their PR messages via command-line scripts. You are able to pass a message via stdin and edit it further in the text editor. For example, this feature could be implemented as:

```sh
diff --no-color -M --stat --summary master...feature |  hub pull-request -F - -e
```

Since I think that this feature wouldn't get much use, and I'd like to keep hub as lightweight as possible, I think it's a better idea that you and other people implement such details in their own environments via wrapper scripts.

weakish:
> Why would people attach diffstat information to the PR,
> when that information is readily available in the “Files” tab
> of the pull request in the web interface?

- The `File` tab does not provides a short summary like diffstat.
- The `File` tab is not available in Email.

> new features available in 2.3.0-pre ... `hub pull-request -F - -e`

Not aware of this before. `-e` is helpful here. 
Will piping `diffstat` to `hub -F - -e` replace the default generated pull-request message?
If so, an additional option like `--append` (append file content to message) is needed.

mislav:
Closing because I don't think this feature is very useful since GitHub PRs should ultimately be viewed on GitHub.com (or a similar client that generates diffs). If you want to script this for yourself or for your tream, you can wrap `pull-request -F <file>`. Thank you for your suggestion!

PR #1265, opened 2016-09-12
New `hub sync` command to update local branches

### LINKED ISSUE #618: GitHub for Mac style sync?
Perhaps we could have a `hub sync` command that performs the equivalent of pressing the 'Sync' button in the GitHub for Mac app? I can implement this once I know exactly what the Sync button does. According to the [GitHub for Mac FAQ](https://help.github.com/articles/how-can-i-push-or-pull):

> We do the equivalent of a git pull --rebase (but make sure to never rewrite merges).

Maybe `hub sync` can simply do something like `git pull --rebase=preserve`. Or we could wrap `git pull` and add a `--sync` option that does the same thing.

Does anyone else think this would be a nice feature to have? Any ideas/thoughts?

### LINKED ISSUE #906: Add command to clean all merged branches
I realize this isn't specific to github, but it may be nice to get the feature in here sooner than git.

It would be nice to have some sort of `git flush` type command that 1: performs `git fetch -p` and 2: `git branch --merged | grep -v "\*" | xargs -n 1 git branch -d` to clean all local merged branches.

It would also be nice if it logged an informational message in the console that linked you to github's branches page so we can review and remove all merged or closed branches.

mislav:
Ported from [my bash version](https://github.com/mislav/dotfiles/blob/370154a5adc482afd89606c126bcd18df2560013/bin/git-sync).

Here was it in action: https://twitter.com/mislav/statuses/677091160482054144

Unlike GitHub Desktop, the sync command **doesn't ever push any branches**. It only deals with incoming changes from upstream. Should it be pushing branches?

Fixes #906, fixes #618

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ HELP_CMD = \
 	share/man/man1/hub-pull-request.1 \
 	share/man/man1/hub-release.1 \
 	share/man/man1/hub-issue.1 \
+	share/man/man1/hub-sync.1 \
 
 HELP_EXT = \
 	share/man/man1/hub-am.1 \
--- a/commands/runner.go
+++ b/commands/runner.go
@@ -101,7 +101,10 @@ func (r *Runner) Execute() ExecError {
 		return execErr
 	}
 
-	err = git.Run(args.Command, args.Params...)
+	gitArgs := []string{args.Command}
+	gitArgs = append(gitArgs, args.Params...)
+
+	err = git.Run(gitArgs...)
 	return newExecError(err)
 }
 
--- a/commands/sync.go
+++ b/commands/sync.go
@@ -1,0 +1,132 @@
+package commands
+
+import (
+	"fmt"
+	"os"
+	"regexp"
+	"strings"
+
+	"github.com/github/hub/git"
+	"github.com/github/hub/github"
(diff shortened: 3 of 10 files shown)


prologic:
I still don't see `git sync` / `hub sync` (_I have `alias git=hub`_) deleting local branches that have been merged into origin/master or in my case upstream/master

What am I missing? I'm working with the "forking model" where `upstream` is the upstream repro and `origin` is my fork.

mislav:
@prologic `hub sync` doesn't yet do a great job of cleaning up merged branches in the triangular model (i.e. fork vs. upstream). It mostly handles that in the single-repo workflow.

But just to check: did you delete your branches on `origin` after they were merged into `upstream/master`? `hub sync` will only delete a merged branch if it finds that its push target (in your case, `origin/mybranch`) was deleted.

prologic:
> But just to check: did you delete your branches on `origin` after they were merged into `upstream/master`? `hub sync` will only delete a merged branch if it finds that its push target (in your case, `origin/mybranch`) was deleted.

Looks like I wasn't doing this by looking at the Github UI. But then again I'm not sure Github is able to do this automatically either? Anyway I enabled the option in the UI and deleted old branches. Let's see... :)

Issue #123, opened 2011-12-22
Enterprise: Creating Pull Request shows incorrect info


HighwayofLife:
Creating a Fork of a repo then creating a pull request using hub produces some interesting results...
Steps to reproduce:

```
$ GITHUB_HOST=github.enterprise.com git clone dalewis/repo
Cloning into 'repo' ...
$ cd repo; git checkout -b feature
//... do work
$ git commit -am "example message";git push origin feature
...
To git@github.enterprise.com:dalewis/repo.git
 * [new branch]      feature -> feature
$ git pull-request
```

The template that appears shows the following...

```
# Requesting a pull to dalewis:master from highwayoflife:feature
#
# ...
```

However, "highwayoflife" is my username on github, it's not used in our Enterprise domain. It should be requesting a pull into the source repo (in this example: "org"):

```
# Requesting a pull to org:master from dalewis:feature
#
# ...
```

Submitting the pull request produces an error...

```
Error creating pull request:  (HTTP 401)
Check your token configuration (`git config github.token`)
```

This is likely because it's trying to create a pull request across domains (github.enterprise -> github)

PR #1266, opened 2016-09-12
Reuse existing git remote for PR head in `checkout`

### LINKED ISSUE #1238: Setup tracking branch when checking out PR
I've just tried `hub checkout https://github.com/pytest-dev/pytest-django/pull/380` with hub from master (2.3.0-pre3), but apart from what is mentioned in the man page, it does not seem to setup a tracking branch (according to `git rev-parse --verify @{u} --abbrev-ref`)?

Does this maybe only happen when the remote exists already?

The man page (hub-checkout) seems to refer to the old behavior also still?!

```
EXAMPLES
       $ hub checkout https://github.com/jingweno/gh/pull/73
       > git remote add -f --no-tags -t feature git://github:com/jingweno/gh.git
       > git checkout --track -B jingweno-feature jingweno/feature
```

mislav:
When there is already a git remote for the head repo of a PR to be checked out, fetch the corresponding branch from that remote instead of fetching the special `refs/pull/XY/head` ref.

This sets up a pull/push workflow for branches that the user might have write access to.

Fixes #1238

/cc @blueyed

### DIFF
--- a/commands/checkout.go
+++ b/commands/checkout.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"regexp"
 
+	"github.com/github/hub/git"
 	"github.com/github/hub/github"
 	"github.com/github/hub/utils"
 )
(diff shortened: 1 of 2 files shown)


mislav:
@rcorre This is yet another argument in favor of implicit branch names vs. explicitly given on command-line re: https://github.com/github/hub/pull/1250. Let's say that I'm checking out PR `#123`, which happened to be authored by you, and I already have this remote set up:

```
rcorre git@github.com:rcorre/hub.git
```

What will now happen is the following:

```sh
$ hub checkout <PR-URL>
# ...
# git fetch rcorre +refs/heads/some-fixes:refs/remotes/rcorre/some-fixes
# git checkout -b some-fixes --track rcorre/some-fixes
```

The upstream tracking configuration allows me to git push back to `rcorre/some-fixes` if I have write access to your fork. This approach also works great for same-repo pull requests.

If `checkout` forced me to specify an explicit branch name in these cases, this workflow would be made a little more complicated for the user.

rcorre:
For implicit naming, we have:

- easier for the case where I want to push back to the PR (`checkout URL`)
- harder for the case where I just want to look at the PR (`checkout URL --no-branch`)
- more code
- less familiar to git users (where did this branch come from? Why can't I use `-b`?)

In favor of explicit branch naming, we have:

- harder for the case where I want to push back to the PR (`checkout URL -b mybranch`)
- easier for the case where I just want to look at the PR (`checkout URL`)
- less code
- closer to behavior of 'plain git' (`-b` works as expected)

For me, 'just looking' at a PR is much more common than wanting to push back to it, so explicit naming makes more sense.
That's just my workflow though.

One other point is that for 'explicit naming', `checkout -b` will only set up tracking if `autosetupmerge=true` or they provide `--track` explicitly.
I think this is better for power users who expect hub to respect their git settings, but harder for beginners who might not know such an option exists.

mislav:
@rcorre I've merged this tweak for now, but let's continue the interesting discussion in your PR.

PR #1277, opened 2016-09-21
Use --cmd instead of -c to set default filetype for vim

JelteF:
You are overriding the filetype set by the user using the -c option. This makes
it hard to set a filetype in the vimrc.

### DIFF
--- a/github/editor.go
+++ b/github/editor.go
@@ -113,7 +113,7 @@ func openTextEditor(program, file string) error {
 	editCmd := cmd.New(program)
 	r := regexp.MustCompile("[mg]?vi[m]$")
 	if r.MatchString(program) {
-		editCmd.WithArg("-c")
+		editCmd.WithArg("--cmd")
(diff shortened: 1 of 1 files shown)


mislav:
We set the filetype because we want files named `PULLREQ_EDITMSG` to get the same treatment as git commit messages.

How does setting the filetype hurt your workflow?

JelteF:
The problem was that you could not override the setting with a vimrc anymore, I changed my pull request and now it keeps the filetype setting but it allows overriding because of the --cmd option

This is from vim --help:
```
   --cmd <command>	Execute <command> before loading any vimrc file
   -c <command>		Execute <command> after loading the first file
```

JelteF:
Is there a reason why this is not merged? I changed te pull request to use another method to allow overriding the filetype, which keeps the default set by you.

mislav:
Nice; sorry that I haven't seen your update back then. The workaround looks good. Let me try it myself locally and then I'll merge.

PR #1284, opened 2016-09-27
Changing comment color to gray-ish (matching github's)

hmemcpy:
This makes the commends stand out less in the page, easier to read.

### DIFF
--- a/index.html
+++ b/index.html
@@ -128,7 +128,7 @@
   }
 
   pre .comment {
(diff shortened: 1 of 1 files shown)


mislav:
Screenshot before/after would be appreciated, but this looks okay. Thanks!

hmemcpy:
Bah! Of course, sorry! But thank you for merging!

PR #1285, opened 2016-09-29
Use Markdown filetype in vim for PR messages

### LINKED ISSUE #1276: Add md extensions to PULLREQ_EDITMSG and ISSUE_EDITMSG
This would allow editors to infer the file format correctly.

glittershark:
Use markdown for pull request messages, with modifications to highlight
comments as comments, the first line up to the 50th character as green,
and the second line as red, all pulled from the `gitcommit` filetype
that ships with Vim.

Fixes #1276


mislav:
Interesting; thanks. I'm not opposed to something like this, although it *is* clunky to maintain. Let me try this out myself and see whether it is the direction we want to follow.

PR #1294, opened 2016-10-04
Add a --push flag to hub pull-request.

nex3:
This removes an extra step from the process of publishing a pull
request. It also gives the user a chance to back out of the entire
process by saving an empty pull request message.

### DIFF
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -46,7 +46,7 @@ pull-request -i <ISSUE>
 		Put the URL of the new pull request to clipboard instead of printing it.
 
 	-p, --push
-		Push the local HEAD to --head before creating the pull request.
+		Push the current branch to <HEAD> before creating the pull request.
 
 	-b, --base <BASE>
 		The base branch in "[OWNER:]BRANCH" format. Defaults to the default branch
@@ -207,6 +207,11 @@ func pullRequest(cmd *Command, args *Args) {
 		headTracking = fmt.Sprintf("%s/%s", remote.Name, head)
 	}
 
+
+	if flagPullRequestPush && remote == nil {
+		utils.Check(fmt.Errorf("Can't find remote for %s", head))
+	}
+
 	if cmd.FlagPassed("message") {
 		title, body = readMsg(flagPullRequestMessage)
 	} else if cmd.FlagPassed("file") {
@@ -233,14 +238,11 @@ func pullRequest(cmd *Command, args *Args) {
 	}
 
 	if flagPullRequestPush {
-		if remote == nil {
-			utils.Check(fmt.Errorf("Can't find remote for %s", head))
-		}
-
 		if args.Noop {
 			args.Before(fmt.Sprintf("Would push to %s/%s", remote.Name, head), "")
 		} else {
-			git.Push(remote.Name, "HEAD", head)
+			err = git.Spawn("push", remote.Name, fmt.Sprintf("HEAD:%s", head))
+			utils.Check(err)
 		}
 	}
 
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -45,6 +45,9 @@ pull-request -i <ISSUE>
 	-c, --copy
 		Put the URL of the new pull request to clipboard instead of printing it.
 
+	-p, --push
+		Push the local HEAD to --head before creating the pull request.
+
 	-b, --base <BASE>
 		The base branch in "[OWNER:]BRANCH" format. Defaults to the default branch
 		(usually "master").
@@ -77,6 +80,7 @@ var (
 	flagPullRequestBrowse,
 	flagPullRequestCopy,
 	flagPullRequestEdit,
+	flagPullRequestPush,
 	flagPullRequestForce bool
 
 	flagPullRequestMilestone uint64
@@ -93,6 +97,7 @@ func init() {
 	cmdPullRequest.Flag.BoolVarP(&flagPullRequestCopy, "copy", "c", false, "COPY")
 	cmdPullRequest.Flag.StringVarP(&flagPullRequestMessage, "message", "m", "", "MESSAGE")
 	cmdPullRequest.Flag.BoolVarP(&flagPullRequestEdit, "edit", "e", false, "EDIT")
+	cmdPullRequest.Flag.BoolVarP(&flagPullRequestPush, "push", "p", false, "PUSH")
 	cmdPullRequest.Flag.BoolVarP(&flagPullRequestForce, "force", "f", false, "FORCE")
 	cmdPullRequest.Flag.StringVarP(&flagPullRequestFile, "file", "F", "", "FILE")
 	cmdPullRequest.Flag.VarP(&flagPullRequestAssignees, "assign", "a", "USERS")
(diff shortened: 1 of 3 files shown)


features/pull_request.feature
```diff
@@ -812,3 +812,36 @@ BODY
...
       """
+
+  Scenario: Default message with --push
+    Given the text editor adds:
+      """
+      """
+    Given the GitHub API server:
+      """
+      post('/repos/mislav/coral/pulls') {
+        halt 400 if request.content_charset != 'utf-8'
```
mislav: You don't need this check. It's been put in another scenario to verify what the API client sends, but it doesn't need to be repeated more than once.

features/pull_request.feature
```diff
@@ -812,3 +812,36 @@ BODY
...
+        assert :title => 'This is somewhat of a longish title that does not get wrapped & references #1234',
+               :body => nil
+        status 201
+        json :html_url => "the://url"
+      }
+      """
+    Given I am on the "master" branch pushed to "origin/master"
+    When I successfully run `git checkout --quiet -b topic`
+    Given I make a commit with message "This is somewhat of a longish title that does not get wrapped & references #1234"
```
mislav: Your scenario is not meant to assert that hub successfully uses the message of a single commit for the pull request title & body. Other scenarios are taking care of that already. You can just make an anonymous commit with `Given I make a commit`.

features/pull_request.feature
```diff
@@ -812,3 +812,36 @@ BODY
...
+      post('/repos/mislav/coral/pulls') {
+        halt 400 if request.content_charset != 'utf-8'
+        assert :title => 'This is somewhat of a longish title that does not get wrapped & references #1234',
+               :body => nil
+        status 201
+        json :html_url => "the://url"
+      }
+      """
+    Given I am on the "master" branch pushed to "origin/master"
+    When I successfully run `git checkout --quiet -b topic`
```
mislav: Would it work if you just  said `Given I am on the "topic" branch` on the previous line instead of starting with the "master" branch?

features/pull_request.feature
```diff
@@ -812,3 +812,36 @@ BODY
...
+    When I successfully run `hub pull-request -p`
+    Then the output should contain exactly "the://url\n"
+    And "git push origin HEAD:topic" should be run
+
+  Scenario: Text editor fails with --push
+    Given the text editor exits with error status
+    And I am on the "master" branch pushed to "origin/master"
+    And an empty file named ".git/PULLREQ_EDITMSG"
+    When I successfully run `git checkout --quiet -b topic`
+    Given I make a commit with message "Title"
```
mislav: Ditto, you don't need the `with message …` part.

git/git.go
```diff
@@ -223,6 +223,10 @@ func Log(sha1, sha2 string) (string, error) {
 	return outputs, nil
 }
 
+func Push(remote, source, dest string) error {
+	return Spawn("push", remote, fmt.Sprintf("%s:%s", source, dest))
```
mislav: This is only being used in a single place in the codebase. I don't  think it warrants its own method. I would just inline the `git.Spawn` invocation in `pull-request.go`

commands/pull_request.go
```diff
@@ -222,6 +232,18 @@ func pullRequest(cmd *Command, args *Args) {
...
 
+	if flagPullRequestPush {
+		if remote == nil {
+			utils.Check(fmt.Errorf("Can't find remote for %s", head))
+		}
+
+		if args.Noop {
+			args.Before(fmt.Sprintf("Would push to %s/%s", remote.Name, head), "")
+		} else {
+			git.Push(remote.Name, "HEAD", head)
```
mislav: There's something we need to take care of after pushing the branch. If we proceeded to make an API request immediately to try to create a pull request for a branch that was just pushed, the API might return an error that said `head` branch doesn't exist. This is because there is a little bit of delay between pushing a branch and GitHub registering that this branch exist right now. This is because GitHub processes git pushes via background jobs.

The tricky part is determining how long does this delay need to be. Maybe start with 5 seconds? In my experience, several seconds is usually all that's needed for GitHub to process it.

A more robust approach would be to try the API POST request, and if it's a 422 with `head` mentioned in validation errors, retry the POST automatically after several seconds, then again, and ultimately fail after a few tries. This might be tricky. Do you think it would be worth it?

commands/pull_request.go
```diff
@@ -45,6 +45,9 @@ pull-request -i <ISSUE>
 	-c, --copy
 		Put the URL of the new pull request to clipboard instead of printing it.
 
+	-p, --push
+		Push the local HEAD to --head before creating the pull request.
```
nex3: Done.

commands/pull_request.go
```diff
@@ -188,27 +193,32 @@ func pullRequest(cmd *Command, args *Args) {
...
-			baseTracking = fmt.Sprintf("%s/%s", remote.Name, base)
-		}
-		if remote == nil || !baseProject.SameAs(headProject) {
-			remote = gitRemoteForProject(headProject)
-		}
-		if remote != nil {
-			headTracking = fmt.Sprintf("%s/%s", remote.Name, head)
+		headForMessage := headTracking
+		if flagPullRequestPush {
+			headForMessage = head
```
nex3: The message is generated before the branch is pushed, so that the user has a chance to back out of pushing by passing an empty message. This means that you want to generate the message based on the local head, since the remote head doesn't have all the relevant commits yet.

commands/pull_request.go
```diff
@@ -222,6 +232,18 @@ func pullRequest(cmd *Command, args *Args) {
 		utils.Check(fmt.Errorf("Aborting due to empty pull request title"))
 	}
 
+	if flagPullRequestPush {
+		if remote == nil {
+			utils.Check(fmt.Errorf("Can't find remote for %s", head))
```
nex3: Done.

commands/pull_request.go
nex3: I've been using this utility personally and in practice I haven't had this race condition bite me, so it would be a shame to delay all requests by multiple seconds when only a few would benefit. I think it's probably worth implementing the retry logic, but I have a couple questions:

* What would the 422 response's body look like?
* How should I structure the retry logic? It looks like only the `Client.CreatePullRequest()` method has access to the status code, but it seems wrong to always retry in that method, since any other call won't be susceptible to this race condition.

features/pull_request.feature
```diff
@@ -812,3 +812,36 @@ BODY
       Error creating pull request: Temporary Redirect (HTTP 307)
       Refused to follow redirect to https://disney.com/mouse\n
       """
+
+  Scenario: Default message with --push
+    Given the text editor adds:
```
nex3: Done. I do want to run the editor here to ensure that the automatically-generated message is correct.

features/pull_request.feature
nex3: Done.

features/pull_request.feature
nex3: I think I need the previous line to push to `origin/master`, since the message is generated from commits based on `origin/master`.

features/pull_request.feature
nex3: I think validating the message is important here, since the logic is different with `--push` than otherwise. In particular, `--push` should construct the message based on changes from `origin/master` to the local `head`, whereas without it the message is based on changes from `origin/master` to the remote `head`.

features/pull_request.feature
nex3: Done.

git/git.go
nex3: Done.

nex3:
I think retrying the POST is the way to go, but I'm not sure what the right way to structure that code would be.

mislav:
Thanks! Added the retry logic in 2437c48e657649b98f6812669522b8e54d988f98

Issue #1288, opened 2016-09-29
Update in Chocolatey request


Kristinita:
Now via Chocolatey I can download [only 2.2.1 version](https://chocolatey.org/packages/github-hub).

I [wrote issue](https://github.com/twsouthwick/github-hub-chocolatey/issues/1) to maintainer hub on Chocolatey twsouthwick, but he not answers to me [and you](https://github.com/github/hub/issues/794#issuecomment-246472692). Please, assign new maintainer, because hub not upgrade in Chocolatey many times.

Thanks.

dougludlow:
To summarize what I saw in the linked issues, it appears that both the [github-hub](https://chocolatey.org/packages/github-hub) (v2.2.1) and [hub](https://chocolatey.org/packages/hub) (v2.2.5) chocolately packages have both gone stale. Attempts have been made to contact the maintainers (@twsouthwick, @dtgm), but without avail. 

Chocolatey apparently has a [triage process](https://chocolatey.org/docs/package-triage-process#package-is-outdated) for what to do when a package is outdated. They even have a process when a software vendor [wishes to take over the package maintenance](https://chocolatey.org/docs/package-triage-process#i-want-to-take-overhelp-with-package-maintenance-for-my-software).

I'm more than happy to take steps to transfer the maintenance of one of the packages to myself, unless anyone else would like it. The other option would be for the maintainers for github/hub to claim the package and take it over themselves.

twsouthwick:
Hi - just saw this. I've responded to the issue here (https://github.com/twsouthwick/github-hub-chocolatey/issues/1). As stated there, I'm no longer using the tool so I haven't been tracking chocolatey updates. I'd be happy to transfer ownership (Chocolatey as well as the GitHub project for the chocolatey wrapper) to whoever would like to take it

Kristinita:
Now we can download [latest 2.2.9 version](https://chocolatey.org/packages/hub) in Chocolatey. @dougludlow , @mislav , if you all wants maintainers of package together @dtgm, it would be nice) .

Thanks.

mislav:
@twsouthwick Could you transfer the ownership of the package to me and outline the process of upgrading the package version for me so I can hopefully automate it?

twsouthwick:
Sure! It'll have to be tomorrow as I'm out of town right now. I look into the process and let you know.

mislav:
Thanks @twsouthwick!

twsouthwick:
[...]

PR #1299, opened 2016-10-06
Add chocolatey and powershell instructions.

dougludlow:


mislav:
Hi thanks for your contibution! The docs look really solid. But I have two issues with the content. First of all, Chocolatey is nice and all, but our releases there are really out of date and I don't recommend installing them because of known bugs. I don't know how to upgrade those packages. So before we can add this documentation, we need to figure that out. See #1288

Second, those Powershell instructions look really complicated, with separate steps if your config file already exists or not. Is it possible to make a single command that works regardless of whether a config file is already there?

dougludlow:
Hi @mislav, sure I'll take a look at #1288. I've been using the chocolately version just fine for a while, but I understand.

Regarding a single command, I played around with one. I'm no powershell expert, but in powershell it's a very complicated/long one liner, something like this:

``` sh
Write-Output "`nSet-Alias git hub" | %{If (Test-Path $PROFILE) { Out-File -Encoding utf8 -Append -InputObject $_ -Filepath $PROFILE } Else { New-Item -Force -Path $PROFILE -Value $_ -Type file }}
```

If you're fine with that I can put it in instead. I thought it might be simpler to break it up into multiple commands.

mislav:
You're right, it does look simpler if we break it up. Let's go for the
shortest possible instructions. We don't want to have people copy/pasting
huge chunks of code in their terminal that they don't understand.

dougludlow:
@mislav I can't remember where we landed on this. Was there anything I needed to change here?

mislav:
Thanks for the patience!

PR #1306, opened 2016-10-14
Better error message when remote origin is not GitHub

raboof:
Previously the actual error was discarded, though that might
contain useful information.

In my case, for example, the actual origin URL that didn't match was not the
one I expected.

### DIFF
--- a/features/ci_status.feature
+++ b/features/ci_status.feature
@@ -85,7 +85,7 @@ Feature: hub ci-status
   Scenario: Non-GitHub repo
     Given the "origin" remote has url "mygh:Manganeez/repo.git"
     When I run `hub ci-status`
-    Then the stderr should contain "Aborted: the origin remote doesn't point to a GitHub repository.\n"
+    Then the stderr should contain "Aborted: the origin remote doesn't point to a GitHub repository: Invalid GitHub URL: ssh://mygh/Manganeez/repo.git.\n"
     And the exit status should be 1
 
   Scenario: Enterprise CI statuses
--- a/github/localrepo.go
+++ b/github/localrepo.go
@@ -191,14 +191,14 @@ func (r *GitHubRepo) MainRemote() (remote *Remote, err error) {
 func (r *GitHubRepo) MainProject() (project *Project, err error) {
 	origin, err := r.MainRemote()
 	if err != nil {
-		err = fmt.Errorf("Aborted: the origin remote doesn't point to a GitHub repository.")
+		err = fmt.Errorf("Aborted: the origin remote doesn't point to a GitHub repository: %v.", err)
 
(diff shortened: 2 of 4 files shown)


github/localrepo.go
```diff
@@ -191,14 +191,14 @@ func (r *GitHubRepo) MainRemote() (remote *Remote, err error) {
 func (r *GitHubRepo) MainProject() (project *Project, err error) {
 	origin, err := r.MainRemote()
 	if err != nil {
-		err = fmt.Errorf("Aborted: the origin remote doesn't point to a GitHub repository.")
+		err = fmt.Errorf("Aborted: the origin remote doesn't point to a GitHub repository: %v.", err)
```
mislav: Since this error message will get displayed on a very long line, and doesn't feel like a single sentence anymore, how about breaking it up into two lines?

```
Aborted: the origin remote doesn't point to a GitHub repository.
Invalid GitHub URL: ssh://mygh/Manganeez/repo.git
```

raboof: Good one! I've indented the second line a bit to make it move obvious that it's the continuation of the error message on the previous line, not a subsequent error.

github/localrepo.go
```diff
@@ -191,14 +191,14 @@ func (r *GitHubRepo) MainRemote() (remote *Remote, err error) {
 func (r *GitHubRepo) MainProject() (project *Project, err error) {
 	origin, err := r.MainRemote()
 	if err != nil {
-		err = fmt.Errorf("Aborted: the origin remote doesn't point to a GitHub repository.")
+		err = fmt.Errorf("Aborted: the origin remote doesn't point to a GitHub repository:\n %v.", err)
```
mislav: There's no need to indent the following line, since those two lines will likely be the only lines of stderr output from the program, which will abort at this point. There will be no confusion that the 2nd line refers to the 1st. Also, the first line should end in a period so that it forms a sentence and doesn't have the colon twice.

features/ci_status.feature
```diff
@@ -85,7 +85,7 @@ Feature: hub ci-status
   Scenario: Non-GitHub repo
     Given the "origin" remote has url "mygh:Manganeez/repo.git"
     When I run `hub ci-status`
-    Then the stderr should contain "Aborted: the origin remote doesn't point to a GitHub repository.\n"
+    Then the stderr should contain "Aborted: the origin remote doesn't point to a GitHub repository:\n Invalid GitHub URL: ssh://mygh/Manganeez/repo.git.\n"
```
mislav: Please use the multine syntax for this step:

```
Then the stderr should contain:
  """
  line
  another line\n
  """
```

It's easier to read.

PR #1312, opened 2016-10-25
Add command to manage a branching workflow

boris-rea:
I add a branching model like GitFlow.
You can use it for GitFlow and GitHub Flow.
it uses config like GitFlow in .git/config .


mislav:
Hi, thank you for your absolutely huge contribution. I can see that you put a lot of work in it. However, I don't think an implementation of git-flow is a good fit for hub. From our own CONTRIBUTING documentation:

> * A feature that encapsulates a git workflow *not specific* to GitHub is **not**
  a good fit for hub, since something like that is best implemented as an
  external script.

git-flow already has a good implementation, and it's called git-flow. It should be installed separately from hub, but it can be used at the same time. There's no need to replicate the functionality of git-flow in here.

Also, the diffstat of this pull request is a little bit scary: `+13,372 -2`. I generally avoid pulling contributions that have enourmous diffs that are hard to review. The size of this diff seems mostly affected by the fact that, for some reason, you decided to check in a whole nested copy of the `github/hub` project within the vendor directory.

PR #1318, opened 2016-10-28
Switch default remote URL to https

matschaffer:
Github doesn't seem to honor `git://` anymore.

### DIFF
--- a/github/project.go
+++ b/github/project.go
@@ -72,12 +72,12 @@ func (p *Project) GitURL(name, owner string, isSSH bool) (url string) {
 
 	host := rawHost(p.Host)
 
-	if preferredProtocol() == "https" {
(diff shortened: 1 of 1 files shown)


mislav:
> Github doesn't seem to honor `git://` anymore

What do you mean by that?

I appreciate the suggestion, and I was meaning to switch to HTTPS as default in the next release. However, your change is pretty substantial, and has broken many tests in the suite. Do you plan you update those as well?

Also, the `HTTPS is preferred` tests need to be renamed to "git protocol is preferred" and changed to test the non-default option which is `git`.

2m:
I have bumped into the same issue where remotes have `git://` as a scheme when checking out a PR using hub.

I had to change the start of the remote URL to `git@github.com:...` manually in order to be able to push additional commits to the PR branch.

PR #1324, opened 2016-10-30
Check if config location is writeable before authenticating

### LINKED ISSUE #1314: Check if config location is writeable before authenticating
### 1. Clone this repo

```
$ git clone https://github.com/github/hub
Cloning into 'hub'...
remote: Counting objects: 13524, done.
remote: Compressing objects: 100% (35/35), done.
remote: Total 13524 (delta 20), reused 0 (delta 0), pack-reused 13489
Receiving objects: 100% (13524/13524), 3.57 MiB | 1.17 MiB/s, done.
Resolving deltas: 100% (8530/8530), done.
```
### 2. Make a fork

```
$ cd hub
$ hub fork
github.com username: <user.name>
github.com password for <user.name> (never stored):
Error forking repository: Unprocessable Entity (HTTP 422)
Duplicate value for "description"
```

akshatgoel:
Fixes #1314

### DIFF
--- a/features/authentication.feature
+++ b/features/authentication.feature
@@ -95,7 +95,7 @@ Feature: OAuth authentication
       Duplicate value for "description"
       """
     And the exit status should be 1
-    And the file "../home/.config/hub" should contain ""
+    And the file "../home/.config/hub" should not exist
 
   Scenario: Credentials from GITHUB_USER & GITHUB_PASSWORD
     Given the GitHub API server:
@@ -172,7 +172,7 @@ Feature: OAuth authentication
 
       """
     And the exit status should be 1
-    And the file "../home/.config/hub" should contain ""
+    And the file "../home/.config/hub" should not exist
 
   Scenario: Personal access token used instead of password
     Given the GitHub API server:
@@ -192,7 +192,7 @@ Feature: OAuth authentication
 
       """
     And the exit status should be 1
-    And the file "../home/.config/hub" should contain ""
+    And the file "../home/.config/hub" should not exist
 
   Scenario: Two-factor authentication, create authorization
     Given the GitHub API server:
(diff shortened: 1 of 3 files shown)


features/authentication.feature
```diff
@@ -332,3 +332,15 @@ Feature: OAuth authentication
     Then the output should contain "github.com username:"
     And the output should contain "missing user"
     And the file "../home/.config/hub" should not contain "user"
+    
+  Scenario: Config file is not writeable, should exit before asking for credentails
+      Given I set the environment variables to:
```
mislav: Nice job discovering this step built into Aruba. However, since we're only testing one value, I think a Cucumber table is an overkill. There's a simpler step in our `steps.rb` in the format of:

```
Given $FOO is "bar"
```

features/authentication.feature
```diff
@@ -95,7 +95,7 @@ Feature: OAuth authentication
       Duplicate value for "description"
       """
     And the exit status should be 1
-    And the file "../home/.config/hub" should not exist
+    And the file "../home/.config/hub" should contain ""
```
mislav: Ah, nice catch. These tests did expect no file being created, but we create one anyway to check for writeability.

How about a workaround: `CheckWriteable()` would first test if the file is present before trying to write it. If it wasn't present, remember that. Then, after checking for writeability, and if the file wasn't present previously, delete it. How does that sound?

features/authentication.feature
```diff
@@ -332,3 +332,15 @@ Feature: OAuth authentication
     Then the output should contain "github.com username:"
     And the output should contain "missing user"
     And the file "../home/.config/hub" should not contain "user"
+    
+  Scenario: Config file is not writeable, should exit before asking for credentails
+      Given I set the environment variables to:
+      | variable           | value      |
+      | HUB_CONFIG | /InvalidConfigFile |
```
mislav: In addition to testing a custom location via HUB_CONFIG environment variable, can you make a scenario where hub tries to write the config in the default location, but that location isn't writeable or an existing file is read only?

We could maybe use these built-in Aruba steps:

```rb
Given /^a directory named "([^"]*)" with mode "([^"]*)"$/ do |dir_name, dir_mode|
  create_dir(dir_name)
  chmod(dir_mode, dir_name)
end

Given /^a file named "([^"]*)" with mode "([^"]*)" and with:$/ do |file_name, file_mode, file_content|
  write_file(file_name, file_content)
  chmod(file_mode, file_name)
end
```

The default config location will be `$HOME/.config/hub`, where "HOME" in the context of tests is [defined here](https://github.com/github/hub/blob/4a7f8cb53a4ef1364e348d45b45628430c077eba/features/support/env.rb#L38).

github/config.go
```diff
@@ -51,6 +51,7 @@ func (c *Config) PromptForHost(host string) (h *Host, err error) {
 	h = c.Find(host)
 	if h != nil {
 		if h.User == "" {
+			utils.Check(newConfigService().CheckWriteable(configsFile()))
```
mislav: There's very little need that `CheckWriteable` is a method on `configService`. How about we make it a standalone function so we don't have to instantiate a dummy `configService` object to access the method?

features/authentication.feature
akshatgoel: I agree. Updated. Also updated new tests to check that file is not present after the test.

features/authentication.feature
akshatgoel: Thanks. Missed that as $ was hidden in regex.

github/config.go
akshatgoel: Agreed. Updated.

mislav:
Thanks, everything looks great! 👍

PR #1327, opened 2016-10-31
Add ability to delete releases

### LINKED ISSUE #1326: Add ability to delete a release
I have a workflow where I automatically create a new nightly release every night in order to easily distribute tarballs/rpms to internal testers. I don't really want to keep all the old releases around on the release page (I do keep the tags, but they don't clutter things up as much).

While I can create releases all day with the `hub release` command, I can't delete old ones. Can we get this added?

wesbland:
To mirror all the other `hub release` commands, allow the user to delete
releases from the command line. The new subcommand takes only one
argument (the tag of the release being deleted).

Fixes #1326

### DIFF
--- a/commands/release.go
+++ b/commands/release.go
@@ -21,6 +21,7 @@ release [--include-drafts]
 release show <TAG>
 release create [-dpoc] [-a <FILE>] [-m <MESSAGE>|-F <FILE>] [-t <TARGET>] <TAG>
 release edit [<options>] <TAG>
+release delete <TAG>
 `,
 		Long: `Manage GitHub releases.
 
@@ -49,6 +50,9 @@ With '--include-drafts', include draft releases in the listing.
 
 	* _download_:
 	  Download the assets attached to release for the specified <TAG>.
+
+	* _delete_:
+	  Delete the release and associated assets for the specified <TAG>.
 
 ## Options:
 	-d, --draft
@@ -111,6 +115,11 @@ hub(1), git-tag(1)
 		Run: downloadRelease,
 	}
 
+	cmdDeleteRelase = &Command{
+		Key: "delete",
+		Run: deleteRelease,
+	}
+
 	flagReleaseIncludeDrafts,
 	flagReleaseShowDownloads,
 	flagReleaseDraft,
@@ -153,6 +162,7 @@ func init() {
 	cmdRelease.Use(cmdCreateRelease)
 	cmdRelease.Use(cmdEditRelease)
 	cmdRelease.Use(cmdDownloadRelease)
+	cmdRelease.Use(cmdDeleteRelase)
 	CmdRunner.Use(cmdRelease)
 }
 
@@ -421,6 +431,35 @@ func editRelease(cmd *Command, args *Args) {
 	}
 
 	uploadAssets(gh, release, flagReleaseAssets, args)
+	args.NoForward()
+}
+
+func deleteRelease(cmd *Command, args *Args) {
+	tagName := cmd.Arg(0)
+	if tagName == "" {
+		utils.Check(fmt.Errorf("Missing argument TAG"))
+		return
+	}
+
+	localRepo, err := github.LocalRepo()
+	utils.Check(err)
+
+	project, err := localRepo.CurrentProject()
+	utils.Check(err)
+
+	gh := github.NewClient(project.Host)
+
+	release, err := gh.FetchRelease(project, tagName)
+	utils.Check(err)
+
+	if args.Noop {
+		message := fmt.Sprintf("Deleting release related to %s...", tagName)
+		ui.Println(message)
+	} else {
+		err = gh.DeleteRelease(release)
+		utils.Check(err)
+	}
+
 	args.NoForward()
 }
 
--- a/features/release.feature
+++ b/features/release.feature
@@ -367,6 +367,24 @@ MARKDOWN
           ASSET_TARBALL
           """
 
+  Scenario: Delete a release
+    Given the GitHub API server:
+      """
+      get('/repos/mislav/will_paginate/releases') {
+          json [
+            { url: 'https://api.github.com/repos/mislav/will_paginate/releases/123',
+              tag_name: 'v1.2.0',
+            },
+          ]
+      }
+
+      delete('/repos/mislav/will_paginate/releases/123') {
+        status 204
+      }
+      """
+    When I successfully run `hub release delete v1.2.0`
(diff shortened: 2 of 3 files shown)


commands/release.go
```diff
@@ -424,6 +433,37 @@ func editRelease(cmd *Command, args *Args) {
...
+	project, err := localRepo.CurrentProject()
+	utils.Check(err)
+
+	gh := github.NewClient(project.Host)
+
+	release, err := gh.FetchRelease(project, tagName)
+	utils.Check(err)
+
+	/*cs := git.CommentChar()
+	message, err := renderReleaseTpl("Deleting", cs, tagName, project.String(), flagReleaseCommitish)
```
mislav: Copy/pasta slipped in? 😉

commands/release.go
```diff
@@ -424,6 +433,37 @@ func editRelease(cmd *Command, args *Args) {
...
+	utils.Check(err)
+	message = fmt.Sprintf("%s\n\n%s\n%s", release.Name, release.Body, message)*/
+	message := fmt.Sprintf("Deleting release related to %s...", tagName)
```
mislav: A command shouldn't output what it's doing, only if there's something useful as a result (such as the URL of a freshly created release). This command doesn't have anything useful to report, so it shouldn't print anything.

However, hub commands can be run in `--noop` mode, where they simply print what would be done, but don't actually make the changes. In this mode, it's fine to read from the API, but not to write to it. Instead, it should output something like `Would delete release for tag %s`.

features/release.feature
```diff
@@ -367,6 +367,36 @@ MARKDOWN
...
+      get('/repos/mislav/will_paginate/releases') {
+          json [
+            { url: 'https://api.github.com/repos/mislav/will_paginate/releases/123',
+              upload_url: 'https://api.github.com/uploads/assets{?name,label}',
+              tag_name: 'v1.2.0',
+              name: 'will_paginate 1.2.0',
+              draft: true,
+              prerelease: false,
+              assets: [
+                { url: 'https://api.github.com/repos/mislav/will_paginate/assets/9876',
```
mislav: It's fine that you copied the full payload from GitHub's API, but typically in tests I reduce the JSON payload to just fields we read. In hub we don't actually read most JSON fields, so it's fine if they're not there at all. It makes the tests read cleaner.

For this payload, it would be fine if we only kept the fields `id`, `url`, and `tag_name`.

features/release.feature
```diff
@@ -367,6 +367,36 @@ MARKDOWN
...
+      }
+
+      delete('/repos/mislav/will_paginate/releases/123') {
+        status 204
+      }
+      """
+    When I successfully run `hub release delete v1.2.0`
+    Then the output should contain exactly:
+      """
+      Deleting release related to v1.2.0...\n
```
mislav: When you get rid of this output, as advised in my previous comment, then it's going to be hard to assert if the command actually did anything.

If you're comfortable with Ruby, you can amend the api server to keep the list of releases in memory while it's running. Then you can issue several requests to it. For example:

```rb
releases = [
  { tag_name: 'v1.2.0',
    id: '...',
    url: '...
  },
  { tag_name: 'v1.2.1',
    id: '...',
    url: '...
  },
]

get('/repos/mislav/will_paginate/releases') {
  json releases
}

delete('/repos/mislav/will_paginate/releases/:id') {
  if releases.reject! { |r| r.id.to_s == params[:id] }
    status 204
  else
    status 404
  end
}
```

Then you can do `hub release delete v1.2.0`, verify that the output is blank, then do `hub release` and verify that "v1.2.0" no longer appears in the output. Or, you could try to run `hub release delete ...` twice and verify that the 2nd run fails. What do you think?

commands/release.go
wesbland: Gah! Yes. I'll get rid of that.

commands/release.go
wesbland: I went back and forth on this. Originally I didn't have any output. Then I thought it might make sense to have some sort of feedback (in the same way that uploading artifacts for the release does). I'm happy to go back to the original version (with `--noop` mode).

features/release.feature
wesbland: Definitely happy to have the feedback here. Implementing the feature took me 10 minutes. Figuring out how to use the dang tester took the rest of the hour. 😄

features/release.feature
wesbland: I like the idea, though I'm obviously not as familiar with the implementation as you. I'm not sure I'd be able to whip up something more complex then the simple check for the status code. I'm usually a C developer, not so much web APIs.

If you really want to see this, I can try to figure it out, but I'd be just as happy if it just showed up. It seems like the `204` return code should be enough to validate that it worked. I'll have to verify this though.

features/release.feature
mislav: Haha, sorry about that! What was the hardest part?

features/release.feature
mislav: Oh sure, you already added code that verifies that 204 was returned. That part is fine. However, if the `hub release delete` command has no output, and exists with a success status, the test still can't verify that it actually did anything, i.e. that it did issue a DELETE request. That's why I suggested a bit of trickery with the test server.

My suggested code above could work, but it hasn't been tested. If you'd like me to improve the test instead, I can definitely take over once you're done with other things.

features/release.feature
wesbland: None of it was that hard. I just wasn't familiar with how things worked. Specifically that I was setting up the JSON object to be returned and the acting on it with the second command. As with all things, once I understood what it was looking for it made more sense.

features/release.feature
wesbland: I think I'm gonna vote for that one. I'll push my updates for the rest of the fixes here in just a second.

wesbland:
I agree that this probably won't get super wide usage, but it seems like good practice that if you're going to have a way to create something, you have a way to delete it. Obviously, that doesn't apply for issues and PRs that *can't* be deleted, but for everything else (branches, tags, etc.) there is at least some way of making it work. Even if it's not recommended.

This will be especially nice for my workflow of automatically updating the nightly release to point to the new tag every night.

mislav:
The updates look good, thank you! I'll try to extend the cucumber test with a bit more Ruby code to ensure that a release was in fact deleted (right now we have no verification that a `DELETE` HTTP request was actually made during the execution of the test), but other than that, this is good to go. 👍

BryanH:
Can this be merged, please?

Issue #491, opened 2014-02-20
Error creating pull request: Unauthorized (HTTP 401)


pwaller:
With apologies to @mislav, I'm reviving #83, #413, #412, #435.

I'm reviving it because this has broken my (and others in my org) ability to use hub at all for months and I wasn't aware of the workarounds available in the above linked issues. I had assumed something had broken in my environment.

It broke for me because as a security precaution I clobbered all of my existing OAUTH tokens, which were originally generated by signing in at the command line. I expected the previous process (enter credentials at the command line) to work, but it doesn't any more, with no explanation as to why.

I did try to generate a ["personal access token"](https://github.com/settings/applications) and enter it when hub prompted me for a password but of course that didn't work. I expected it to work because of the description "They can be used instead of a password for Git over HTTPS, or can be used to authenticate to the API over Basic Authentication". But it doesn't.

The current experience is: `hub pull-request`, enter credentials, enter message, wait a few seconds, etc.

```
Error creating pull request: Unauthorized (HTTP 401)
```

Please close this issue once it is fully resolved such that the user experience at the command line is not broken, otherwise it's hard to teach newbies to use hub. One has to know that having 2FA enabled means you have to generate a token and edit a config file which is not at all apparent from the UI.

viktornordling:
> ```yaml
> github.com:
> - user: pwaller
> - oauth_token: ACCESS_TOKEN
> ```

This led me astray. Should be:

```
 github.com:
 - user: pwaller
   oauth_token: ACCESS_TOKEN
```

Hopefully this helps someone, five years later! 😂

hackaprende:
> @pwaller I'm sorry you had troubles authenticating. Let's start from the beginning: what is the outcome that you get when hub asks you for your GH credentials? It should ask you for username, password (your actual GH password; never stored), and 2FA code if you have that enabled. Does it ask you for all three, and fails after that? Is the error message that you pasted the full error output?
> 
> After you entered your actual GH credentials + 2FA code, even if the `pull-request` operation ultimately failed, can you check did hub manage to create a token for itself? It should be visible under [Personal access tokens](https://github.com/settings/applications), and its name should simply be "hub" with "repo" scope.
> 
> Ultimately, if you're still blocked on this, here is how you can provide auth info to hub manually:
> 
> 1. Create a Personal Access Token (like you did already). This acts as an OAuth token
> 2. Save the following information in `~/.config/hub`:
>    ```yaml
>    github.com:
>    - user: pwaller
[...]

PR #1329, opened 2016-11-03
Add authentication instructions to README

greg-1-anderson:
Authentication options are discussed in #491 and #463; it would be good to have some of this information in the README as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,13 @@ complement existing completion scripts that ship with git.
 * [hub bash completion](https://github.com/github/hub/blob/master/etc/hub.bash_completion.sh)
 * [hub zsh completion](https://github.com/github/hub/blob/master/etc/hub.zsh_completion)
 
+Authentication
+--------------
+
+Many hub commands must authenticate against GitHub to operate. Often, hub will prompt you for your credentials when they are needed.
+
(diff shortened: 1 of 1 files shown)


mislav:
Thanks for the suggestions. But, there are 2 reasons why I won't accept this PR right now:

1. The preferred way to authenticate right now isn't the GITHUB_TOKEN environment variable, since that requires people to generate the Personal Access Token and populate the variable manually. Instead, the preferred way to authenticate is by hub prompting you for username/password when it needs to for the first time. This method needs to documentation, since it happens by itself.

2. The next version of hub is going to change the authentication strategy to have a dedicated command to set up authentication #225. This means it will no longer prompt you interactively, but you will be required to set it up the first time (and additionally for every GitHub Enterprise host). When that ships, that command will have documentation of its own.

greg-1-anderson:
The information on authenticating via `GITHUB_TOKEN` is not easy to find. Since this information is very interesting to folks setting up hub to work in CI environments, and will continue to be supported, is there some other document  where these instructions could be recorded?  https://hub.github.com or https://hub.github.com/hub.1.html, perhaps?

mislav:
We added GITHUB_TOKEN docs to the man page in the 2.2-stable branch, but I
forgot to publish it to the site. It's now published! Thanks for the
reminder.

Issue #1305, opened 2016-10-14
Hub expands alias that git wouldn't


aweibell:
The command

```
hub branch -d my_local_branch
```

results in this error:

> **fatal: cannot use -a** with -d

I consider this a major bug. **I did not use `-a`**, but somehow that is added by hub!

Also, when running `hub branch` I only expect to see my local branches, but all the remote branches are listed as well. Looks like hub adds the `-a` option for all `branch` commands? Users should be allowed to add the `-a` option themselves when they find it applicable.

Plain old `git branch` works as expected in both situations mentioned above.

mislav:
This is strange. Hub shouldn't ever add extra arguments, especially not to `branch`.

1. Which version of hub do you use?
2. Do you have any git aliases related to `branch`?
2. Do you get the same result with `command git branch -d ...`, or does that work properly?
3. What does `HUB_VERBOSE=1 hub branch -d ...` show?

aweibell:
I agree that it is strange ;)

 1. I experienced the issue with both hub-linux-amd64-2.2.3.tgz and hub-linux-amd64-2.3.0-pre8.tgz
 2. I thought I had eliminated all aliases by testing without scm-breeze and oh-my-zsh, but it appears someone (a script I've tested or something, I assume) has set some **git aliases** without my knowledge: :open_mouth:  
> [alias]
>    br = branch
>    branch = branch -ra
>    bra = branch -ra

So that changes the situation, of course. BUT:

 3. That's exactly what made me look at it as a hub bug. The `git` command behaves different from the `hub` command here!

```bash
~> git branch
* master

~> hub branch
* master
  remotes/origin/master
```

 4. Hub seems to evaluate the alias:
```
HUB_VERBOSE=1 hub branch
$ git config alias.branch
$ git branch -ra
* master
  remotes/origin/master

~> HUB_VERBOSE=1 hub bra
$ git config alias.bra
$ git branch -ra
* master
  remotes/origin/master

~> HUB_VERBOSE=1 hub br
$ git config alias.br
$ git branch
* master
```
I did the same exercise with git, and it does not seem to bother about the alias!

```
~> GIT_TRACE=1 git branch
22:07:33.953967 git.c:344               trace: built-in: git 'branch'
* master

~> GIT_TRACE=1 git br
22:07:41.121808 git.c:561               trace: exec: 'git-br'
22:07:41.121840 run-command.c:334       trace: run_command: 'git-br'
[...]

PR #1332, opened 2016-11-05
Do not expand aliases which are in-built git and hub commands

akshatgoel:
Issue #1305

Initial version for the fix. I think it is a bit computation for every command. Will be glad to get any comment for any possible optimization. 

One possible optimization is we check for alias first and if we find it, before expanding, we check if it is in-built command.

```
func expandAlias(args *Args) {
	cmd := args.Command
	expandedCmd, err := git.Alias(cmd)
	// Don't expand built in commands : Issue #1305
	
	if err == nil && expandedCmd != "" && !isBuiltInCommand(cmd) {
		words, e := splitAliasCmd(expandedCmd)
		if e == nil {
			args.Command = words[0]
			args.PrependParams(words[1:]...)
		}
	}
}
```

### DIFF
--- a/commands/runner.go
+++ b/commands/runner.go
@@ -6,6 +6,7 @@ import (
 	"os/exec"
 	"strings"
 	"syscall"
+	"regexp"
 
 	"github.com/github/hub/cmd"
 	"github.com/github/hub/git"
@@ -90,7 +91,11 @@ func (r *Runner) Execute() ExecError {
 	utils.Check(err)
 
 	git.GlobalFlags = args.GlobalFlags // preserve git global flags
-	expandAlias(args)
+
+	// Don't expand built in commands : Issue #1305
+	if !isBuiltInCommand(args.Command) {
+		expandAlias(args)
+	}
 
 	cmd := r.Lookup(args.Command)
 	if cmd != nil && cmd.Runnable() {
@@ -160,6 +165,11 @@ func executeCommands(cmds []*cmd.Cmd, execFinal bool) error {
 	return nil
 }
 
+func isBuiltInCommand(command string) bool {
+	builtInCommands := getListOfBuiltInCommands()
+	return builtInCommands[command]
+}
+
 func expandAlias(args *Args) {
 	cmd := args.Command
 	expandedCmd, err := git.Alias(cmd)
@@ -187,4 +197,27 @@ func splitAliasCmd(cmd string) ([]string, error) {
 	}
 
 	return words, nil
+}
+
+func getListOfBuiltInCommands() map[string]bool {
+	gitHelpOutput, _ := cmd.New("git help -a").CombinedOutput()
+	builtInCommands := parseGitHelpOutput(gitHelpOutput)
+	for hubCommand, _ := range CmdRunner.All() {
+		builtInCommands[hubCommand] = true
+	}
+	return builtInCommands
+}
+
+func parseGitHelpOutput(gitHelpOutput string) map[string]bool {
+	builtInCommands := make(map[string]bool)
+	re := regexp.MustCompile(`(?m)^  ([a-z- 0-9]+)`)
+	for _, commands := range re.FindAllString(gitHelpOutput, -1) {
+		for _, command := range strings.Split(commands, " ") {
+			trimmedCommand := strings.Trim(command, " ")
+			if trimmedCommand != "" {
+				builtInCommands[trimmedCommand] = true
+			}
+		}
+	}
+	return builtInCommands
 }
--- a/commands/runner.go
+++ b/commands/runner.go
@@ -90,9 +90,8 @@ func (r *Runner) Execute() ExecError {
 	utils.Check(err)
 
 	git.GlobalFlags = args.GlobalFlags // preserve git global flags
-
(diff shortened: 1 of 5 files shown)


commands/runner.go
```diff
@@ -188,3 +198,26 @@ func splitAliasCmd(cmd string) ([]string, error) {
...
+func getListOfBuiltInCommands() map[string]bool {
+	gitHelpOutput, _ := cmd.New("git help -a").CombinedOutput()
+	builtInCommands := parseGitHelpOutput(gitHelpOutput)
+	for hubCommand, _ := range CmdRunner.All() {
+		builtInCommands[hubCommand] = true
+	}
+	return builtInCommands
+}
+
+func parseGitHelpOutput(gitHelpOutput string) map[string]bool {
```
mislav: If you move this function to the `git/` package, then you can use `gitOutput("help", "-a")` and get output broken into lines. Then you can iterate through those lines and have a simpler regex that just verifies that a line starts with exactly 2 spaces, and `Split()` the whole line without having to `FindAllString`.

features/clone.feature
```diff
@@ -286,3 +286,17 @@ Feature: hub clone
...
+  Scenario: If alias named clone exists, it should not be expanded.
+      Given the GitHub API server:
+        """
+        get('/repos/rtomayko/ronn') {
+          json :private => false,
+               :name => 'ronn', :owner => { :login => 'rtomayko' },
+               :permissions => { :push => false }
+        }
+        """
+      When I successfully run `git config --global alias.clone "clone -v"`
```
mislav: Wouldn't this test have passed even if the alias got expanded to `clone -v`?

commands/runner.go
```diff
@@ -188,3 +198,26 @@ func splitAliasCmd(cmd string) ([]string, error) {
 
 	return words, nil
 }
+
+func getListOfBuiltInCommands() map[string]bool {
+	gitHelpOutput, _ := cmd.New("git help -a").CombinedOutput()
+	builtInCommands := parseGitHelpOutput(gitHelpOutput)
+	for hubCommand, _ := range CmdRunner.All() {
+		builtInCommands[hubCommand] = true
```
mislav: It's not necessary that we write custom commands into `builtInCommands` hash just for the purpose of the check. Instead, `isBuiltInCommand` could iterate over each command in `CmdRunner.All()` and return `true` early if any command matched. Otherwise, it should proceed getting a list of commands from `git help -a`.

features/clone.feature
akshatgoel: No.
If alias is expanded, command run will be `git clone -v git://github.com/rtomayko/ronn.git\n`. However, step `Then it should clone "git://github.com/rtomayko/ronn.git"` expects that command run is actually `git clone git://github.com/rtomayko/ronn.git\n`. If alias is not expanded (which is success behavior for this test), the latter command is run.

commands/runner.go
akshatgoel: I checked this. gitOutput trims the spaces in command output. Thus, we can not differentiate command list from gitOutput response. I can make a non-trimming version of gitOutput which we can use. Let me know your thoughts?

mislav: You could change `gitOutput` to only trim the trailing newline. I don't think that should break anything.

features/clone.feature
akshatgoel: I have removed this test and added another more intuitive and non-conflicting test to alias `branch` to  `branch -a`. However, I am not sure the feature it should go to. I am thinking of creating a new file named git.feature to put tests which are related to git commands.

commands/runner.go
akshatgoel: Almost. It breaks a test in sync.feature and rightly so. We get list of branches using `branch --list` and trim the spaces on both ends before returning from gitOutput. We then do rev-parse on the list we get. If we don't trim the list, we end up checking for something like `refs/remotes/origin/  bugfix`. Adding trim to git.LocalBranches() instead of gitOutput fixes this.

mislav: > Adding trim to git.LocalBranches() instead of gitOutput fixes this.

👍

features/clone.feature
mislav: > I am thinking of creating a new file named git.feature to put tests which are related to git commands.

Sure, if you think that's best. Maybe name the file `git_compatibility.feature` for clarity?

akshatgoel:
Apologies for spamming the commit history and PRs.
I am still trying to setup github workflow on my machine. For future, I will be reviewing changes locally using UI tool so that I don't ruin the commit history.

mislav:
> For future, I will be reviewing changes locally using UI tool so that I don't ruin the commit history.

That's not a problem at all. Since you're the only person working on this branch at the moment, you should even feel free to force-push to this branch with cleaned-up history. What I often do on my own projects or contributions to other projects is that I, when the code is finalized and approved, do `git rebase -i origin/master` locally, squash the commits that do small fixes/tweaks together, and force-push the result to the same PR.

Many people frown upon force-pushing, but I think it's a perfectly valid method of housekeeping when you are working on some code by yourself.

mislav:
Thanks @akshatgoel!

PR #1342, opened 2016-11-18
Add ability to specify remote per-command

### LINKED ISSUE #1297: Support multi-level remotes
In my setup, my company has an internal GitHub Enterprise server and we contribute patches to an external github.com repository. So I have a setup that looks like this (names I'll be using in `[]`):

    github.com/project/repo [upstream]
               | |
                V
    github.company.com/team/repo-fork [origin]
               | |
                V
    github.company.com/user/repo-fork-fork [user]

I have branches where I sometimes need to be able to interact with that sometimes go to `origin` and sometimes go to `upstream`. As far as I can tell, it's not possible to create an issue (or any other features like `pull-request`, etc.) in `origin` and then create an issue in `upstream` without doing a bunch of remote renaming.

I'd love to see an command line option to specify the remote name for all the major commands. Off the top of my head, `issue` and `pull-request` would be priorities, but all commands that interact with a remote repository would be great.

wesbland:
Adds a flag to all appropriate commands to specify the remote to which the
command should apply. If no remote is specified, the usual hierarchy of origin
naming is adhered to.

Fixes #1297


mislav:
Thanks! I'm just wondering, have you considered:

1. `--remote` flag could be sort of a "global" hub flag, that would be processed and saved somewhere before dispatching to the actual hub command. That way we could avoid duplicating the processing code within every individual command.

2. I was thinking of adding a `--repo` global hub flag where you can specify a `<owner>/<repo>` pair that explicitly selects a repository to operate on, so you can execute commands outside of a local git repository. Do you think having such a command would supersede the `--remote` command?

wesbland:
I did consider the `--remote` flag, but there were so many commands where it didn't apply that I didn't think it was worth it. I think it would also still require at least some repetition to check whether to use the specified remote or the default one. It's your show though. If you think the global flag is a better idea, it's fine with me.

I think I mentioned this on the issue (#1297), but I'd personally prefer `--remote` to `--repo` because it would work better with the enterprise side. I go back and forth between the two (even with the same local repository) so it would be a little gross to have to change flags based on which remote I'm talking about. I can understand the allure of being able to create an issue without having to check out the repository/add the remote though, so maybe both would be ok.

Looks like I broke a bunch of the tests though so I'll have to take another look to see what happened. In the meantime, I'd be interested to hear if this direction seems reasonable before I keep going down this road.

wesbland:
That seems to have fixed the failed tests (needed to update some output on an error when forking).

I still need to add new tests for the new flags.

wesbland:
Looks like there were some conflicts after some new patches. I had to change the `pull-request` command's remote flag to be `-t`. I could change them all to be consistent, but for now I left the rest as `-r`.

rouzier:
In order to keep things consistent you could use -R instead

frwickst:
@wesbland Any update regarding this? 
I was about to start using _hub_ but then had to stop when I noticed that it forces me to use `origin` as remote for pull-requests (and for many other things as well it seems).

wesbland:
@frwickst - I updated the PR not long ago (March 9th according to the commit date). I think the main thing holding this PR up is the lack of testing for the new flags. I think I made an attempt at writing good tests with my last PR https://github.com/github/hub/pull/1327 and it didn't go well.

I don't know if @mislav has an opinion on this one or not since it's been open for quite a while.

I did change things per the suggestion above by @rouzier to use `-R` for the short flag.

wesbland:
Oops. Forgot the `pr` command. That's added now too.

frwickst:
@mislav Is there anything holding this PR back?

wesbland:
Updated this PR again to resolve the merge conflicts with the latest adds.

Issue #1346, opened 2016-11-26
pull-request repeatedly hitting 404


RichardLitt:
At some point in the last few weeks, `git pull-request` broke, and I am not sure why. 

```sh
> git pull-request 
Error creating pull request: Not Found (HTTP 404)
Not Found
Are you sure that github.com/RichardLitt/university-of-remainia exists?
```

Yes, I am sure that it exists. 

```
origin	git@github.com:RichardLitt/university-of-remainia.git (fetch)
origin	git@github.com:RichardLitt/university-of-remainia.git (push)
```

I've tried reinstalling hub using homebrew, but the issue remains. Any ideas?

mislav:
I'm sorry that you're having trouble. Using `HUB_VERBOSE=1 hub pull-request` you can get a verbose HTTP log of all that's happening. Maybe that will shed some light on the subject.

Is it possible that you revoked the Personal Access Token that hub has stored in `~/.config/hub`?

Also, which hub version?

mislav:
Can you upgrade hub and try again?

RichardLitt:
Uh...

```
🐕  hub --version
git version 2.3.0
hub version 2.2.0
21:34 ~/src/university-of-remainia (feat/added-sidebar) 🐕  brew upgrade hub
Error: hub 2.2.9 already installed
```

This is weird. Why isn't hub using the right version?

mislav:
@RichardLitt You probably have two hubs.

RichardLitt:
Ok. You were right. Removed both versions. 

```sh
06:38 ~ 🐕  hub
-bash: hub: command not found
06:38 ~ 🐕  brew install hub
Updating Homebrew...
==> Auto-updated Homebrew!
Updated 1 tap (homebrew/core).
==> Updated Formulae
agda                          ffmpeg ✔                      mpv

==> Downloading https://homebrew.bintray.com/bottles/hub-2.2.9.el_capitan.bottle.tar.gz
Already downloaded: /Users/richard/Library/Caches/Homebrew/hub-2.2.9.el_capitan.bottle.tar.gz
==> Pouring hub-2.2.9.el_capitan.bottle.tar.gz
==> Caveats
Bash completion has been installed to:
  /usr/local/etc/bash_completion.d

zsh completion has been installed to:
  /usr/local/share/zsh/site-functions
==> Summary
🍺  /usr/local/Cellar/hub/2.2.9: 10 files, 9M
06:39 ~ 🐕  cd src/university-of-remainia/
[...]

PR #1347, opened 2016-11-26
Added note about GitHub tokens

RichardLitt:
See #1346 for my confusion


mislav:
Thanks for the suggestion, but I think this is best handled from the program rather than a note in documentation. I intend to provide better error messages on authentication errors, and those messages should include the information about where the token was obtained from (`~/.config/hub` or GITHUB_TOKEN).

RichardLitt:
Works for me. Thanks.

PR #1353, opened 2016-12-05
Update go-update to the latest version

### LINKED ISSUE #1351: Update github.com/inconshreveable/go-update
The version of go-update in this repo is out of date, and refers to a package (go-update/download) that doesn't exist in the master version of that project. 

This causes problems for people that are revendoring; for example https://github.com/Shyp/go-git/blob/master/git.go#L153-L156 imports github.com/github/hub/commands as a library, and an upstream package main wants to use its own vendor directory. This means we have to manually set in place an older version of that package.

I think it'd be easier for everyone if the go-update API's used in this package were updated to the latest version and the included dependency was also updated.

kevinburke:
Fixes #1351.


kevinburke:
I couldn't get the tests to pass locally and somehow `git amend` is now telling me

```
$FOO "${BAR}": : command not found
error: There was a problem with the editor '$FOO "${BAR}"'.
Please supply the message using either -m or -F option.
```

kevinburke:
I think this branch is failing for the same reason `master` is failing.

kevinburke:
The Go parts passed at least

mislav:
I've got rid of the dependency instead. Thanks for reporting! d400838ecf7653923713b4c3a84dd625c5b5ce7a

PR #1357, opened 2016-12-07
Add --set-upstream to git push commands.

nex3:
This ensures that a branch pushed using "hub pull-request" has its
upstream branch set to the target of the push.


mislav:
Thanks!

PR #1375, opened 2017-01-06
Add clipboard copy (-c) to browse command

tomas-edwardsson:
The -c parameter will put the url into the clipboard instead of opening in the browser.


mislav:
Looks good! Usually I would ask for additional tests, but since I don't have a good practice established yet for testing clipboard changes, this is enough for now.

PR #1396, opened 2017-02-03
Set merge configuration on checkout.

nex3:
When checking out a pull request that doesn't match an existing remote,
this sets the branch's remote and merge configuration fields. If the
pull request is modifiable by the maintainer, it uses the SSH URL to
that you can push to the pull request.

### DIFF
--- a/commands/checkout.go
+++ b/commands/checkout.go
@@ -119,9 +119,19 @@ func transformCheckoutArgs(args *Args) error {
 				newBranchName = fmt.Sprintf("%s-%s", pullRequest.Head.Repo.Owner.Login, pullRequest.Head.Ref)
 			}
 		}
-		refSpec := fmt.Sprintf("refs/pull/%s/head:%s", id, newBranchName)
 		newArgs = append(newArgs, newBranchName)
-		args.Before("git", "fetch", baseRemote.Name, refSpec)
+
+		ref := fmt.Sprintf("refs/pull/%s/head", id)
+		args.Before("git", "fetch", baseRemote.Name, fmt.Sprintf("%s:%s", ref, newBranchName))
+
+		remote := baseRemote.Name
+		mergeRef := ref
+		if pullRequest.MaintainerCanModify && pullRequest.Head.Repo != nil {
+			remote = pullRequest.Head.Repo.SshUrl
+			mergeRef = fmt.Sprintf("refs/heads/%s", pullRequest.Head.Ref)
+		}
+		args.Before("git", "config", fmt.Sprintf("branch.%s.remote", newBranchName), remote)
+		args.Before("git", "config", fmt.Sprintf("branch.%s.merge", newBranchName), mergeRef)
 	}
 	replaceCheckoutParam(args, checkoutURL, newArgs...)
 	return nil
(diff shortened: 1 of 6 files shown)


features/support/fakebin/git
```diff
@@ -4,7 +4,7 @@
 # asserted that they ran.
 
 command="$1"
-[ "$command" = "config" ] || echo git "$@" >> "$HOME"/.history
+[ "$command" = "config" -a "${#@}" -le 2 ] || echo git "$@" >> "$HOME"/.history
```
mislav: A shorter expression for `${#@}` is `$#`. But, I don't think that we need to log the `git config` statements into `.history` log file. Instead, couldn't we read from `git config` to verify that this configuration is being set? We could add a new custom step that reads like:

```
Then the branch should merge "refs/heads/fixes" from remote "git@github.com:mislav/jekyll.git"
```

and the step could use `git config` internally to extract this information. How does that sound?

commands/checkout.go
```diff
@@ -119,9 +119,19 @@ func transformCheckoutArgs(args *Args) error {
...
 		newArgs = append(newArgs, newBranchName)
-		args.Before("git", "fetch", baseRemote.Name, refSpec)
+
+		ref := fmt.Sprintf("refs/pull/%s/head", id)
+		args.Before("git", "fetch", baseRemote.Name, fmt.Sprintf("%s:%s", ref, newBranchName))
+
+		remote := baseRemote.Name
+		mergeRef := ref
+		if pullRequest.MaintainerCanModify && pullRequest.Head.Repo != nil {
+			remote = pullRequest.Head.Repo.SshUrl
```
mislav: People who have `hub.protocol` set to "https" will want that this URL is `https:`, because they might not have SSH keys configured for GitHub, and instead use HTTPS authentication with credential helper.

Perhaps you could [convert a Repo object to Project](https://github.com/github/hub/blob/fe5b59b5489f2ac647c51914bf8e1b0e9993451f/commands/cherry_pick.go#L86) and [use its `GitURL()` method](https://github.com/github/hub/blob/fe5b59b5489f2ac647c51914bf8e1b0e9993451f/github/project.go#L75-L78)? Then you don't have to worry about the logic of constructing a proper clone/push URL, and you don't have to add the extra `SshUrl` property to the struct either.

commands/checkout.go
```diff
@@ -119,9 +119,19 @@ func transformCheckoutArgs(args *Args) error {
...
+			mergeRef = fmt.Sprintf("refs/heads/%s", pullRequest.Head.Ref)
+		}
+		args.Before("git", "config", fmt.Sprintf("branch.%s.remote", newBranchName), remote)
+		args.Before("git", "config", fmt.Sprintf("branch.%s.merge", newBranchName), mergeRef)
```
mislav: Is it really necessary to set this configuration in the case when the PR is coming from the same repository, i.e. **not** from a maintainer-modifiable fork? I thought that [we already configure such branches properly](https://github.com/github/hub/blob/fe5b59b5489f2ac647c51914bf8e1b0e9993451f/features/checkout.feature#L97-L99).

features/support/fakebin/git
nex3: Done.

commands/checkout.go
nex3: Done.

commands/checkout.go
nex3: If the PR is coming from the same repository, `headRemote` won't be `nil` and we'll hit the previous branch. But if it's coming from a different repository but is *not* maintainer-modifiable, we still want to set the remote so `git pull` works.

mislav:
Thanks Natalie! This was a high quality pull request. 👍

PR #1403, opened 2017-02-10
Add a "hub pr checkout" command

walles:
It wraps functionality already provided by "hub checkout". The advantage
with "hub pr checkout" are twofold:

* Discoverability. It wasn't until I started browsing the hub source
  code that I realized this functionality was present.
* Usability. Just typing the PR number is easier than having to figure
  out its URL.


mislav:
Thanks for proposing this. Both your advantage points are valid. However, I'm reluctant to add a command called `review` because:

1. It mostly duplicates functionality of another existing command;
2. Users might think that the `review` command provides functionality related to GitHub Reviews (e.g. requesting a review; leaving an approved review, requesting changes, responding to a review).

Instead, I would like to make it easier to check out a PR by its number rather than pasting in a whole URL. How does that sound?

walles:
I'm fine with fixing this in some other way, as long as the solution addresses both the discoverability and the usability problems.

For the usability part, just moving this PR's code into the `checkout` sub command should be fine, but what would you propose for discoverability?

I'd like the existence of this feature to be fairly obvious from all of:
* `hub help` output
* `hub checkout --help` output
* `hub help checkout` output

@mislav, what's your alternative suggestion that covers this? Or would just renaming the new subcommand to something that doesn't have `review` in it be a way forward?

mislav:
I don't have a better idea for a separate command name. I still think `checkout` most closely describes what we're doing, since we're switching to another branch.

You have a good point about this not being discoverable. We intentionally don't want to change git's own help system. So, `git help checkout` shows you git documentation without any hub extensions. We don't want to try to hook into that process and add extra lines. Instead, you can do `hub help hub-checkout` (this is available in 2.3.0-prereleases) to access the man page for what extensions hub provides on top of this command. Or, you can do `hub help hub` for a overview of everything that hub does.

walles:
@mislav, how about `hub pull-request checkout 1234`?

And then updating the help text for both `hub help` and `hub pull-request` to document the new sub command.

That would be discoverable and not clash with anything, wouldn't it?

mislav:
That's not a bad idea. People are already requesting a command to list all the pull requests, similar to `hub issue`. Maybe this should

- `hub pr`: list pull requests
- `hub pr create`: create a pull request (previously `hub pull-request`)
- `hub pr checkout 123`: checkout a pull request
- `hub pr open 123`: open the pull request in the web browser

I don't think we should overload the existing `hub pull-request` command since that would be lengthy to type and possibly confusing to users. Thoughts?

walles:
If you think a new `hub pr <subcommand>` is the way to go I'm fine with that.

Unless somebody else beats me to it I'll turn this PR into a `hub pr checkout` subcommand.

walles:
Implementation changed to:
* `hub pr`: print `hub pr` help text
* `hub pr checkout 123`: checkout PR 123 source code

noam-c:
Modest suggestion: would there be any opposition to `hub pr new` instead of `hub pr create` for making a PR?
If there are already `create` subcommands elsewhere in hub, then please ignore this comment for consistency's sake.

walles:
@mislav, ready for re-review.

@noam-c, I primarily want to get this in. After the new `pr` command is in, giving it additional sub commands is something that can be discussed in separate PRs / tickets.

commands/pr.go
```diff
@@ -0,0 +1,75 @@
...
+	baseProject, err := localRepo.MainProject()
+	utils.Check(err)
+	host, err := github.CurrentConfig().PromptForHost(baseProject.Host)
+	utils.Check(err)
+	client := github.NewClientWithHost(host)
+	pr, err := client.PullRequest(baseProject, prNumberString)
+	utils.Check(err)
+
+	if args.ParamsSize() == 1 {
+		args.Replace(args.Executable, "checkout", pr.HtmlUrl)
```
mislav: It's a cool hack how you tweaked an argument list to call into existing `transformCheckoutArgs()` function to avoid replicating a lot of existing `hub checkout` logic.

However, this approach looks up the pull request in the API twice: once in your function, and the 2nd time within `transformCheckoutArgs`. It would be great if you could split `transformCheckoutArgs` function into two functions, of which one would accept a PullRequest object.

walles:
@mislav, ready for re-review.

Tests added, redundant PR fetching removed.

mislav:
Thanks for the updates. I didn't like how the complex logic was copy-pasted between `checkout` and `pr`, so I've pushed my proposed solution for `pr` command to be able to invoke `transformCheckoutArgs` and thus get all that logic for free. Can you take a look and see if that looks alright?

Also, I've removed most scenarios from `pr.feature` since only a few basic ones are needed, mostly just to assert command-line argument processing. The rest of the complex logic (implemented within `transformCheckoutArgs`) is already verified through the scenarios in `checkout.feature`.

walles:
@mislav will you accept this PR in its current form?

I'm fine with it, +1!

walles:
Thanks @mislav :)

mislav:
@walles Thanks for your patience!

PR #1412, opened 2017-02-15
Add a `--org` flag to `hub fork`

### LINKED ISSUE #505: Feature request: optional organization argument to `fork`, etc.
Currently, `hub fork` and `hub pull-request` only will create forks in [user]/[repo].

I'd really like to be able to (optionally) fork the project under an [organization]/[repo].

There are a number of reasons why this would make sense, but here's mine: I always put projects for which I'm submitting PRs into an organization I've created [for this particular purpose](https://github.com/pullreq) since, well, [some people aren't so good](https://github.com/brianmario/charlock_holmes/pull/55#issuecomment-30437830) at following up on PRs in a timely manner; and I prefer not to clutter up "my" repos with a bunch of forks which need to remain for an indeterminate length of time.

I'd hack on this myself, but for the "hub's gonna go go" business.

blackjid:
Fixes #505

### DIFF
--- a/commands/fork.go
+++ b/commands/fork.go
@@ -10,17 +10,24 @@ import (
 
 var cmdFork = &Command{
 	Run:   fork,
-	Usage: "fork [--no-remote]",
+	Usage: "fork [--no-remote] [--org=<ORGANIZATION>]",
 	Long: `Fork the current project on GitHub and add a git remote for it.
 
 ## Options:
 	--no-remote
 		Skip adding a git remote for the fork.
 
+	--org=<ORGANIZATION>
+		Fork the repository within this organization.
+
 ## Examples:
 		$ hub fork
 		[ repo forked on GitHub ]
 		> git remote add -f USER git@github.com:USER/REPO.git
+
+		$ hub fork --org=ORGANIZATION
+		[ repo forked on Github into the ORGANIZATION organization]
+		> git remote add -f ORGANIZATION git@github.com:ORGANIZATION/REPO.git
 
 ## See also:
 
@@ -28,10 +35,15 @@ hub-clone(1), hub(1)
 `,
 }
 
-var flagForkNoRemote bool
+var (
+	flagForkNoRemote bool
+
+	flagForkOrganization string
+)
 
 func init() {
 	cmdFork.Flag.BoolVar(&flagForkNoRemote, "no-remote", false, "")
+	cmdFork.Flag.StringVarP(&flagForkOrganization, "org", "", "", "ORGANIZATION")
 
 	CmdRunner.Use(cmdFork)
 }
@@ -56,7 +68,14 @@ func fork(cmd *Command, args *Args) {
 		utils.Check(fmt.Errorf("Error creating fork: %s", err))
 	}
 
-	forkProject := github.NewProject(host.User, project.Name, project.Host)
+	params := map[string]interface{}{}
+	forkOwner := host.User
+	if flagForkOrganization != "" {
+		forkOwner = flagForkOrganization
+		params["organization"] = forkOwner
+	}
+
+	forkProject := github.NewProject(forkOwner, project.Name, project.Host)
 	newRemoteName := forkProject.Owner
 
 	client := github.NewClient(project.Host)
@@ -73,7 +92,7 @@ func fork(cmd *Command, args *Args) {
 		}
 	} else {
 		if !args.Noop {
-			newRepo, err := client.ForkRepository(project)
+			newRepo, err := client.ForkRepository(project, params)
 			utils.Check(err)
 			forkProject.Owner = newRepo.Owner.Login
 			forkProject.Name = newRepo.Name
--- a/features/fork.feature
+++ b/features/fork.feature
@@ -13,6 +13,7 @@ Feature: hub fork
       }
       get('/repos/mislav/dotfiles', :host_name => 'api.github.com') { 404 }
       post('/repos/evilchelu/dotfiles/forks', :host_name => 'api.github.com') {
+        assert :organization => nil
         status 202
         json :name => 'dotfiles', :owner => { :login => 'mislav' }
       }
@@ -202,3 +203,17 @@ Scenario: Related fork already exists
     And "git.my.org" is a whitelisted Enterprise host
     When I successfully run `hub fork`
     Then the url for "mislav" should be "git@git.my.org:mislav/dotfiles.git"
+
+  Scenario: Fork a repo to a specific organization
+    Given the GitHub API server:
+      """
+      get('/repos/acme/dotfiles') { 404 }
+      post('/repos/evilchelu/dotfiles/forks') {
+        assert :organization => "acme"
+        status 202
+        json :name => 'dotfiles', :owner => { :login => 'acme' }
+      }
+      """
+    When I successfully run `hub fork --org=acme`
+    Then the output should contain exactly "new remote: acme\n"
+    Then the url for "acme" should be "git@github.com:acme/dotfiles.git"
--- a/github/client.go
+++ b/github/client.go
@@ -419,13 +419,12 @@ type RepositoryPermissions struct {
 	Pull  bool `json:"pull"`
 }
 
-func (client *Client) ForkRepository(project *Project) (repo *Repository, err error) {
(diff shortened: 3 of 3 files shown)


features/fork.feature
```diff
@@ -202,3 +202,22 @@ Scenario: Related fork already exists
...
+        halt 400 unless request.env['HTTP_X_ORIGINAL_SCHEME'] == 'https'
+        halt 401 unless request.env['HTTP_AUTHORIZATION'] == 'token OTOKEN'
+      }
+      get('/repos/github/dotfiles', :host_name => 'api.github.com') { 404 }
+      post('/repos/evilchelu/dotfiles/forks', :host_name => 'api.github.com') {
+        status 202
+        json :name => 'dotfiles', :owner => { :login => 'mislav' }
+      }
+      """
+    When I successfully run `hub fork --org github`
```
blackjid: For some reason, when I execute the command with `--org` I need to give the value with this syntax

`--org=github`  instead of `--org github`

Without the `=` sign I get an error 

```
flag needs an argument: --org
```

Any idea what I'm missing?

mislav: I think it might be a limitation of the flag-parsing package that we use? https://github.com/ogier/pflag#command-line-flag-syntax

Just use the `--org=ORG` syntax for now and I can try to figure this out later.

features/fork.feature
```diff
@@ -202,3 +202,22 @@ Scenario: Related fork already exists
     And "git.my.org" is a whitelisted Enterprise host
     When I successfully run `hub fork`
     Then the url for "mislav" should be "git@git.my.org:mislav/dotfiles.git"
+
+  Scenario: --org github
```
mislav: Maybe a human-readable scenario name, like "Fork a repo to a specific organization"?

features/fork.feature
```diff
@@ -202,3 +202,22 @@ Scenario: Related fork already exists
...
+    Given the GitHub API server:
+      """
+      before {
```
mislav: This `before` block is not needed for this scenario. It was desired for some other scenarios, but doesn't need to be copy/pasted further.

features/fork.feature
```diff
@@ -202,3 +202,22 @@ Scenario: Related fork already exists
...
+        halt 400 unless request.env['HTTP_X_ORIGINAL_SCHEME'] == 'https'
+        halt 401 unless request.env['HTTP_AUTHORIZATION'] == 'token OTOKEN'
+      }
+      get('/repos/github/dotfiles', :host_name => 'api.github.com') { 404 }
```
mislav: The `:host_name` verification is not needed in this scenario. You can take it out.

features/fork.feature
```diff
@@ -202,3 +202,22 @@ Scenario: Related fork already exists
...
+      post('/repos/evilchelu/dotfiles/forks', :host_name => 'api.github.com') {
+        status 202
```
mislav: This `post` block should verify that the org parameter was passed:

```rb
assert :organization => "github"
```

features/fork.feature
```diff
@@ -202,3 +202,22 @@ Scenario: Related fork already exists
...
+        json :name => 'dotfiles', :owner => { :login => 'mislav' }
```
mislav: The `:login` should be "github", not "mislav"

features/fork.feature
mislav: Also, could you use another organization name in tests, like "acme" (a made-up name)? Using "github" in tests can be confusing since a lot of things around the codebase are also named "github".

features/fork.feature
```diff
@@ -202,3 +202,22 @@ Scenario: Related fork already exists
...
+      }
+      """
+    When I successfully run `hub fork --org github`
+    Then the output should contain exactly "new remote: github\n"
+    And "git remote add -f github git://github.com/evilchelu/dotfiles.git" should be run
+    And "git remote set-url github git@github.com:github/dotfiles.git" should be run
```
mislav: The lines that verify `git remote add/set-url` were called are not needed in this scenario.

commands/fork.go
```diff
@@ -73,9 +92,9 @@ func fork(cmd *Command, args *Args) {
 		}
 	} else {
 		if !args.Noop {
-			newRepo, err := client.ForkRepository(project)
+			newRepo, err := client.ForkRepository(project, params)
 			utils.Check(err)
-			forkProject.Owner = newRepo.Owner.Login
```
mislav: This line should stay the way it was. We want to read the information about the new repo from the response from the server.

features/fork.feature
blackjid: perfect... I'll do that

features/fork.feature
blackjid: ok, I was following the no-remote scenario style... But I prefer the one you propose

features/fork.feature
blackjid: 👍

features/fork.feature
blackjid: I really though I already added this assert... :)

blackjid:
@mislav thanks for the comments.

I think I address all the issues.

Just a though... how do you feel about using `--org` flag? which is not short nor long??
I feel it should be something like `-o, --organization`, but:

- organization is really super long to type.
- the `-o` is usually used for `--open` in a browser.
- Based on the comments on this issue https://github.com/ogier/pflag/issues/17 the short flag `-o` should work with a space, but not the long one.

The other syntax I played with that I feel could work well is

```
hub fork [<ORGANIZATION>]
hub fork acme
```

Just like the `create` command accept `[<ORGANIZATION/>USER]`

well, let me know if we need more changes and before merging so I can squash the commits.

commands/fork.go
```diff
@@ -10,28 +10,40 @@ import (
 
 var cmdFork = &Command{
 	Run:   fork,
-	Usage: "fork [--no-remote]",
+	Usage: "fork [--no-remote] [--org <ORGANIZATION>]",
```
mislav: Because we're using `--org=` syntax for now, docs should reflect that too.

commands/fork.go
```diff
@@ -10,28 +10,40 @@ import (
...
 	Long: `Fork the current project on GitHub and add a git remote for it.
 
 ## Options:
 	--no-remote
 		Skip adding a git remote for the fork.
 
+	--org <ORGANIZATION>
```
mislav: Ditto `=`

commands/fork.go
```diff
@@ -10,28 +10,40 @@ import (
...
+		Fork the repository within this organization.
+
 ## Examples:
 		$ hub fork
 		[ repo forked on GitHub ]
 		> git remote add -f USER git@github.com:USER/REPO.git
 
+		$ hub fork --org ORGANIZATION
```
mislav: Ditto `=`

features/fork.feature
```diff
@@ -202,3 +202,17 @@ Scenario: Related fork already exists
     And "git.my.org" is a whitelisted Enterprise host
     When I successfully run `hub fork`
     Then the url for "mislav" should be "git@git.my.org:mislav/dotfiles.git"
+
+  Scenario: Fork a repo to a specific organization
+    Given the GitHub API server:
+      """
+      get('/repos/acme/dotfiles') { 404 }
+      post('/repos/evilchelu/dotfiles/forks') {
+        assert :organization => "acme"
```
mislav: This is good. It just occurred to me that we should add an assert to one of the vanilla scenarios (no `--org` parameter) that `:organization` is nil. Could you add that too?

blackjid: 👍

blackjid:
Done, I went ahead and squash the commits...
Let me know if there is something else...

features/fork.feature
```diff
@@ -13,6 +13,7 @@ Feature: hub fork
       }
       get('/repos/mislav/dotfiles', :host_name => 'api.github.com') { 404 }
       post('/repos/evilchelu/dotfiles/forks', :host_name => 'api.github.com') {
+        assert :organization => nil
```
blackjid: here I added the assert

mislav:
Thanks for your patience and prompt updates!

blackjid:
Your welcome! thank you!

kasbah:
Hey, thanks for this. Very useful. This doesn't yet work with `hub pull-request`. Is that right?

Issue #1303, opened 2016-10-12
List pull requests again


vermiculus:
#1232 mentions the need for a command to list out pull requests (taken away by 8563e678d2f3cfc8d1bcc4ed3f38f7226c5d5012, probably for semantic reasons), but I thought I'd reiterate the need here.

Related to vermiculus/magithub#44.

akshatgoel:
@mislav I am looking to work on this issue. Here is my current understanding of the issue:
Currently hub only has commands to create a PR but not list them. We should be able to list PRs as we do with issues right now (as added by the commit).

Proposal:
I am thinking of moving current pull-request behavior to sub-command create to maintain parity with issue as well as to make listing (which is stateless operation compared to create) as default.

What do you think?

mislav:
@akshatgoel That's a good plan. How do you envision this command and its subcommands to be called?

timja-kainos:
Alternatively `hub ls pull-request`?

vermiculus:
Even if there could be an option to skip [this section of code](https://github.com/github/hub/blob/master/commands/issue.go#L243), that would be sufficient:

```
for _, issue := range issues {
	if issue.PullRequest != nil {
		continue
	}
	ui.Printf(formatIssue(issue, flagIssueFormat, colorize))
}
```

vermiculus:
@mislav If such a flag was added this week, do you have an estimate on when the change would make it into a release?

In other words, any idea when `2.3.0-pre9` will be tagged?

mislav:
I could tag another prerelease after that. Thank you!

PR #1413, opened 2017-02-15
Add option to show pull requests again

vermiculus:
This change adds a long-form flag `--show-pulls` to `hub issue` that directs hub to include pull requests in the issue listing.  The default value is false; the default behavior does not change.

Temporary fix until a better abstraction for pull requests can be designed in #1303.

Related: vermiculus/magithub#44

### DIFF
--- a/commands/issue.go
+++ b/commands/issue.go
@@ -131,8 +131,8 @@ With no arguments, show a list of open issues.
 	-^ --sort-ascending
 		Sort by ascending dates instead of descending.
 
-	--show-pulls
-		Show pull requests as well.
+	--include-pulls
+		Include pull requests as well as issues.
 `,
 	}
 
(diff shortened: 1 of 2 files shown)


commands/issue.go
```diff
@@ -130,6 +130,9 @@ With no arguments, show a list of open issues.
 
 	-^ --sort-ascending
 		Sort by ascending dates instead of descending.
+
+	--show-pulls
```
mislav: Could this be called `--include-pulls` instead?

mislav:
Ooops. I guess I should have “requested changes” instead of approving this PR 😅 But it's on the right track! 👍

vermiculus:
@mislav There you go! 😄

mislav:
Looks good!

withrocks:
I'm interested in using this feature to list which of my local branches have an open pull request on github.

In order to do that, I hoped I could get the name of the branch the pull request is based on (e.g. as one of the format placeholders). Is that supported?

mislav:
@withrocks This would be an excellent feature, but it isn't supported because the API endpoint queried here doesn't contain extended data about PRs, such as their head and base branches. https://developer.github.com/v3/issues/#list-issues-for-a-repository

We would probably need to create a `hub pr list` subcommand to query the Pull Requests endpoint specifically. https://developer.github.com/v3/pulls/#list-pull-requests

Issue #1406, opened 2017-02-12
How to use hub with custom http URL without SSL


Swoorup:

vermiculus:
What exactly is the problem you're running into? What's its current behavior?

It's worth noting that [the normal GitHub API can *only* be accessed over HTTPS](https://developer.github.com/v3/#schema).

lwiechec:
hi there, I _think_ I have a similar 'issue'. We are running GitHub enterprise internally, and it is running on `http://` (so, no SSL). I have configured hub to use our GitHub hostname (using git config's `hub.host` setting) but I guess hub is hardcoded to use `https://` here.

I know that HTTPS is a way to go, but for GitHub Enterprise installations, when they sometimes are on isolated networks (like in our case), there is no need for HTTPS. It would be good to have an option here, IMHO

Swoorup:
That is exactly my problem. ^

mislav:
Right now, you have to configure this manually within the `~/.config/hub` file. Find the existing config for your GHE host and add the `protocol` line:

```yml
example.com:
- user: mislav
  oauth_token: REDACTED
  protocol: http
```

lwiechec:
@mislav hmm, I thing it should be in `~/.config/hub`, not `~/.hub/config` :)

Swoorup:
@mislav I can confirm that does not work.

mislav:
@Swoorup What exactly did you try? (Please show `~/.config/hub`, which sensitive data redacted.) What command did you run next? What happened instead?

This feature was around since hub 2.2.0. It's a bit of a hassle, but it always worked. 🤔

lwiechec:
I can confirm that it works.

Swoorup:
```
github:
- user: swoorup
  oauth_token: REDACTED
  protocol: http
```

I ran the `hub issue` command on the repository whose URL is in format of `git://`. It then asked me, for my username and password. After entering it, it would take quite a while and show me errors. I could see it tried to access a https site instead.

mislav:
In the first line of what you pasted there is `github:`. Instead of that, it should be the exact name of your host where GHE is installed, in the same way it appears in the `git://HOSTNAME/...` url for your git remote. Can you double-check that it matches?

Swoorup:
Its github :) without the [dot] com

mislav:
[...]

PR #1414, opened 2017-02-15
adding docs on how to configure hub on GHE on HTTP

lwiechec:
as explained by @mislav in #1406 

(I though this might bring some added value. I don't know if the docs are generated with some tool so - I wrote the 'raw' HTML)


mislav:
Thanks, but I don't want this to stay the official method of configuring GHE hosts for HTTP. It's a hassle. Instead, I want us to have a dedicated command for authentication #225. With that, people will be able to configure GHE hosts much more easily regardless of it being over HTTPS or HTTP. I plan this for the next release. Then, this documentation would immediately be obsolete.

PR #1417, opened 2017-02-15
Update LICENSE Copyright Date to 2017

Phineas:


mislav:
Thanks, but I'll avoid bumping up the date just yet. Having the date updated every year is not strictly necessary (see http://ben.balter.com/2015/06/03/copyright-notices-for-websites-and-open-source-projects/). There might be times when we revise the copyright statement and the license, but we prefer to do it on our own time rather than via non-maintainer contributions.

My general advice is that it doesn't make much sense to send pull requests for other people's LICENSE files. That's always a tricky subject. Instead, please focus on other parts of the project. Thanks!

Phineas:
@mislav Would it be better to remove it altogether then?

mislav:
Well, the copyright notice was there in the original project (while it was still being done by Chris Wanstrath), and there's no real reason to remove it. We might update it to reflect who is really behind this project right now (GitHub as a compay, as opposed to a single individual). I will need to consult with my colleagues about that.

PR #1425, opened 2017-03-02
Add plan9port rc shell support

pocket7878:
Add plan9port rc shell support to `hub alias`

### DIFF
--- a/commands/alias.go
+++ b/commands/alias.go
@@ -52,7 +52,7 @@ func alias(command *Command, args *Args) {
 		utils.Check(fmt.Errorf("Error: couldn't detect shell type. Please specify your shell with `%s`", cmd))
 	}
 
-	shells := []string{"bash", "zsh", "sh", "ksh", "csh", "tcsh", "fish"}
+	shells := []string{"bash", "zsh", "sh", "ksh", "csh", "tcsh", "fish", "rc"}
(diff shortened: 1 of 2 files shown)


commands/alias.go
```diff
@@ -92,6 +94,8 @@ func alias(command *Command, args *Args) {
 			profile = "~/.cshrc"
 		case "tcsh":
 			profile = "~/.tcshrc"
+		case "rc":
+			profile = "$home/lib/profile"
```
mislav: Can we suggest `~` as a shorthand for `$home`, or is it not supported in rc?

commands/alias.go
```diff
@@ -105,6 +109,8 @@ func alias(command *Command, args *Args) {
 			eval = `function git --wraps hub --description 'Alias for hub, which wraps git to provide extra functionality with GitHub.'
 	hub $argv
 end`
+		case "rc":
+			eval = "eval `{hub alias -s}"
```
mislav: Does this syntax correctly support if the output of `hub alias -s` spans over multiple lines? In posix shells, we needed to use double quotes for that.

commands/alias.go
pocket7878: ~ expansion is not supported in rc.

mislav:
Thank you! This is great.

PR #1447, opened 2017-04-04
Add Fedora Linux installation instructions

jhasse:
The latest hub version is available in Fedora's official repositories: https://admin.fedoraproject.org/pkgdb/package/rpms/hub/


mislav:
Thank you!

Issue #1358, opened 2016-12-08
Support passing in reviewers when creating a pull request


JordonPhillips:
The concept of [reviewers](https://github.com/blog/2291-introducing-review-requests) was added to the frontend today. It would be great to be able to specify those on the command line when creating the pr.

mislav:
Good idea, but the feature is not yet available through the GitHub API.

dkpalmer:
Possibly available now via https://developer.github.com/v3/pulls/review_requests/#create-a-review-request ?

hkdobrev:
> The Pull Request Review API is currently available for developers to preview. During the preview period, the API may change without advance notice. Please see the blog post for full details.
> 
> To access the API during the preview period, you must provide a custom media type in the Accept header:
> 
> `application/vnd.github.black-cat-preview+json`

Does that mean only access tokens from developers in the early access program would be able to use this API or everybody will be if they provide this media type?

mislav:
Oh nice. That means hub only needs to set this `Accept` header and everything will work. This is not related to the Early Access program.

So, what would be the API for requesting reviewers through hub?

hkdobrev:
@mislav I would imagine a flag to the `pull-request` command similar to assignees would be welcome.

mislav:
Hmm how about:

```
	-r, --reviewer <USERS>
		A comma-separated list of GitHub handles to request a reviews from.
```

julien51:
Any idea why this not working with 2.2.9?

```
git pull-request -r benwerd
unknown shorthand flag: 'r' in -r
```

```
$ git --version
git version 2.2.1
hub version 2.2.9
```

Thanks!

harmnot:
is for `-r` need using his username ? or email ? could I do space also ??

mislav:
@harmnot `hub pull-request -r user1,user2,user3`

Just usernames, no spaces, no emails.

harmnot:
I see, is hub pull-request as merge request like on Gitlab ? because I want to merge request from my branch to Dev, what is that command for it ?

mislav:
Yes, a Pull Request is basically a merge request. https://help.github.com/en/articles/about-pull-requests

I'm not sure what exactly you're trying to do, but this might help: https://hub.github.com/hub-pull-request.1.html

harmnot:
I get error do this : `hub pull-request --base gitlab<username>:development --head gitlab<username>:mybranchnow -a gitlab<username>`

the error is something like this: 
[...]

PR #1422, opened 2017-02-23
Add support for reviewers to hub pull-request

ConradIrwin:
Fixes #1358

PR #1453, opened 2017-04-17
Add support for review requests in pull-request command

tomasv:
Fixes https://github.com/github/hub/issues/1358

I took the code from https://github.com/github/hub/pull/1422 and attempted to complete it. I hope @ConradIrwin does not mind. 👼 

I'm not sure what the cleanest way of overriding `Accept` for one request would be. Most options involve adding the `apiPayloadVersion` argument to a lot of function signatures and passing it through. Can this be improved?


ConradIrwin:
Thanks @tomasv !

hkdobrev:
Thanks for this! However, the title is a bit misleading as this adds support for only adding reviewers, not submitting reviews. I think it should be renamed similar to the issue it closes.

mislav:
I've pushed some cleanups to this. Instead of setting a changed `Accept` header per single request, I've simply added the `Accept: application/vnd.github.black-cat-preview+json` value to all requests. This should have no adverse effect to the rest of the API functionality, and is only recognized by API endpoints that deal with reviewer functionality. Let me know what you think!

mislav:
Hmm, I've just tested sending the 2 header values to GitHub API, but it doesn't recognize it:

```
> POST https://api.github.com/repos/mislav/playground/pulls/30/requested_reviewers
> Authorization: token [REDACTED]
> Accept: application/vnd.github.v3+json;charset=utf-8
> Accept: application/vnd.github.black-cat-preview+json
{"reviewers":["mislav"]}

< HTTP 415
{"message":"If you would like to help us test the Pull Request Reviews API during its preview period, you must specify a custom media type in the 'Accept' header. Please see the docs for full details.","documentation_url":"https://developer.github.com/v3"}
```

It's possible that GitHub API doesn't deal well with multiple values of `Accept` header, even though it should in theory. I will follow up to see whether this is a bug in our API.

tomasv:
@mislav is that an accurate representation of header values?

The final raw request header as I understand should look like this: `Accept: application/vnd.github.v3+json;charset=utf-8, application/vnd.github.black-cat-preview+json`. Relevant RFC: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

mislav:
@tomasv Specifying a header twice or specifying it once with two values separated by a comma should be equivalent. But, I'm just checking the implementation of this in the GitHub API, and it looks like we only take the 1st value of `Accept` and discard the rest. I'll check with the API team to see whether this was desired behavior or an oversight.

In the meantime, I'll try switching back to a custom `Accept` header just for this request.

evnm:
Thank you for adding this! Will it be included in 2.3.0?

mislav:
Yes, and I'll try to make a prerelease soon including this.

AlexVKO:
hey, Is it currently working?

mislav:
@AlexVKO You can try this functionality out if you build hub from the master branch. It's not yet released.

PR #1470, opened 2017-05-09
Remove trailing slashes from the $TMPDIR variable

eljobe:
This is helpful for when TMPDIR already has a trailing slash. We can't
be sure that it will on all operating systems, but this will remove it
if it is there.

### DIFF
--- a/script/build
+++ b/script/build
@@ -14,9 +14,8 @@ windows=
 
 setup_gopath() {
   TMPDIR="${LOCALAPPDATA:-$TMPDIR}"
-  MY_TMPDIR=${TMPDIR:-/tmp}
-  # TMPDIR may or may not have had a trailing slash.
-  TMP_GOPATH="${MY_TMPDIR%/}/go"
+  TMPDIR=${TMPDIR:-/tmp}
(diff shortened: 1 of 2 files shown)


script/build
```diff
@@ -14,7 +14,9 @@ windows=
 
 setup_gopath() {
   TMPDIR="${LOCALAPPDATA:-$TMPDIR}"
-  TMP_GOPATH="${TMPDIR:-/tmp}/go"
+  MY_TMPDIR=${TMPDIR:-/tmp}
```
mislav: There is no need to create a new variable `MY_TMPDIR` here. You can just override TMPDIR.

script/build
```diff
@@ -14,7 +14,9 @@ windows=
...
+  # TMPDIR may or may not have had a trailing slash.
```
mislav: There is no need for these code comments. It should be evident that we're stripping a trailing slash.

script/build
eljobe: Done

script/build
eljobe: Removed

PR #1471, opened 2017-05-09
Allow the unsorted expansion for the fish shell

eljobe:
Before this change, the `pu` would sometimes expand to `pull
pull-request push` if the order isn't important, we should take this
fix.

However, if the order is important, we should probably switch the
expectation such that the options are presented alphabetically as
observed.


mislav:
Thanks for the fix!

PR #1472, opened 2017-05-09
Fix bash completion test on MacOS X

eljobe:
However my git installation was run, it installed the
git-completion.{bash,zsh} scripts under
"${GIT_PREFIX}/contrib/completion/" instead of any of the paths that
this test case was expecting.

To make matters worse, the only way to know that this was the problem
was that each scenaro failed with a message about timing out while
waiting for completion to happen in tmux. Just in case there are
other paths out there where the git-distributed completion scripts
may live, I've also added some detection logic to raise an exception
during the "Given" clause of the scenario if we cannot find the
git-distributed bash completion scripts in any of the places we're
looking for them. This would have made it a lot easier to diagonse
the failures.

### DIFF
--- a/features/support/completion.rb
+++ b/features/support/completion.rb
@@ -241,7 +241,7 @@ Given(/^I'm using ((?:zsh|git)-distributed) base git completions$/) do |type|
       end
     end
     if 'bash' == shell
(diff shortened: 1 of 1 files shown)


mislav:
Thanks for this! Sorry for the trouble.

PR #1474, opened 2017-05-09
Allow use of gio instead of xdg-open

### LINKED ISSUE #1473: xdg-open is deprecated under gnome 3.24
`xdg-open` has been deprecated in favor of [`gio`](https://developer.gnome.org/gio/stable/) in GNOME.

```
09:10:12 /home/.../src/comparator:timing
↳ hub browse
This tool has been deprecated, use 'gio open' instead.
See 'gio help open' for more info.

09:10:16 /home/.../src/comparator:timing
↳ [3248:3312:0509/091016.632518:ERROR:browser_gpu_channel_host_factory.cc(103)] Failed to launch GPU process.
Created new window in existing browser session.

09:10:19 /home/.../src/comparator:timing
↳ pacman -Q hub
hub 2.2.9-1
09:10:23 /home/.../src/comparator:timing
↳ pacman -Q gnome-desktop
gnome-desktop 1:3.24.1-1
09:10:29 /home/.../src/comparator:timing
↳ uname -a
Linux moe 4.10.13-1-ARCH #1 SMP PREEMPT Thu Apr 27 12:15:09 CEST 2017 x86_64 GNU/Linux
```

toddself:
Closes #1473.

This is a little messy, but gio requires the `open` argument to it, so you need to pass it back, but the launcher checks the paths, so you can't put `gio open` into the list of browsers, and even if you could, you're stripping the spaces out when you return it, so it would get lost as well.

Happy for feedback/direction on how to make this better.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -27,25 +27,27 @@ func ConcatPaths(paths ...string) string {
 
 func BrowserLauncher() ([]string, error) {
 	browser := os.Getenv("BROWSER")
+	args := ""
 	if browser == "" {
-		browser = searchBrowserLauncher(runtime.GOOS)
+		browser, args = searchBrowserLauncher(runtime.GOOS)
 	}
 
 	if browser == "" {
 		return nil, errors.New("Please set $BROWSER to a web launcher")
 	}
 
-	return strings.Split(browser, " "), nil
+	return strings.Split(browser+" "+args, " "), nil
 }
 
-func searchBrowserLauncher(goos string) (browser string) {
+func searchBrowserLauncher(goos string) (browser string, args string) {
+	args = ""
 	switch goos {
 	case "darwin":
 		browser = "open"
 	case "windows":
 		browser = "cmd /c start"
 	default:
-		candidates := []string{"xdg-open", "cygstart", "x-www-browser", "firefox",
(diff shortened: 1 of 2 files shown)


flying-sheep:
This is a misunderstanding. It's a bug in xdg-open, not some deprecation:

https://bugs.freedesktop.org/show_bug.cgi?id=100862

xdg-open is preferable to gio, as it actually wraps gio and other desktop environments’ equivalents.

Please don't make hub break on KDE, xfce, … by merging this.

mislav:
@flying-sheep So, we don't have to change anything?

flying-sheep:
you don’t.

[xdg-open](https://cgit.freedesktop.org/xdg/xdg-utils/tree/scripts/xdg-open.in) tries to use the most appropriate url/path opening binary for every desktop environment

this includes gfvs-open, gio open, kde-open, and surprisingly many more!

the message “This tool has been deprecated, use 'gio open' instead” is printed by `gfvs-open`, which is tried by `xdg-open` before `gio open`. once the order is fixed, the message will disappear.

toddself:
Closing as this is no longer an issue.

Thanks for the help @flying-sheep and @mislav

PR #1475, opened 2017-05-11
Upgrade to Go 1.8

eljobe:
The only thing that has substansively changed is that Go 1.8 handles
redirect logic more safely than previous versions. This means we can
drop our special handling to avoid following redirects to other
domains. We were only doing that to protect against the possibility
of leaking auth headers. With Go 1.8, the auth headers are not
forwarded when following a redirect to another domain, so we don't
need our special handling any more.

As long as people are attempting to build with the Makefile, the new
check_go_version script should cause the build to stop if our
collaborators aren't using at least version 1.8 of go.


script/build
```diff
@@ -8,6 +8,9 @@
 # With `test`, runs tests instead.
 
 set -e
+source script/check_go_version
+
+check_go_version
```
mislav: Let's just check the Go version in `script/bootstrap` and not here.

script/bootstrap
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 set -e
+source script/check_go_version
```
mislav: I think the whole external script approach might be a bit of an overkill. How about:

```sh
go_ver="$(go version | grep -o '1.[[:digit:]]\+')"
if [ "${go_ver#*.}" -lt 8 ]; then
  echo "You need to install Go 1.8 or higher to build hub" >&2
  STATUS=1
fi
```

script/build
eljobe: We can. But ...

`script/bootstrap` doesn't run if the user already has bin/cucumber in their working directory.

The scenario I'm trying to avoid is this:
1. Developer builds with go1.6 using `make test-all`
2. Developer does a `git pull`
3. Developer does another `make test-all`

At this point, the tests all pass, but the developer has a version of hub that is susceptible to following redirects and forwarding auth headers. If they use it, they're open to a security risk, and they had no way to know that they shouldn't have been building with an older version of go.

script/bootstrap
eljobe: I'm not a fan of this because it isn't very future-proof (or past-proof, for that matter.)

If someone has a version of Go <1.0 (say, 0.7) then they will get a somewhat confusing error when running this script:
`-bash: [: : integer expression expected`

Same is true for future versions >=2.0.

I'm not committed to having it in a separate script (unless you agree that we need to handle the scenario I mentioned above), but it would be nice to keep the more permissive REGEX.

script/build
eljobe: I've made this change. We can always make it more robust if someone complains.

script/bootstrap
eljobe: I made the change you requested. We can make it more robust if someone complains. Also, I should mention that this change causes `script bootstrap` to no longer print the version of go being used (which it used to do.)  Is that okay, or should I add it back in?

mislav:
Thank you! I followed your advice and enforced a version check at build time too.

PR #1496, opened 2017-06-20
reset-console

### LINKED ISSUE #1481: Reading in the pull request message file from stdin and editing it messes up the console
Running the following command on hub version 2.3.0-pre9 on centos 7
Causing the terminal to become unresponsive (Which can be fixed by typing reset)

```
git diff --no-color -M --patch-with-stat  master...HEAD | perl -pi -e's/^/#/' | hub pull-request -F - --edit
```

The solution seems to be to close stdin and reopen the tty.
I have a working example in c.
I am not sure how this should be done in go.

```c
#include <fcntl.h>
#include <stdio.h>
#include <unistd.h>
#include <errno.h>

int main(void)
{
  char buf[BUFSIZ];
  int pid, status;
  char * vim_argv[] = {"/usr/bin/vim", NULL};
  int error;

  puts("Reading from stdin...");
//  read all the stdin
  while(fgets(buf, sizeof(buf), stdin));

// Reopen stdin using the tty
  close(0);
  open("/dev/tty", O_RDONLY);

  pid = fork();
  if (pid == -1) {
    printf("Error forking\n");
    return 1;
  }
  if (pid == 0) {
    error = execv("/usr/bin/vim", vim_argv);
    perror("exec vim");
    return errno;
  } else {
      waitpid(pid, &status, 0);
  }
  return 0;
}
```

rouzier:
This pull request resets stdin to be read from the current console on unix based system.

To avoid the vim error 'Vim: Warning: Input is not from a terminal'

When reading in the pull request message from stdin and editing it afterwards.

For example the following command will no longer causing issues for a terminal (Which is the command I used to create this pull request)

```
(git rev-parse --abbrev-ref HEAD 2>/dev/null;(git diff --no-color -M --patch-with-stat  github/master...HEAD | perl -pi -e's/^/#/') )  | hub pull-request -b github:master -F - --edit
```

Fixes #1481

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -14,8 +14,11 @@ import (
 )
 
 type Cmd struct {
-	Name string
-	Args []string
+	Name   string
+	Args   []string
+	Stdin  *os.File
+	Stdout *os.File
+	Stderr *os.File
 }
 
 func (cmd Cmd) String() string {
@@ -63,9 +66,9 @@ func (cmd *Cmd) Run() error {
 func (cmd *Cmd) Spawn() error {
 	verboseLog(cmd)
 	c := exec.Command(cmd.Name, cmd.Args...)
-	c.Stdin = os.Stdin
-	c.Stdout = os.Stdout
-	c.Stderr = os.Stderr
+	c.Stdin = cmd.Stdin
+	c.Stdout = cmd.Stdout
+	c.Stderr = cmd.Stderr
 
 	return c.Run()
 }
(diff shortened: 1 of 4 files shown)


mislav:
Interesting approach; thank you for trying to solve this!

I'm wondering whether there might be a potentially less intrusive way, i.e. one that doesn't force us to reset anything for the curent `hub` process.

The root of the problem is that the child `vim` process inherits the stdin from its parent `hub` process, which is undesireable in this case. Can't we simply change the stdin for the child `vim` process to be connected to the original terminal? Something like:

```go
editCmd.Stdin = ...
editCmd.Spawn()
```

rouzier:
Let me try that approach.

mislav:
This looks great. I wanted to ask what would happen if tty was unavailable in the first place, for example in environments was called from a cron job or within a process that doesn't have a terminal, but opening `vim` wouldn't work or make sense in that case anyway, so I think we're safe like this.

Have you tried that this fixes the problem for you?

rouzier:
Yes I have and it work perfectly.

mislav:
Thank you for the hard work!

PR #1497, opened 2017-06-23
Sort hub commands

joshbetz:
The rest of the sections are sorted alphabetically. This makes the hub commands easier to read by sorting them the same way.


mislav:
This makes sense; thanks!

Issue #1451, opened 2017-04-12
hub pull-request for existing pull request


sfahlberg:
I'm wondering if there's a way for me to open up an existing pull request using hub. I can go to the branch page using `hub browse` or `hub compare` but I want to be able to see my existing PR when  I run `hub pull-request`and it detects a pr, it would open that page up in a browser for the developer to inspect

mislav:
Agreed! This is a duplicate of #897.

Issue #897, opened 2015-05-11
New command/option to open an existing pull request


blueyed:
I would like to have a command to open/browse an existing pull request.

Since `pull-request` is able to see if there's a pull-request already (at least when it tries to create one), this should be possible.

It should default to the current branch with its tracking remote/repo.

Related issue: https://github.com/github/hub/issues/880 - which is about displaying the URL to a pull request.

parkr:
`hub browse` gets you most of the way there.

blueyed:
Yes, it helps in general.

But given https://github.com/github/hub/issues/896 you won't be on your fork / the branch necessarily.

(I assume that you mean that the branch is opened and you can click the link to the PR?!)

dwijnand:
```bash
git config --global alias.pr-open \
  '!open $(hub pr list -f %U -h dwijnand:$(git rev-parse --abbrev-ref HEAD))'
```

to add a `pr-open` alias to your `~/.gitconfig`

Replace **dwijnand** with your username, and `open` if not on macOS.

jasonkuhrt:
Suggest command name/structure:

```
$ git pr browse <#>
```

mislav:
`hub pr show` has now landed in master https://github.com/github/hub/pull/2141

dentuzhik:
This is amazing, thank you 🥇 
That time when issue hangs in your subscription for more than two years and then it finally pays off 💃

mislav:
@dentuzhik You're welcome! Credit goes to @ashemesh @hmemcpy who eventually made contributions towards the feature ❤️

hmemcpy:
Thanks very much for pushing it through! 🎉

Issue #867, opened 2015-04-14
"A pull request already exists" error should show the url


wbolster:
When running `hub pull-request`, the "A pull request already exists for ..." error message should also show the URL, just like a successful attempt would do.

mislav:
Good idea!

We'll need to make an extra HTTP request to fetch the existing PR ID, but it might be worth it. Will keep for consideration.

parkr:
I would absolutely love this. We do a form of PR deploys instead of branch deploys (just naming semantics). It'd be nice to be able to run `hub pull-request` and get back either a URL I can parse or just the PR number that I can then use. Any of the following would be excellent:

```text
fatal: pull request for branch `my-feature` already exists: #1652
```

```text
https://github.com/org/repo/pull/1652
```

(this is nice because I can do `hub pull-request | xargs open`)

```text
Your pull request already exists: https://github.com/org/repo/pull/1652
```

sigginet:
+1
I'm automating github pull requests and merges and this would be just a beautiful addition...
It would allow me to deploy fully without any manual interaction (saving me and other project maintainers so many precious hours each month)
If there is enough interest, I'll pull up the go documentation and write the thing myself. (just do +1 and tag me)

sigginet:
Alternatively, if you could see the branch names from `hub issue` or `git issue` then that would be a super-awesome feature.
We could check the exit status of `hub pull-request`, then based on that, see if a PR exists or not and take the appropriate action

muescha:
i would like to see more an:

`hub pull-request --check` check if an exisiting pull request exists and display the id and/or name

PR #1498, opened 2017-06-25
[WIP] Adding 'git pull-request —show' to list the URL(s) for existing PRs

hmemcpy:
Adding a new switch to the `pull-request` command to show the URL, if the PR already exists (helps solve #897, #1107 and #1451)
This is a prerequisite for other features, such as #867

Keeping WIP because I want to add tests for this.

### DIFF
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -91,6 +91,7 @@ var (
 	flagPullRequestCopy,
 	flagPullRequestEdit,
 	flagPullRequestPush,
+	flagPullRequestShow,
 	flagPullRequestForce bool
 
 	flagPullRequestMilestone uint64
@@ -109,6 +110,7 @@ func init() {
 	cmdPullRequest.Flag.StringVarP(&flagPullRequestMessage, "message", "m", "", "MESSAGE")
 	cmdPullRequest.Flag.BoolVarP(&flagPullRequestEdit, "edit", "e", false, "EDIT")
 	cmdPullRequest.Flag.BoolVarP(&flagPullRequestPush, "push", "p", false, "PUSH")
+	cmdPullRequest.Flag.BoolVarP(&flagPullRequestShow, "show", "s", false, "SHOW")
 	cmdPullRequest.Flag.BoolVarP(&flagPullRequestForce, "force", "f", false, "FORCE")
 	cmdPullRequest.Flag.StringVarP(&flagPullRequestFile, "file", "F", "", "FILE")
 	cmdPullRequest.Flag.VarP(&flagPullRequestAssignees, "assign", "a", "USERS")
@@ -183,6 +185,18 @@ func pullRequest(cmd *Command, args *Args) {
 		} else {
 			head = trackedBranch.ShortName()
 		}
+	}
+
+	if flagPullRequestShow {
+		pullRequests, err := client.GetPullRequestUrlsForHead(headProject, head)
+		utils.Check(err)
+
+		for _, pr := range pullRequests {
+			ui.Println(pr.HtmlUrl)
+		}
+
+		args.NoForward()
+		return
 	}
 
 	if headRepo, err := client.Repository(headProject); err == nil {
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -179,6 +179,7 @@ func pullRequest(cmd *Command, args *Args) {
 		}
 	}
 
+	filterParams := map[string]interface{}{}
 	if head == "" {
 		if trackedBranch == nil {
 			head = currentBranch.ShortName()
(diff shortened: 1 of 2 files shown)


commands/pull_request.go
```diff
@@ -185,6 +187,18 @@ func pullRequest(cmd *Command, args *Args) {
 		}
 	}
 
+	if flagPullRequestShow {
+		pullRequests, err := client.GetPullRequestUrlsForHead(headProject, head)
+		utils.Check(err)
+
+		for _, pr := range pullRequests {
+			ui.Println(pr.HtmlUrl)
```
mislav: This prints all results found.  There can only ever be one open PR for a given branch, so if there is more than 1 result, the rest will be closed PRs. Some questions:

1. Do we want to visually indicate open/closed status?
2. How will we order the results?
3. What about past PRs for whom is the branch name is identical to the current branch name, but they were completely different branches at different times that happened to share the same name? Will we try to filter those out from results?
4. Do we want to consider only showing the open PR, or nothing if there are no open PRs?

github/client.go
```diff
@@ -134,6 +134,23 @@ func (client *Client) CreatePullRequest(project *Project, params map[string]inte
 	return
 }
 
+func (client *Client) GetPullRequestUrlsForHead(headProject *Project, branch string) (pr []PullRequest, err error) {
```
mislav: Could we have a generic `FetchPullRequests` method that accepts an arbitrary dictionary of query parameters, not just `head`, so we can use it for more purposes in the future? See `FetchIssues`.

github/client.go
```diff
@@ -134,6 +134,23 @@ func (client *Client) CreatePullRequest(project *Project, params map[string]inte
...
+	api, err := client.simpleApi()
+	if err != nil {
+		return
+	}
+
+	res, err := api.Get(fmt.Sprintf("repos/%s/%s/pulls?head=%s:%s", headProject.Owner, headProject.Name, headProject.Owner, branch))
```
mislav: The `owner` and `repo` values in `repos/:owner/:repo/...` URLs should not be the head project. Instead, it should be the "base" project (which is considered to be the "canonical" version of the project at which issues and PRs are opened).

hmemcpy:
@mislav Thank you very much for your comments.

I initially tried to add this to `pr`, but quickly realized I needed a lot of the same code that `pull-request` already did (like, getting the `headProject`, etc). Also, I think `pr` and `pull-request` are conceptually different - `pr` is for checking out other people's PRs, but `pull-request` is for creating/working with your own. So it felt more natural adding it there (and I didn't need to duplicate/move code to do it!)

mislav:
@hmemcpy I understand that you want to avoid duplicating a lot of the logic between `pr` and `pull-request`. That's why we could extract some of that logic to functions that can be invoked from either of the commands.

`hub pull-request` is a command that existed for a long time and was only ever meant to be for creating pull requests. However, now that we want to add additional functionality for PRs to hub, I thought that creating a separate command  with a shorter name (`hub pr`) would be better so we can add other subcommands to it (like checking out a PR based on its number, listing existing PRs, filtering PRs based on some criteria, showing the PR for the current branch, etc). That's why I suggested that this new functionality lives under the `hub pr` command rather than overloading `hub pull-request`. Let me know if this makes sense to you, and what your thoughts about this API are.

hmemcpy:
@mislav hey! Sorry for disappearing, your comments make a lot of sense. I'll try to do some refactoring to extract the common bits so I can reuse them from different commands. Regarding `pr` vs `pull-request`, you're right, of course. I only wanted to minimize the amount of changes for submitting this feature.
But I'll look into it. Golang seems friendly enough so even I can do it :))

github/client.go
hmemcpy: I tried using "base" initially, but I couldn't get a valid response, I even contacted github support. Their response was:

> You'll want to specify the head branch, not the base branch, like this:
> https://api.github.com/repos/fsharp/FAKE/pulls?head=fsharp:FSI
> The head is the source of the pull request (the branch from which the pull request is made), and the base is the target of the pull request (the branch to which the pull request is made).

Hope this makes sense...

czj:
Hi ! First, thanks a lot for you combined great work :)

Enhancing the functionality of `hub pr` would be really useful for our workflow. Both `hub pr show` and `hub pr open` would fit nicely in our scripts as they serve different purpose & usages.

acaloiaro:
Hey @hmemcp -- I'm curious if you're still interested in completing this feature and able to do so. It's something I'd love to see in `hub`.

I'd be happy to pick it up if you don't think you'll be able to.

mislav:
`hub pr show` has now landed in master https://github.com/github/hub/pull/2141

Issue #718, opened 2014-12-02
Debian/Ubuntu packages


dbarnett:
Would it be possible to provide deb packages to easily install hub on Ubuntu and Debian?

gsuess:
@mislav ubuntu and debian support more than one architecture, so one pre-compiled binary will not be enough unless for some reason you want to only support x86 or x64. However cross-compiling go shouldn't be that hard.

The SO answer mentioned here tells you how to generate a .deb package. This is a package file that contains the pre-compiled binaries and a list of dependencies.

Once you have a .deb for each architecture and ubuntu/debian version that you wish to support you can immediately distribute  it through a PPA as has been mentioned by dbarnett. I guess this answer may be helpful: http://askubuntu.com/a/71516/49920

You can then request it to be posted to the debian repositories:
https://wiki.debian.org/DebianMentorsFaq#How_do_I_make_my_first_package.3F.

lazyfrosch:
There is an ITP for a similar tool on Debian:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=763345

Mte90:
+1

thedrow:
+1

mpolakis:
+1

rroblak:
@sometimesfood created a [`.deb` PPA for hub](https://launchpad.net/~sometimesfood/+archive/ubuntu/hub/+packages) a while back. I wonder if he'd be willing to update the package version (it's old at 1.10.1) or allow someone else to do so?

sometimesfood:
@rroblak My packaging script uses Debian's Ruby-specific packaging tool chain, so it will probably not be too useful.

Still, if there is enough interest, I can look into creating a new Debian package and updating the PPA. Creating packages is not too difficult but it can be quite intimidating for a beginner (_lots_ of tools, lots of policies). Also, to make matters worse, the Debian packaging tools and docs for golang seem to be rather new.

sometimesfood:
Btw, just for reference: Michael Stapelberg has compiled some nice examples on how to package a simple Go application for Debian:
http://anonscm.debian.org/cgit/collab-maint/dh-golang.git/tree/example

Also, the debian/rules and debian/control files of this package seem to be a nice reference:
https://packages.debian.org/sid/codesearch

alfredocdmiranda:
Hey guys, I just created two .deb packages (amd64 and i386). I've tested amd64 and it is working fine. It's missing the completion scripts because I didn't know where to put them.

[...]

PR #1502, opened 2017-07-04
Update README.md with Debian installation instructions

mslinn:


mislav:
Sorry, but this references a ppa maintained by someone else (`cpick`) who is not a maintainer of this project. Should cpick stop updating the ppa in the future, these instructions would become obsolete.

I would rather create a ppa under our control and keep it up to date after every official release. I've asked how to do that in https://github.com/github/hub/issues/718#issuecomment-245729552 and if anyone has ideas how to set that up, please write there. Thanks!

cpick:
@mislav for what it's worth, I recorded the steps I take every time I update my PPA here:
https://github.com/github/hub/issues/718#issuecomment-292311758

mslinn:
I see my earlier comment has been deleted/censored. Uncool.

mislav:
@mslinn I don't think anyone would delete your comment. I didn't receive any email notification for a comment that was later deleted. Was it in this thread, and if so, was it the text of the originally opened PR that is now gone, or a comment you've submitted later on?

@cpick Yes thank you for that. I will study that, but for now, the problem might be that I can't automate that via CI, since I can't spawn virtual machines via Vagrant within Travis CI environment. However, I could create a setup for publishing those PPA from my own machine.

cpick:
@mislav I only use Vagrant to start from a fresh install and ensure I haven't forgotten to list some dependency.  Really the process could skip all the Vagrant steps and still work just the same.  (There are also other tools that attempt to accomplish the same thing, eg `pbuilder`, I'm just less familiar with them.)  Most CI environments provide enough isolation between runs that they likely don't need any additional sandboxing.

Issue #1504, opened 2017-07-08
Windows Go install


ofek:
I'm not too familiar with Go's tooling. When I do `go get github.com/github/hub` it says `go\src\github.com\github\hub\commands\apply.go:8:2: found packages github (branch.go) and commands (reset_console_windows.go) in C:\Users\Ofek\go\src\github.com\github\hub\github`.

kitsuphat2528:
Ok

ofek:
Fixed by #1505

PR #1505, opened 2017-07-08
Fix windows build

pschultz:


pschultz:
Fixes #1504

pschultz:
Consider adding `GOOS=windows go build` to .travis.yml to catch this in the future.

ofek:
I can confirm this now works on Windows :+1:

mislav:
Thank you!!

Issue #1514, opened 2017-07-19
Build fails on linux when install has windows line endings.


HarrySarson:
I am running Bash on Windows and cloned the hub repo with windows line endings as I stored it in the windows part of my hard drive.

When I ran `make` the script failed with an non intuitive error message which I tracked down to the presence of `\r`'s in the ./script/build file (and also in other files).

Running `dos2unix` solved the problem but I thought it worth flagging up as I don't think this is a work flow unique to me.

It may be worth adding a catch for this and producing a more helpful error message.

**stack trace:**

```
$ make install prefix=/usr/local                                          
script/build: line 9: $'\r': command not found                            
: invalid optionne 10: set: -                                             
set: usage: set [-abefhkmnptuvxBCHP] [-o option-name] [--] [arg ...]      
script/build: line 11: $'\r': command not found                           
script/build: line 14: $'\r': command not found                           
script/build: line 15: syntax error near unexpected token `$'{\r''        
'cript/build: line 15: `setup_gopath() {                                  
script/build -o bin/hub                                                   
script/build: line 9: $'\r': command not found                            
: invalid optionne 10: set: -                                             
set: usage: set [-abefhkmnptuvxBCHP] [-o option-name] [--] [arg ...]      
script/build: line 11: $'\r': command not found                           
script/build: line 14: $'\r': command not found                           
script/build: line 15: syntax error near unexpected token `$'{\r''        
'cript/build: line 15: `setup_gopath() {                                  
make: *** [bin/hub] Error 2                                               

```

mislav:
This has some up before. However, I don't know how to fix it. Bash scripts shouldn't have windows line endings. Do you have any suggestion how to detect this?

HarrySarson:
hmm, I had a similar issue in another repository and did some more digging.

You can overwrite the users `core.autocrlf` settings for a repository in the `.gitattribute` file. Moreover you can do so by matching against patterns. Therefore, all the `*.sh` files could be set to use the linux line endings regardless of the users global settings.

I can't conceive of any user expecting to have windows line endings in a bash script file and it would allow the repository to build when people - like me - clone with silly settings.

mislav:
Would you submit a PR that adds a `.gitattribute` file instead of changing our build scripts?

HarrySarson:
Yep, can do :) will be Tuesday evening probably.

mislav:
Thank you for https://github.com/github/hub/pull/1557!

yubing24:
[...]

PR #1516, opened 2017-07-21
Nice error message if windows line endings are present in bash script

HarrySarson:
I have found a way to make a nice error message when `./script/build` is run on a linux machine when there are windows line endings present.

The script reads the first character and checks if it is `\r` or not.

Not very pretty I will admit although it can be tidied up somewhat.

To address #1514


HarrySarson:
To test it you can add `\r` to file-with-newline. On an install where all the files have `\r`'s in then the build will still crash horribly but not before printing the help full error message.

mislav:
Thanks for trying to improve the error message. I'm still not perfectly happy with the approach, especially because it adds a blank file to the project just to test for line endings. Perhaps a `.gitattributes` file in our project's root would be a better approach?

HarrySarson:
I guess you could search in any file for the first instance of `\n` and then check if the previous character is `\r`

PR #1519, opened 2017-07-28
Fix editor behavior when core.commentChar is set to auto

### LINKED ISSUE #1512: git's commentChar=auto option is not handled well
Here's a comment generated by running `hub pull-request`:

image

e-beach:
Fixes #1512.

I didn't write a test for whether the feature works for releases; all that's changed is using `#` rather than `auto`, which doesn't seem very important. Other changes are tested.

Never coded in Go before, please tell me if I screwed anything up :smiley:.

### DIFF
--- a/commands/issue.go
+++ b/commands/issue.go
@@ -367,7 +367,7 @@ func createIssue(cmd *Command, args *Args) {
 		title, body, editor, err = readMsgFromFile(flagIssueFile, flagIssueEdit, "ISSUE", "issue")
 		utils.Check(err)
 	} else {
-		cs := git.CommentChar()
+		cs := git.CommentCharNoAuto()
 		message := strings.Replace(fmt.Sprintf(`
 # Creating an issue for %s
 #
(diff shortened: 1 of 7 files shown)


e-beach:
Closed as I need to examine behavior some more.

mislav:
Hi @e-beach thank you for proposing this PR. I had reviewed it over the weekend, but the implementation didn't sit well with me. You must have thought the same because you closed this PR.

I have an idea for another approach, and I'm going to try to implement it myself. It will require some shuffling of existing code around, since currently the code that sets up the text editor is a bit of a mess.

e-beach:
The behavior of this PR is broken in several ways, and, as a result, after this PR but before your response I had opened #1522, which includes some logical corrections. It probably does not resolve the root design issues you mention, so I'll close it as well.

mareksuscak:
Hey, @mislav, did you make any progress on this? If not, maybe share your thoughts with us and someone could contribute.

PR #1522, opened 2017-07-28
Fix comments when commentChar=auto

### LINKED ISSUE #1512: git's commentChar=auto option is not handled well

e-beach:
Fixes #1512.

I didn't write a test for whether the feature works for releases; all that's changed is using # rather than auto, which doesn't seem very important.

There is also the extreme possibility that every candidate comment character is already in the message. In this case, i just use "auto", but I'm not sure if this is right.

I've not coded in Go before, so there might be issues with code organization?


e-beach:
See https://github.com/github/hub/pull/1522

e-beach:
See #1519

PR #1524, opened 2017-07-28
added wildcard cloning

atrakh:
Hello,
This change implements wildcard cloning of all public repos (no authentication implemented) associated with a user or organization. Note that this is not a regex wildcard, but a hardcoded asterisk.

Usage: `$ git clone atrakh/*`

I understand the implementation is not optimal. I'm mostly interested in knowing if this pattern would be a viable addition to hub. :smile:

### DIFF
--- a/commands/clone.go
+++ b/commands/clone.go
@@ -5,8 +5,10 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/github/hub/cmd"
 	"github.com/github/hub/github"
 	"github.com/github/hub/utils"
+	"github.com/octokit/go-octokit/octokit"
 )
 
 var cmdClone = &Command{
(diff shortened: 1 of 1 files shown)


e-beach:
Although I'm not the maintainer , I think a valid question is "Why would you want to do this"?

mislav:
Thank you for this contribution, but I don't think this is a useful addition to the hub feature set. Most users of hub will never need to clone every repository under a specific username or organization.

Hub might gain features in the future to allow shell scripts access to the data returned from GitHub API. Then, you will be able to write a bash script that queries the names of all repos under an org, for example, and issue `hub clone` for each of them. However, the feature of wildcard cloning will not be implemented in hub directly.

Issue #1499, opened 2017-06-26
`hub fork` fails if fork (and remote) already exists


jasonkarns:
While attempting to replicate the issue I _thought_ I was seeing for https://github.com/github/hub/issues/1411#issuecomment-311148349, I uncovered a teeny quirk with how `hub fork` behaves for existing forks.

- If fork does not exist: `hub fork` creates the fork on github, adds the remote for the new fork, and fetches. Returns 0 status.

- If fork exists but remote doesn't: `hub fork` adds the remote for the fork and fetches. Returns 0 status.

- If fork exists and remote exists: `hub fork` errors out with `fatal: remote jasonkarns already exists`

Given that forking an already-forked repo safely adds the fork as a remote, I'd _kinda_ expect the last scenario to just trigger a fetch. That is, I'd expect all three scenarios above to result in a local clone with: 

1. named remote for the fork
3. remote recently fetched
4. successful return code

Right now, only the first two scenarios do this. (The third just needs to do a fetch and exit cleanly.)

I have a hard time describing the second scenario above (fork exists, but remote doesn't), but it's like a partially-idempotent command. If the third scenario were implemented as requested, the `hub fork` command would be fully idempotent.

mislav:
What you're describing makes sense. I'd accept a PR that makes the command idempotent, since it's already half-idempotent.

e-beach:
It's odd that version 2.3.10 of hub behaves slightly differently than the release version. In this case, the pre-release version does not show an error message, but does exit with an error status.
```
# both fork and remote already exist

$ hub-2.3.10 fork
$ echo $?
128
$ hub-2.2.9 fork
fatal: remote e-beach already exists.
$ echo $?
128
```

e-beach:
The change in behavior occurs in d937d287037b9a24022e16337ea5e06ffdc726ab. It looks like the change to console file descriptors in that commit is somehow broken.

jasonkarns:
woot!, thanks @e-beach

PR #1533, opened 2017-08-04
Pass file descriptors to cmd.NewWithArray

e-beach:
This fixes a change in the behavior of hub console output introduced in d937d287037b9a24022e16337ea5e06ffdc726a.

That commit changed the way file descriptors are passed around so that commands created by `cmd.NewWithArray` have nil stdin, stdout, and stderr. I believe not passing file descriptors to the function was an oversight.

 An example of a resultant change in behavior is that the `hub fork` command will error-out with code 128 when the forked repo and remote already exist, but will print no error message. Before d937d287037b9a24022e16337ea5e06ffdc726a, `hub fork` prints the normal error message in this case. I noticed this from the discussion in #1499.

Am I correct in thinking that this is a bug?

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -105,7 +105,7 @@ func New(cmd string) *Cmd {
 }
 
 func NewWithArray(cmd []string) *Cmd {
-	return &Cmd{Name: cmd[0], Args: cmd[1:]}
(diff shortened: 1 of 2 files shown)


mislav:
Thank you! Really great catch.

e-beach:
This probably fixes quite a few bugs in the `2.3.0-pre10` version of hub. I was just trying to figure out why `hub pr checkout` failed silently for me in an edge case, and it was because of this bug.

PR #1535, opened 2017-08-11
Make fork command idempotent

### LINKED ISSUE #1499: `hub fork` fails if fork (and remote) already exists

e-beach:
Fixes #1499.

This PR makes `hub fork` succeed, with no operation, if a remote pointing to the forked repository already exists.

### DIFF
--- a/commands/fork.go
+++ b/commands/fork.go
@@ -6,7 +6,6 @@ import (
 	"github.com/github/hub/github"
 	"github.com/github/hub/ui"
 	"github.com/github/hub/utils"
-	"github.com/github/hub/git"
 )
 
 var cmdFork = &Command{
@@ -109,11 +108,13 @@ func fork(cmd *Command, args *Args) {
 
 		// Check to see if the remote already exists and points to the fork.
 		currentRemote, err := localRepo.RemoteByName(newRemoteName)
-		if (err == nil) {
-
-			parsedURL, err := git.ParseURL(url)
-			if err == nil && parsedURL.String() == currentRemote.URL.String(){
-				return // Already forked, nothing to do.
+		if err == nil {
+			currentProject, err := currentRemote.Project()
+			if err == nil {
+				if currentProject.SameAs(forkProject){
+					ui.Printf("existing remote: %s\n", newRemoteName)
+					return
+				}
 			}
 		}
 
--- a/commands/fork.go
+++ b/commands/fork.go
@@ -6,6 +6,8 @@ import (
 	"github.com/github/hub/github"
 	"github.com/github/hub/ui"
 	"github.com/github/hub/utils"
+	"github.com/github/hub/git"
+
 )
 
 var cmdFork = &Command{
@@ -49,6 +51,7 @@ func init() {
 
 	CmdRunner.Use(cmdFork)
 }
+
 
 func fork(cmd *Command, args *Args) {
 	localRepo, err := github.LocalRepo()
@@ -102,8 +105,20 @@ func fork(cmd *Command, args *Args) {
 
 	args.NoForward()
 	if !flagForkNoRemote {
+
 		originURL := originRemote.URL.String()
 		url := forkProject.GitURL("", "", true)
+
+		// Check to see if the remote already exists and points to the fork.
+		currentRemote, err := localRepo.RemoteByName(newRemoteName)
+		if (err == nil) {
+
+			parsedURL, err := git.ParseURL(url)
+			if err == nil && parsedURL.String() == currentRemote.URL.String(){
+				return // Already forked, nothing to do.
+			}
+		}
+
 		args.Before("git", "remote", "add", "-f", newRemoteName, originURL)
 		args.Before("git", "remote", "set-url", newRemoteName, url)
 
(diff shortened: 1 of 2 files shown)


features/fork.feature
```diff
@@ -164,6 +164,22 @@ Scenario: Related fork already exists
      Then the exit status should be 0
      And the url for "mislav" should be "git@github.com:mislav/dotfiles.git"
 
+   Scenario: Related fork and related remote, but with differing protocol, already exist
+       Given the "mislav" remote has url "https://github.com/mislav/dotfiles.git"
```
e-beach: If the remote was created using https protocol, then `hub fork` could fail even if the remote already exists pointing to the correct forked repo. This could be corrected with some more commits converting https:// github urls to ssh equivalents, which I believe be very simple.

commands/fork.go
```diff
@@ -102,8 +103,20 @@ func fork(cmd *Command, args *Args) {
...
+
 		originURL := originRemote.URL.String()
 		url := forkProject.GitURL("", "", true)
+
+		// Check to see if the remote already exists and points to the fork.
+		currentRemote, err := localRepo.RemoteByName(newRemoteName)
+		if (err == nil) {
+
+			parsedURL, err := git.ParseURL(url)
+			if err == nil && parsedURL.String() == currentRemote.URL.String(){
```
mislav: Rather than comparing string representations of two urls, let's rather construct Project instances using `remote.Project()` and use `project.SameAs(anotherProject)` to compare them. This is going to detect whether a git remote already exists for this project, but ignore the differences in protocol, which I think is desireable in this case.

For example, if a remote was going to be added for `https://github.com/foo/bar.git`, but one already exists for `git@github.com:foo/bar.git` or even `https://github.com/foo/bar` (no `.git` suffix), those differences are irrelevant and should be ignored.

features/fork.feature
```diff
@@ -152,6 +152,33 @@ Scenario: Related fork already exists
       """
     And the url for "mislav" should be "git@github.com:mislav/unrelated.git"
 
+ Scenario: Related fork and related remote already exist
+     Given the "mislav" remote has url "git@github.com:mislav/dotfiles.git"
+     Given the GitHub API server:
+     """
+       get('/repos/mislav/dotfiles') {
+             json :parent => { :html_url => 'https://github.com/EvilChelu/Dotfiles' }
+           }
```
mislav: These lines seem to be indented one level too deep

features/fork.feature
```diff
@@ -152,6 +152,33 @@ Scenario: Related fork already exists
...
+  Scenario: Related fork and related remote, but with differing protocol, already exist
+      Given the "mislav" remote has url "https://github.com/mislav/dotfiles.git"
+      Given the GitHub API server:
+      """
+        get('/repos/mislav/dotfiles') {
+              json :parent => { :html_url => 'https://github.com/EvilChelu/Dotfiles' }
+            }
+      """
+      When I run `hub fork`
+      Then the exit status should be 128
```
mislav: As I've stated in above comment, I think this scenario should exit with success status and keep the `mislav` remote intact. The difference in protocol shouldn't matter (the orignal protocol of the existing `mislav` remote should "win").

mislav:
I wrote you [this on Twitter](https://twitter.com/mislav/status/895984530024169472) but it has only now ocurred to me that you might not be checking it regularly. So: thanks for all the help!

features/fork.feature
e-beach: Thanks; this test exists because I anticipated there was a better way of doing things, so I'm glad to change the behavior.

e-beach:
Alright, I adjusted my code to print the remote name. I assume the breaking change is acceptable given the benefit of simplicity.

 With the current behavior, there is no reason `bump-formula-pr.rb` should have to call `hub` three times, correct? Each invocation will print the name of the remote, so its not clear to me that script is written the way it is. I can also do a PR on homebrew to make it work with the change.

mislav:
Thanks for your work on this! I will open a PR for `bump-formula-pr` and CC you on it for 👀 .

PR #1537, opened 2017-08-14
Document the availability of the development version.

mareksuscak:
I checked the [Chocolatey Gallery](https://chocolatey.org/packages/hub) and it doesn't seem like there's a prerelease of `hub`. So I haven't documented that but Chocolatey itself [supports prereleases](https://chocolatey.github.io/usage.html#installing-packages).

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,8 @@ git version 1.7.6
 hub version 2.2.3
 ```
 
-if you want to use the latest unstable version, use:
+If you want to get access to new `hub` features earlier and help with its
+development by reporting bugs, you can install the prerelease version:
 
 ``` sh
 $ brew install --devel hub
--- a/README.md
+++ b/README.md
@@ -33,6 +33,12 @@ git version 1.7.6
 hub version 2.2.3
 ```
 
+if you want to use the latest unstable version, use:
+
+``` sh
+$ brew install --devel hub
+```
+
 #### Chocolatey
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,6 +33,12 @@ git version 1.7.6
 hub version 2.2.3
 ```
 
+if you want to use the latest unstable version, use:
```
mislav: 1. Please start sentences with a capital letter. That would be "If".
2. A prerelease version is not necessarily "unstable". They pass all CI tests. It's just that their feature set changes often and some flags might be removed or renamed.

     Also, nobody really wants to "use the latest unstable version". If you phrase it like that, that sounds like there is no benefit to using `--devel`, and that there is only a negative consequence: instability. But, the actual benefit is that you gain access to new features quicker. So, let's instead phrase it using a language that reveals a benefit that a user might get from using the `--devel` flag: “If you want to get access to new hub features earlier and help report bugs, you can install the prerelease version”.

mareksuscak: An excellent feedback, thank you. I'll incorporate the language. 

PS: I considered phrasing it the same way as you've just proposed but wasn't really sure if that is true so thanks for clearing that up.

mareksuscak:
This is now ready to be re-reviewed @mislav.

mislav:
Thank you!

PR #1542, opened 2017-08-17
Check modtime of message file after it is edited by the user

### LINKED ISSUE #1428: UX: quitting the editor without editing the default commit message template creates a pull request anyway
I was creating a pull request for Ansible by running hub pull-request. It downloaded Ansible's PULL_REQUEST_TEMPLATE.md very helpfully and spawned Vim for me.  I remembered about issue #1377 and decide I didn't want `hub` to create a malformed PR for me, I'd rather create one by hand on the web.  So I quit vim with `:q!`.

`hub` created a pull request anyway.

e-beach:
Fixes #1428.

This change would affect the behavior of the editor for issues, releases, and pull requests by setting the modified time of the message file 5 seconds back before it is edited, then verifying that the time-stamp changed.

Some existing tests had to be modified to reflect that the editor must at least `touch` the message file.

### DIFF
--- a/features/pull_request.feature
+++ b/features/pull_request.feature
@@ -94,7 +94,7 @@ Feature: hub pull-request
     Then the output should contain exactly "the://url\n"
 
   Scenario: Single-commit with pull request template
-    Given the git commit editor is "true"
+    Given the git commit editor is "touch"
     Given the GitHub API server:
       """
       post('/repos/mislav/coral/pulls') {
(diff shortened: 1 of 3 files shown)


e-beach:
Yes, I see in that case a non-edited message is the same as a blank message, and we should treat it the same as a blank message.

e-beach:
It seems it still needs a white-space touch-up, but it looks correct in my editor.

e-beach:
@mislav This is ready for review.

I took "template" to mean any default message, as opposed to a only GitHub template. Does that match your meaning? I check whether a default message exists by parsing the message pre-edits with `readTitleAndBody`.

e-beach:
@mislav friendly ping

shcallaway:
@mislav Do you mind looking at this? 😄

PR #1545, opened 2017-08-21
Add a label command

eljobe:
This command lists the labels for a given repository. This command
will be enhanced with subcommands which can create new labels on the
repository and modify existing labels (i.e. renaming or changing
their colors).

Other enhancements I plan to include in the future.
  * Display the labels using 256 bit colors similar to how the labels
    look on GitHub. This is similar to the functionality in the
    [ghi](https://github.com/stephencelis/ghi) tool.
  * When displaying issues with the `issue` command, use this same
    256 bit color formatting so that labels are easier to read on
    terminals which don't support true colors. (e.g. `Terminal.app`)

### DIFF
--- a/commands/issue.go
+++ b/commands/issue.go
@@ -19,6 +19,7 @@ var (
 		Usage: `
 issue [-a <ASSIGNEE>] [-c <CREATOR>] [-@ <USER>] [-s <STATE>] [-f <FORMAT>] [-M <MILESTONE>] [-l <LABELS>] [-d <DATE>] [-o <SORT_KEY> [-^]]
 issue create [-oc] [-m <MESSAGE>|-F <FILE>] [-a <USERS>] [-M <MILESTONE>] [-l <LABELS>]
+issue label
 `,
 		Long: `Manage GitHub issues for the current project.
 
@@ -28,6 +29,9 @@ With no arguments, show a list of open issues.
 
 	* _create_:
 		Open an issue in the current project.
+
+    * _label_:
+        List the labels available in this repository.
 
 ## Options:
 	-a, --assignee <ASSIGNEE>
@@ -190,6 +194,7 @@ func init() {
 	cmdIssue.Flag.BoolVarP(&flagIssueIncludePulls, "include-pulls", "", false, "INCLUDE_PULLS")
 
 	cmdIssue.Use(cmdCreateIssue)
+    cmdIssue.Use(cmdLabel)
 	CmdRunner.Use(cmdIssue)
 }
 
--- a/commands/label.go
+++ b/commands/label.go
@@ -1,0 +1,60 @@
+package commands
+
+import (
+	"github.com/github/hub/github"
+	"github.com/github/hub/ui"
+	"github.com/github/hub/utils"
+)
+
+
+var (
+	cmdLabel = &Command{
+		Run: listLabels,
+		Usage: `
+label
+`,
+		Long: `Manage GitHub labels for the current repository.
+
+## Commands:
+
+With no arguments, show a list of open issues.
+`,
+	}
+)
+
+func init() {
+	CmdRunner.Use(cmdLabel)
+}
+
+func listLabels(cmd *Command, args *Args) {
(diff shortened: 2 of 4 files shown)


eljobe:
Do you want me to use a similar sub-command pattern for milestones?  The idea would be to eventually have this set of commands:

- `issue label` (for listing)
- `issue label create`
- `issue label update`
- `issue milestone` (for listing)
- `issue milestone create`
- `issue milestone update`

I have to say, this seems very unintuitive for me. Especially because labels and milestones can both be applied to issues and pull requests, and because they both feel like repository-level concepts. It seems like they would be better suited for users at the top level of the command structure. I understand your desire to have as little conflict as possible with other `git-*` commands, but do we know of existing ones out there for `label` and `milestone`?

eljobe:
One more thing I'm not that crazy about is, `issue label create` sort of looks like you are trying to add a label to a specific issue instead of creating a label (but not necessarily adding it to an issue.)

mislav:
> I have to say, this seems very unintuitive for me. Especially because labels and milestones can both be applied to issues and pull requests, and because they both feel like repository-level concepts.

That's right. However, I also don't anticipate people creating/editing labels and milestones that often from the command line. This, plus being generally conservative in adding new commands, is why I would definitely avoid creating new top-level commands for this. In fact, I'm not even sure if we absolutely need milestone creation or label editing at all. 

I hope that you can understand that hub is increasing in scope with every PR accepted, and part of my job as a maintainer is to _resist_ pulling some proposed features. Every new command that I accept will be more code for me to mainain over the upcoming years. Even though a lot of features are nice-to-haves, it was never our goal to map the entire GitHub API to the command-line. Some interfaces are better left to the web UI (such as creating labels so that you can use the color picker widget).

For now, having a simple subcommand that lists available labels such as you are proposing here is something I could get behind of since it's relatively easy to implement and maintain. Plus, it wouldn't even be so far-fetched to scope this under `hub issue`, since PRs are basically "extended issues" (in a way that they implement all the concept of an Issue, but with extended features).

eljobe:
Well, you're the maintainer. So, it is entirely your prerogative.  I'll restructure the command(s) to be subcommands under the `issue` command.

However, I do think you might be underestimating the utility of this particular set of features for `hub`.   The most common use-case which causes me to want to be able to script label creation, update, and modification is that I am maintaining a set of labels across several related repositories. Imagine an organization which is running their business on GitHub. They are likely to have many dozen if not hundreds of GitHub repositories. Now, suppose this same organization begins to develop guidelines on common sets of labels which should be used on different types of repositories.  They'll want to script the maintenance of those labels. In fact, that's why I've been using [ghi](https://github.com/stephencelis/ghi).

A similar use-case arises when several different repositories participate in a single release-pipeline. This is why it is important to be able to update the set of milestones through scripting as well.

Anyway. I've said my piece. Now, I'll be quiet and just go off and code this stuff up.

casusa12345:
Fh

eljobe:
@casusa12345, I'm not sure what "Fh" means. Maybe, you prematurely posted the comment?

@mislav, I've moved the `label` command to be a subcommand of `issue` as requested. If it's cool with you, I plan to make the addition of colors and the other features we've been discussing as a separate pull request.  Alternatively, I would be happy for you to create a development branch in this repository where I can incrementally pull pieces of the functionality in, and have you review them. Then, once you thing there is a critical mass of functionality available, you can merge in the feature branch. Let me know what works best with your development style.

eljobe:
I've been thinking more about this. I guess it really doesn't matter if the command is a subcommand of some other command. I can always use a git alias to make the command structure work better for my workflow `alias.label=issue label`

timofonic:
@mislav There are lots of people prefer the command live, but living in a UI hell...
* @stephencelis' [ghi](https://github.com/stephencelis/ghi) has 1800+ stars, despite it stagnated (there's hope in @drazisil's fork). Is that enough for a small project with bad naming? ;)

darrenwee:
Can I pick up work on this PR to add the label colors?

eljobe:
Sure

Issue #1349, opened 2016-11-30
Allow edits from maintainers


galeksandrp:
Checkbox not enabled when creating PR from hub.

trevor-vaughan:
This is really causing a hiccup in my CLI workflow. Does anyone know where in the API this can be enabled? I took a look but couldn't find anything.

mislav:
Sorry folks, this feature is not yet exposed in GitHub API. I'm interested in having the option (on by default?) once it's possible.

travi:
my vote would be off by default since i'd rather others send PRs to my branch than push directly to them. that workflow preserves the safety needed to force push to your own branch

vitorgalvao:
> my vote would be off by default since i'd rather others send PRs to my branch than push directly to them. that workflow preserves the safety needed to force push to your own branch

Github has it on by default. Since this is a CLI tool geared in large part specifically for github, it should do whatever the website does by default.

travi:
That's true, but it's also targeted at power users that are more likely to force push. Seems like at least worth being configurable in the `hub` section of `~/.gitconfig`

trevor-vaughan:
I would vote for on by default and configurable.

ilovezfs:
The default seems to have recently changed from off to on. It would be nice to be able to configure this because that was pretty surprising to see the behavior change overnight with no way to control except through the GUI after the fact.

mislav:
I agree that this should be configurable. I also agree that the default should be whatever is the default in the web UI, which is "on". I'll check with my colleagues whether there is a plan to expose this through the API.

galeksandrp:
Currently "Allow edits from maintainers" is set by GitHub when pushing with `hub`.

xenohunter:
I'm too for a setting which would make the default state of that checkbox configurable.

galeksandrp:
Present in API https://developer.github.com/v3/pulls/#update-a-pull-request

mislav:
https://github.com/github/hub/pull/2168

PR #1550, opened 2017-08-27
Allow edits from maintainers

ganmacs:
I added the option `--restrict-maintainer-modfy` to `hub pull-request` command to off the checkbox of 'Allow edits from maintainers' in github\.com Web UI.
And also, restrict-maintainer-modfy can be configured at `~/.gitconfig` like a following code.

```
[hub]
  maintainerModify = false
```
Default value of  restrict-maintainer-modfy is `false` since github\.com Web UI is flase (means **Allow** edits from maintainers )

ref. #1349

### DIFF
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -144,8 +144,9 @@ func pullRequest(cmd *Command, args *Args) {
 	utils.Check(err)
 
 	var (
-		base, head string
-		force      bool
+		base, head       string
+		force            bool
+		maintainerModify bool
 	)
 
 	force = flagPullRequestForce
@@ -205,6 +206,17 @@ func pullRequest(cmd *Command, args *Args) {
 			err = fmt.Errorf("%s\n(use `-f` to force submit a pull request anyway)", err)
 			utils.Check(err)
 		}
+	}
+
+	mm, _ := git.Config("hub.maintainerModify")
+	switch mm {
+	case "", "true":
+		// flagPullRequestRestrictMaintainerModify overwrites hub.maintainerModify
+		maintainerModify = !flagPullRequestRestrictMaintainerModify
+	case "false":
+		maintainerModify = false
+	default:
+		utils.Check(fmt.Errorf("Invalid value for hub.maintainerModify. `true` or `flase` is valid."))
 	}
 
 	var editor *github.Editor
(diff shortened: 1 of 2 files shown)


commands/pull_request.go
```diff
@@ -202,6 +208,17 @@ func pullRequest(cmd *Command, args *Args) {
 		}
 	}
 
+	mm, _ := git.Config("hub.maintainerModify")
```
mislav: Could you skip adding a global configuration option for now? A more generic way of defining default flags for hub commands globally is on the roadmap, and it's likely that it wouldn't be stored in `.gitconfig`.

features/pull_request.feature
```diff
@@ -921,3 +921,15 @@ BODY
       """
     And the output should match /Given up after retrying for 5\.\d seconds\./
     And a file named ".git/PULLREQ_EDITMSG" should exist
+
+  Scenario: Explicit restrict-maintainer-modify
+    Given the GitHub API server:
+      """
+      post('/repos/mislav/coral/pulls') {
+        assert :maintainer_can_modify => false
```
mislav: You should also add this assertion to one of the existing scenarios that doesn't use the new flag. The assertion should be that `maintainer_can_modify` was set to either `true` or that it wasn't set at all.

commands/pull_request.go
```diff
@@ -69,6 +69,9 @@ pull-request -i <ISSUE>
 	-l, --labels <LABELS>
 		Add a comma-separated list of labels to this pull request.
 
+	-R, --restrict-maintainer-modify
```
mislav: I'm still not sold on the flag name `restrict-maintainer-modify`. It doesn't sound like something a human would use to express themselves.

The web UI checkbox is called "allow edits from maintainers". Unchecking that box would be equal to "do not allow edits from maintainers" in human speak. I'm wondering how we could express it in a short flag. Some options:

* `no-edits` (might be too vague/misleading)
* `no-maintainer-edits`
* `restrict-edits`

Any of these make sense? Maybe someone could come up with a better idea?

devnoname120: I like the second one, or maybe `no-maintainer-pushes`?

cwarden:
Sending `maintainer_can_modify` also fixes a confusing error, "Fork collab can't be granted by someone without permission" when creating an pull request from an issue.

See https://github.com/octokit/node-github/pull/491 and https://github.com/sociomantic-tsunami/git-hub/pull/218

commands/pull_request.go
danielcompton: I think `no-maintainer-edits` makes the most sense, especially in the context of the existing checkbox saying "allow edits from maintainers".

Issue #1494, opened 2017-06-15
Support client certificates for GitHub Enterprise


brandur:
This is a bit of a special snowflake situation, but our GitHub Enterprise has been configured with an additional layer of security whereby a client certificate needs to be sent along with all requests to it. 

I've built support for this and am wondering if I could possibly send it upstream. Would you be open to a patch that allows `hub` to support this through either something like a configuration option or an environmental variable?

My proposed interface would be a little like [how client certificates are sent in cURL](https://gheeknet.wordpress.com/2011/07/22/how-to-use-curl-to-access-a-https-url-using-a-client-certificate/), probably requiring that a certificate and private key are stacked into a single PEM-encoded file (with [possible PKCS#12 support through `golang.org/x/crypto/pkcs12`](https://godoc.org/golang.org/x/crypto/pkcs12)). It might be better as environmental variable given how few people are going to ever need this (a little like configuring a proxy).

Any thoughts? Thanks!

mislav:
Thanks for addressing this. I think a configuration option would be the best. Something like this in `~/.config/hub`:

```yaml
github.com:
- user: mislav
  oauth_token: PERSONAL_ACCESS_TOKEN
  protocol: https
  client_certificate: /path/to/cert.pem
```

A PR for this would be most welcome!

brandur:
Thanks for the response and guidance @mislav! I still need to clean up my diff into something patch-worthy, but I wanted to acknowledge that I still plan to do this.

PR #1554, opened 2017-08-30
Add support for sending client certificates

brandur:
Adds support for sending a client certificate to a GitHub Enterprise server.

The client certificate should be encoded with PKCS12 and its path should be set with the configuration option `client_certificate`.

@mislav Can I possibly get a sanity check on this one? I ended up pulling in the PKCS12 package simply because it seems like the most convenient way of encoding a certificate and key. We'd otherwise need to specify the paths to two different files, or stack PEMs into a single file, which I don't think is as widely supported. The nice part about this approach is that something like Mac OS's Keychain makes it quite easy to export a cert/key pair as a `*.p12`.

I'll warn you up front that there is a small possibility of having to add new miscellaneous improvements in the future to help support a more flexible version of this feature. For example, I've required that the PKCS12 file not be passphrase-protected for simplicity, but allowing a passphrase through hub might be desirable for someone else. Others might like to see support for extracting a certificate directly from the Mac OS Keychain (cURL built on Mac supports this) or support for other encodings. It's also possible that no one besides me will ever want or need this option.

Any thoughts? Thanks in advance!

Fixes https://github.com/github/hub/issues/1494.


github/client.go
```diff
@@ -685,12 +694,53 @@ func (client *Client) newOctokitClient(auth octokit.AuthMethod) *octokit.Client
 	host = normalizeHost(host)
 	apiHostURL := client.absolute(host)
 
-	httpClient := newHttpClient(os.Getenv("HUB_TEST_HOST"), os.Getenv("HUB_VERBOSE") != "")
+	tlsConfig, err := client.newTLSConfig()
+	if err != nil {
+		panic(err)
```
brandur: I put this `panic` here as a placeholder, but it may be desirable to refactor this method to return an error.

e-beach: There's also `utils.check(err)` which will quit hub if `err` is not nil.

brandur: Ah yes, great point! I changed over to use that in f8cb480. Thanks.

kandice1978:
I'm ap confused

brandur:
We ended up designing a different way to work around this problem, so closing out.

Issue #1067, opened 2016-01-05
hub complains "Unauthorized (HTTP 401)" and "Bad credentials"


gparker42:
hub 2.2.2 on OS X complains "Unauthorized (HTTP 401)" and "Bad credentials" when I try to check out a pull request. 

I have not used `hub` before. My credentials work in curl and on the web site. I have nothing in any OS X keychain matching "github". My usual github access uses an SSH key and no 2FA.

    % hub --version
    git version 2.6.3 (Apple Git-62)
    hub version 2.2.2
    
    % ls ~/.config/hub
    ls: /Volumes/precious/gparker/.config/hub: No such file or directory
    
[note `git` is not an alias, so this is real git and not hub]

    % git clone git@github.com:/apple/swift.git
    Cloning into 'swift'...
    remote: Counting objects: 318893, done.
    remote: Compressing objects: 100% (123/123), done.
    remote: Total 318893 (delta 42), reused 1 (delta 1), pack-reused 318768
    Receiving objects: 100% (318893/318893), 66.27 MiB | 4.56 MiB/s, done.
    Resolving deltas: 100% (261489/261489), done.
    Checking connectivity... done.
    Checking out files: 100% (9920/9920), done.
    
    % cd swift
    
    % env HUB_VERBOSE=1 hub checkout http://github.com/apple/swift/pull/862
    github.com username: gparker42
    github.com password for gparker42 (never stored): 
    > GET https://api.github.com/repos/apple/swift/pulls/862
    > Authorization: token 
    < HTTP 401
    {"message":"Bad credentials","documentation_url":"https://developer.github.com/v3"}
    Error getting pull request: Unauthorized (HTTP 401)
    Bad credentials
    
[note that I don't have anything in ~/.config/hub after this sequence]
    
    % curl -u gparker42 https://api.github.com/repos/apple/swift/pulls/862
    Enter host password for user 'gparker42':
    {
      "url": "https://api.github.com/repos/apple/swift/pulls/862",
      "id": 54923914,
      "html_url": "https://github.com/apple/swift/pull/862",
      […]

mislav:
Are you sure you haven't removed anything from the below output:

```
% env HUB_VERBOSE=1 hub checkout http://github.com/apple/swift/pull/862
github.com username: gparker42
github.com password for gparker42 (never stored): 
> GET https://api.github.com/repos/apple/swift/pulls/862
> Authorization: token 
< HTTP 401
[...]

PR #1562, opened 2017-09-06
Document HUB_VERBOSE

asfaltboy:
I found this undocumented option to be very useful when trying to understand why the "create" command failed for me. Found and snagged from #1067


mislav:
Thanks for the contribution! However, I felt that just `HUB_VERBOSE` didn't warrant its own section, since it's relatively simple. Instead, I wanted all the missing environment variables to be documented in one place, but I didn't want to send you on a hunt to discover what their names were and what they did. Instead, I did it so myself to save you time: https://github.com/github/hub/pull/1563. What do you think?

asfaltboy:
Brilliant @mislav , this is exactly what I secretly wished for!

PR #1566, opened 2017-09-08
Website update

### LINKED ISSUE #1564: Website design update
Hey folks 👋 

The GitHub design visual language has evolved so much since `hub` was first created, and I think that [the public website](https://hub.github.com) deserves some love. I'm sure that people come to the site and think "Hey this doesn't look very legit" or "How can I trust this, it doesn't even look like GitHub."

I'd like to give it some attention and pull in [Primer](https://github.com/primer/primer-css) for some added GitHubb-y-ness. I believe the repo for the site itself is private; is there a reason for that? Or, if I'm wrong, then I'm having trouble finding the relevant files in this repo.

JasonEtco:
As per the brief discussion in #1564, I've gone ahead and redesigned the website using [Primer](https://github.com/primer/primer-css), GitHub's CSS framework.

Here's a before and after (but I encourage you to pull it and check it out yourself before merging)

<details>
  <summary>Before</summary>
  
</details>

<details>
  <summary>After</summary>
  
</details>

<details>
  <summary>Full screen view</summary>
  
</details>

---

Because Primer is built on SCSS, it makes the most sense to pull in only what is needed and then compile it, so this PR adds Jekyll and the relevant scripts to set it all up. It follows the [scripts-to-rule-them-all](https://github.com/github/scripts-to-rule-them-all) practice.

- [x] General design update
- [x] Responsiveness
- [x] Update README with script instructions
- [x] SEO things (social card, meta tags)

---

**Note:** committing the `node_modules` folder to git is generally a bad idea, but because GitHub Pages can compile our SCSS using Jekyll it needs to have the Primer files available in the `gh-pages` branch. Just wanted to make that clear because I've had the same question in the past.

Closes #1564


mislav:
This is a great start! 🎉  A few notes:

* The redesign isn't a radical departure from the old style, at least not when looking at the screenshots. It feels like only a touch-up. Does this still warrant pulling in a CSS framework (Primer), in that case? I expected that we will be using Primer to make the site look significantly different than it looks now.
* Would it be possible to avoid checking in `node_modules`?

JasonEtco:
@mislav 👋 

I didn't want to make it radically different from what it is now; it can still feel like its own website and property, but by using Primer it still fits within the GitHub visual language. That was a conscious decision, but if you're looking for a total change that's cool too.

> Would it be possible to avoid checking in node_modules

If it's using Primer, then the short answer is no. Pages needs the Primer .scss files in GitHub to compile everything, so it needs to be checked in.

We could alternatively link out to an already compiled Primer bundle, but then we lose out on all the performance gains of not using all of Primer, and we can't use Primer variables in custom styles.

coffee75:
acess

mislav:
Hi @JasonEtco, I'm sorry that I've allowed this to get stalled. In its current form I definitely didn't want to pull it in because of the so many large changes to the gh-pages branch, mostly due to Primer.

I've just done a lighter overhaul of the site where I changed the fonts, the color scheme, and updated the documentation (including all the man pages too!) fc25ac35d2680b863d9c26f5682ffb54322e9ecd. The site could definitely still use improvements, but let's aim to doing such improvements incrementally.

Closing this due to staleness

JasonEtco:
No worries! I appreciate the followup ❤️

PR #1569, opened 2017-09-13
Add a new "assignees" command.

pcorpet:
Add a new "assignees" command to list GitHub users that can be assigned issues or pull requests for the current project.

### DIFF
--- a/commands/assignees.go
+++ b/commands/assignees.go
@@ -1,0 +1,40 @@
+package commands
+
+import (
+	"github.com/github/hub/github"
+	"github.com/github/hub/ui"
+	"github.com/github/hub/utils"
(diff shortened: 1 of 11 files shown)


pcorpet:
@mislav can I do something to make it easier to review?

mislav:
Hi @pcorpet, thank you for your contribution and sorry for the late review.

After some consideration, I have decided that—for now—this is best implemented in user scripts, rather than maintained in hub core. Now that there is a new `hub api` command, I will close this feature request and leave implementing this as an exercise to the reader. https://github.com/github/hub/releases/tag/v2.8.3 https://github.com/github/hub/pull/2016

PR #1573, opened 2017-09-17
travis: update go version

dvrkps:


e-beach:
@dvrkps You might want to check why tests failed. Also, I think you should really say `1.9` instead of `1.x`, variable versions leads to unstable builds.

dvrkps:
@e-beach i think ruby tests fails for some reason.

mislav:
Why are we updating Go version? This project is tailored for Go 1.8.x right now. We do, of course, want to support the latest Go version as well, but first and foremost we want to ensure that we maintain compatibility with Go 1.8 until we're ready to deliberatly drop that compatibility. Specifying `1.x` in Travis configuration means that the test suite wouldn't be run against 1.8 anymore.

Ruby tests sometimes have intermittent failures. I've restarted the build in hopes the failure would clear up.

dvrkps:
@mislav @e-beach I can add both 1.8.x and 1.9.x instead only 1.x for maintain compatibility ?

e-beach:
@dvrkps I think that is what Mislav is saying.

dvrkps:
I close this PR and add new with both go versions.

mislav:
You didn't have to close this PR: you could have simply pushed new commits to your branch. But thank you!

dvrkps:
I know, but i delete fork repo before.

Issue #1511, opened 2017-07-17
`hub release` only show the first 30 releases


pcorpet:
Although the API is not very clear about it (https://developer.github.com/v3/repos/releases/#list-releases-for-a-repository) it is possible to list more than 30 releases, but the `release` command in hub only get the first page.

A fix would be to check the http response header "Link" to check whether there's a `rel=next` resource and fetch it as well. We might want to hide that behind a feature flag. Or not.

mislav:
Good idea. The `hub issue` command in master [already fetches all pages of records](https://github.com/github/hub/blob/6ed9277af121c99a81ebbbcd2dc249c786646a05/github/client.go#L506-L516) by following the `next` link, and `hub release` should follow the same approach.

However, I think both commands should also implement a flag to limit the total amount of entries returned, in case a repository has a zillion issues or releases. That flag could also be used to opt-out of pagination by setting it to a low number such as 10 or 30.

patheticpat:
I started working on this and got the first part working.

But I am unsure about how to implement the flag to limit the number of releases shown. Because the filtering for prereleases and drafts happens on the client, we don't know in advance how many releases we have to request from the server. Should we just get all of them and then display however many the user asked for?

e-beach:
Couldn't a limit argument still place a limit on the total number of releases returned? For example, If you excluded prereleases, with `--limit 30` you'd get 30 non-prerelease results.

patheticpat:
My point was that we don't know the number of releases to retrieve from the server in order to have 30 non-prereleases left after filtering.

PR #1579, opened 2017-10-02
Fetch all releases when doing `hub release`

patheticpat:
As suggested at #1511, instead of fetching only the first 30 releases,
follow the `next` link to fetch all pages.


mislav:
Thank you!

PR #1581, opened 2017-10-03
Add limit flag for issue and release commands

patheticpat:
This is a rather naive implementation, it still fetches all issues or
releases (once #1511 is fixed) from the server and displays just the
requested number of them. But maybe that is good enough?


mislav:
Hi, thanks for your contribution! The code looks good, but the limit is purely cosmetic for now. Ideally, the limit should prevent extraneous HTTP requests from being made. Have you considered that direction of implementation?

patheticpat:
Yes, I thought about that. Since both the issues and releases are filtered before display (for pull requests and drafts or prereleases, respectively) we can't know the number of items we have to request from the server in order to satisfy the limit. We could move the filtering into the client, so it would be able to know when to stop requesting any more items. Should I look further into this option?

mislav:
> we can't know the number of items we have to request from the server in order to satisfy the limit.

We can just keep getting N items per page (where N is the limit, or `limit + 10` or something like that) and paginate until the limit is satisfied or until there are no more pages.

patheticpat:
Yes, but then the filtering logic would have to be moved into the client.

mislav:
> Yes, but then the filtering logic would have to be moved into the client.

I see. Yes, it does sound like more work. But ultimately, one of the main points of limiting is that the command executes faster due to fewer HTTP requests made. Therefore, I think it's worth it.

I can take a stab at filtering in the client if you can't put in the time right now. It's understandable. The code in this PR is a good starting point. 🙇

patheticpat:
It's actually not that much more work, but I didn't want to change the client interface without discussing it first. I'll update the PR tomorrow.

github/client.go
```diff
@@ -274,7 +274,10 @@ type ReleaseAsset struct {
 	ApiUrl      string `json:"url"`
 }
 
-func (client *Client) FetchReleases(project *Project) (releases []Release, err error) {
+// FetchReleases fetches Releases from project, optionally including
+// drafts or excluding prereleases. If limit is positive, it returns at
+// most limit Releases.
+func (client *Client) FetchReleases(project *Project, includeDrafts, excludePrereleases bool, limit int) (releases []Release, err error) {
```
patheticpat: Maybe it would be cleaner to be consistent here and make that parameter `includePrereleases`?

mislav:
Thank you for the help!

PR #1582, opened 2017-10-03
Fix PATH update in windows installation script

patheticpat:
For some reason, when the installation script is run through the VBS
wrapper on Windows 7, the `OLDPATH` gets cut in multiple parts and only
the final part makes it into the new `PATH`. This fix just collects all
the parts and concatenates them again.


mislav:
Thank you! Does the updated script work correctly like before on newer Windows?

patheticpat:
Yes. All it does is reassemble the path in case it gets split into multiple parts in the for loop.

PR #1598, opened 2017-10-15
Added additional way of installation

mehandes:
Added installation through Nix package manager


ashanaakh:
I've created __better cli for github__ https://github.com/ashanaakh/gitius. Check it out! 😉

mehandes:
@ashanaakh Thanks! Will do.

PR #1599, opened 2017-10-16
Ensure that _git completion func is replaced only once.

voithos:
This allows the completion script to be sourced more than once while
still behaving correctly. Previously, the script would attempt to
replace `__git_list_all_commands` blindly, which would yield
multiply-replaced calls like
`__git_list_all_commands_without_hub_without_hub`.

The zsh completions already do something very similar to this to ensure
that the replacement isn't done more than once.

### DIFF
--- a/etc/hub.bash_completion.sh
+++ b/etc/hub.bash_completion.sh
@@ -6,8 +6,8 @@ if ! declare -F _git > /dev/null && declare -F _completion_loader > /dev/null; t
   _completion_loader git
 fi
 
-# Check that git tab completion is available
-if declare -F _git > /dev/null; then
+# Check that git tab completion is available and we haven't already set up completion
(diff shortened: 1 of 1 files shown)


etc/hub.bash_completion.sh
```diff
@@ -8,13 +8,14 @@ fi
 
 # Check that git tab completion is available
 if declare -F _git > /dev/null; then
```
mislav: This existing condition already wraps the entire logic of this file, right? How about if the only change we make is simply expand this condition to be:

```sh
if declare -F _git >/dev/null && ! declare -F __git_list_all_commands_without_hub >/dev/null; then
```

Checking if `__git_list_all_commands_without_hub` is defined means that this file already executed. What do you think?

voithos: Sure, that seems like it'll work fine as well (and it avoids having to add a new global variable). I was originally trying to mimic the zsh completion script, just to get a bit more consistency. Made the change.

PR #1603, opened 2017-10-16
Fix recognition logic for vi editor variants

patheticpat:


mislav:
Oh I merged prematurely. I just spotted an error: the regexp should end in `$`. Care to make another PR?

patheticpat:
Why not `\b`? I was thinking that people might want to pass commandline flags via their editor config.

mislav:
> Why not `\b`? I was thinking that people might want to pass commandline flags via their editor config.

You're right. However, `\b` is a word boundary and would match a program such as `vim-foo-bar`, which is most definitely not vim proper.

We can access the parsed executable name from the `Cmd` instance that's constructed just before the regexp: https://github.com/github/hub/blob/cbcc349e39c359ea2bd1f3f0d2ad56463750a905/cmd/cmd.go#L95-L99

```go
if r.MatchString(editCmd.Name) {
```

patheticpat:
Oh yes, you're right, I didn't think of this. But while we're at it, we should probably also throw in an optional `\.exe` for Windows users.

PR #1604, opened 2017-10-17
Fix recognition logic for vi editor variants again

patheticpat:
This also considers `.exe` files. It still matches names like `foo-vim`, but it is probably safe to assume that this is in fact something vim-like.


mislav:
Thank you!

PR #1606, opened 2017-10-21
Fix git and hub versions to latest ones

superuserAV:
Latest **git** version is _2.14.2_, latest **hub** version is _2.2.9_.


mislav:
Thanks, but we won't be updating documentation every time any of these version numbers change. The output in the examples are supposed to be illustrating what the output will *look* like, not exactly what it will be.

PR #1608, opened 2017-10-24
Add MacPorts installation instructions

chrisvire:
NOTE: It doesn't appear that `hub` provides any variants (e.g. pre-releases) in MacPorts

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,17 @@ development by reporting bugs, you can install the prerelease version:
 $ brew install --devel hub
 ```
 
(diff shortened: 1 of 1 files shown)


mislav:
Thanks for the suggestion, but we prefer Homebrew for installation on macOS, and Homebrew and Macports are known to clash against each other. Since I don't personally recommend Macports, I will refrain from including these instructions. If somebody already has Macports and is experienced with it, they will easily find out how to install hub.

PR #1640, opened 2017-12-31
Add installation method for archlinux

asdofindia:
Arch Linux has hub in its official repository. But I spent an hour installing go and other things because I thought hub would not be in the official repository.


asdofindia:
@sinisterstuf thanks. Done exactly that.

Issue #1280, opened 2016-09-22
Retrieve available issue labels


vermiculus:
It would be super slick if `hub` could retrieve the current list of labels available for use when creating a repository -- vermiculus/magithub#6 (and other hub-wrapping scripts) could really benefit from this.

mislav:
Good idea! Perhaps an `hub issues labels` subcommand would be in order.

vermiculus:
@mislav I'd like to get my hands dirty with some Go – I've never worked with it before. Are there any license restrictions for contributions to hub?

mislav:
@vermiculus Yes, all contributors have to accept [GitHub's CLA](https://cla.github.com/). When you submit your first pull request, the automated status checks will fail with a link to accept the CLA. It's done by reviewing the agreement and pressing a button. It's easy. When you do that, the status check will start passing.

vermiculus:
In case someone has time before I do (and so I don't forget), it would help greatly if labels were paired with some sort of ID and this ID was usable with `issue --format` as `%L`. Consider the following label setup:

1. `A`
2. `B`
3. `A, B`

Any given issue that lists `%L=A, B` would be ambiguous between 1+2 or just 3.

eljobe:
I propose that we make it a separate command (as opposed to a subcommand of the `issue` command).  My arguments:

* Labels are naturally the children of the repository, not the issue. See: https://developer.github.com/v3/issues/labels/
* You can also use these labels on pull requests.
* As a user, I think of operations like creating a label or modifying a label as separate from operations I perform on issues.

I'm starting work on a branch that implements listing labels for a repository right now. But, I probably won't finish it in the 45 minutes before my next meeting.

mislav:
This was done in #1667.

PR #1667, opened 2018-01-30
Add label command with colorized output

darrenwee:
Addresses issue #1280 and based off #1545.

Usage: `hub issue labels` in a repository prints a list of labels with their corresponding colors.

This works okay for repos with a small list of labels (such as this one) but is not as elegant for a repository with a lot of labels. I propose either printing them in lines grouped by their color or paging the output.


commands/issue.go
```diff
@@ -428,3 +437,55 @@ text is the title and the rest is the description.`, project))
...
+	if args.Noop {
+		ui.Printf("Would request list of labels for %s\n", project)
+		return
+	}
+
+	labels, err := gh.FetchLabels(project)
+	utils.Check(err)
+
+	for _, label := range labels {
+		// TODO add a colorize flag for the labels command
```
mislav: Would this be enough?

```go
colorize := ui.IsTerminal(os.Stdout)
```

commands/issue.go
```diff
@@ -428,3 +437,55 @@ text is the title and the rest is the description.`, project))
...
+		utils.Check(err)
+	}
+
+	placeholders := map[string]string{
+		"l": label.Name,
+		"c": fmt.Sprintf("\033[38;5;%d;48;2;%d;%d;%dm %s \033[m",
+			getSuitableTextColor(color), color.Red, color.Green, color.Blue, label.Name),
+	}
+
+	return ui.Expand(format, placeholders, colorize)
```
mislav: I think `ui.Expand` with placeholders might have been an overkill here. If the user doesn't have the option to pass the `--format` string (like for `hub issue` or `hub pr list`), then we could get away with simply:

```go
if colorize {
  return fmt.Sprintf("output with color", ...)
} else {
  return fmt.Sprintf("%s\n", label.Name)
}
```

commands/issue.go
```diff
@@ -428,3 +437,55 @@ text is the title and the rest is the description.`, project))
...
+	}
+
+	color, err := utils.NewColor(label.Color)
+	if err != nil {
+		utils.Check(err)
+	}
+
+	placeholders := map[string]string{
+		"l": label.Name,
+		"c": fmt.Sprintf("\033[38;5;%d;48;2;%d;%d;%dm %s \033[m",
```
mislav: Can the logic to construct the ANSI escape sequence for a specific color be shared between this command and listing issues?

commands/issue.go
```diff
@@ -428,3 +437,55 @@ text is the title and the rest is the description.`, project))
...
+
+	labels, err := gh.FetchLabels(project)
+	utils.Check(err)
+
+	for _, label := range labels {
+		// TODO add a colorize flag for the labels command
+		ui.Printf(formatLabel(label, true))
+	}
+
+	args.NoForward()
```
mislav: `NoForward` needs to be above `if args.Noop()` because of the potential early `return`

commands/issue.go
darrenwee: I was thinking along the lines of a command line argument to control whether to print in plain text or a colorized string, like `hub issue labels --no-color`. I could add that in plus the `IsTerminal` check?

mislav:
Thank you! I have swapped the `--no-color` flag in favor of `--color` (explicit opt-in) because I agree with you that the default colored output isn't very useful when labels stand on their own. I imagine that people will be mostly using this command for scripting anyway, where colors are irrelevant.

PR #1677, opened 2018-02-12
Add milestone to PR by name instead of number

### LINKED ISSUE #1089: Getting an error when trying to set milestone
``` bash
$ hub pull-request -M "3.12"
invalid argument "3.12" for -M: strconv.ParseUint: parsing "3.12": invalid syntax
usage: git pull-request [-f] [-m <MESSAGE>|-F <FILE>|-i <ISSUE>|<ISSUE-URL>] [-o] [-b <BASE>] [-h <HEAD>] [-a <USER>] [-M <MILESTONE>] [-l <LABELS>]
...
$ hub --version
git version 2.5.4 (Apple Git-61)
hub version 2.2.0-124-gd0ad43d
```

venyii:
Now it's possible to add a milestone by name to a pull request instead of looking up the number. If the milestone can't be found it'll stop before creating the PR.

Closes #1089

### DIFF
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -63,8 +63,9 @@ pull-request -i <ISSUE>
 	-a, --assign <USERS>
 		A comma-separated list of GitHub handles to assign to this pull request.
 
-	-M, --milestone <ID>
-		Add this pull request to a GitHub milestone with id <ID>.
+	-M, --milestone <NAME>
+		The milestone name to add to this pull request. Passing the milestone number
+		is deprecated.
 
 	-l, --labels <LABELS>
 		Add a comma-separated list of labels to this pull request.
@@ -85,6 +86,7 @@ var (
 	flagPullRequestHead,
 	flagPullRequestIssue,
 	flagPullRequestMessage,
+	flagPullRequestMilestone,
 	flagPullRequestFile string
 
 	flagPullRequestBrowse,
@@ -92,8 +94,6 @@ var (
 	flagPullRequestEdit,
 	flagPullRequestPush,
 	flagPullRequestForce bool
-
-	flagPullRequestMilestone uint64
 
 	flagPullRequestAssignees,
 	flagPullRequestReviewers,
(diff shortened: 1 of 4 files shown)


features/pull_request.feature
```diff
@@ -757,19 +757,44 @@ BODY
...
+  Scenario: Pull request fails with unknown milestone before it's created
+    Given I am on the "feature" branch with upstream "origin/feature"
+    Given the GitHub API server:
+      """
+      get('/repos/mislav/coral/milestones') {
+        status 200
+        json [
+          { :number => 1234, :title => "prerelease" },
+          { :number => 1337, :title => "v1" },
+          { :number => 41319, :title => "Hello World!" }
```
mislav: You can return an empty JSON array here, since we know in advance that no milestones will match "unkown".

commands/pull_request.go
```diff
@@ -423,3 +428,16 @@ func parsePullRequestIssueNumber(url string) string {
 
 	return ""
 }
+
+func fetchMilestoneNumber(client *github.Client, project *github.Project, name string) (int, error) {
```
mislav: To simplify this method's signature, instead of passing in `client` and `project` to this method, how about passing an already fetched list of milestones?

commands/pull_request.go
```diff
@@ -423,3 +428,16 @@ func parsePullRequestIssueNumber(url string) string {
...
+	milestones, err := client.FetchMilestones(project)
+	utils.Check(err)
+
+	for _, milestone := range milestones {
+		if milestone.Title == name {
```
mislav: How about case-insensitive match?

venyii:
@mislav thanks for the quick review, I implemented your comments - ready for review again :)

venyii:
Yes, thought about this as well but went ahead with it since it was in the 2.3 project 😏.

Off the top of my head, we could use the old behaviour **if** the number is an integer (successful `strconv.Atoi`)? That would of course prohibit the use of milestone names that are just integers, but would maybe be enough BC until 3.0 comes around (if planned)?

I think a completely new option isn't worth it and produces BC breaks anyway in the next major version if we intend to replace the old `-M` behavior.

Just speaking for myself, I used the option once and didn't bother with it again since it was too feasible for me to search the milestone number every time 😃 but not sure how others use this.

Either way, if you've got another idea I've got no problems changing the PR! Just let me know and I'll make the changes

venyii:
ping @mislav, could you take a look at my last comment regarding the BC? I'd love to get this merged :)

mislav:
> Off the top of my head, we could use the old behaviour **if** the number is an integer (successful `strconv.Atoi`)? That would of course prohibit the use of milestone names that are just integers, but would maybe be enough BC until 3.0 comes around (if planned)?

Yes please; let's check if the number looks like an integer and use it as milestone ID for backwards compatibility.

venyii:
@mislav done! It now uses the given number directly if it passes the integer conversion, also added a test for it. I added the deprecation info to the parameter doc, let me know if that's enough info

mislav:
Thank you a ton!! 🏆

PR #1690, opened 2018-03-03
Make remote URLs default to HTTPS

### LINKED ISSUE #1596: hub clone should use HTTPS instead of SSH by default so the GCM can be used
When interacting with a GitHub repo via HTTPS, the [Git Credential Manager for Windows](https://github.com/Microsoft/Git-Credential-Manager-for-Windows) is used by default (it is installed by default with Git for Windows). The GCM is great, because it supports GitHub 2FA, so you don't need to create a token or SSH key but can just log in with your username, password and 2FA token.

However, `hub clone` uses the SSH URL of the repo by default (instead of the HTTPS URL). This means that when a user installs hub and blindly does a `hub clone` on a repo, they won't be able to use the GCM (and thus will need to do extra configuration if they have 2FA set up).

This seems like an oversight. Therefore, I propose using HTTPS instead of SSH by default (at least on Windows).

darrenwee:
Fixes #1596 and closes #1318 

This makes URLs default to using HTTPS instead of the plain git protocol (`git://`).

I think I got all the tests related to this change. Let me know if I missed out anything!

### DIFF
--- a/commands/clone.go
+++ b/commands/clone.go
@@ -104,7 +104,7 @@ func transformCloneArgs(args *Args) {
 				if !isSSH &&
 					args.Command != "submodule" &&
 					!github.IsHttpsProtocol() {
-					isSSH = repo.Private || repo.Permissions.Push
+					isSSH = repo.Private
 				}
 
 				url := project.GitURL(name, owner, isSSH)
--- a/commands/clone.go
+++ b/commands/clone.go
@@ -101,12 +101,6 @@ func transformCloneArgs(args *Args) {
 					}
 				}
 
-				if !isSSH &&
-					args.Command != "submodule" &&
-					!github.IsHttpsProtocol() {
-					isSSH = repo.Private
-				}
-
 				url := project.GitURL(name, owner, isSSH)
 				args.ReplaceParam(i, url)
 			}
--- a/github/project.go
+++ b/github/project.go
@@ -73,14 +73,19 @@ func (p *Project) GitURL(name, owner string, isSSH bool) (url string) {
 	host := rawHost(p.Host)
 
 	if preferredProtocol() == "https" {
-		url = fmt.Sprintf("https://%s/%s/%s.git", host, owner, name)
-	} else if isSSH || preferredProtocol() == "ssh" {
-		url = fmt.Sprintf("git@%s:%s/%s.git", host, owner, name)
-	} else {
-		url = fmt.Sprintf("git://%s/%s/%s.git", host, owner, name)
+		return fmt.Sprintf("https://%s/%s/%s.git", host, owner, name)
 	}
 
-	return url
+	if isSSH || preferredProtocol() == "ssh" {
+		return fmt.Sprintf("git@%s:%s/%s.git", host, owner, name)
+	}
+
+	if preferredProtocol() == "git" {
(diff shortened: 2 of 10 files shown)


commands/clone.go
```diff
@@ -101,12 +101,6 @@ func transformCloneArgs(args *Args) {
 					}
 				}
 
-				if !isSSH &&
-					args.Command != "submodule" &&
-					!github.IsHttpsProtocol() {
-					isSSH = repo.Private || repo.Permissions.Push
```
mislav: Why did this logic go away? I imagine it should still be relevant if someone is explicitly configuring hub to use protocols other than `https` (if they want to keep old behavior)

github/project.go
```diff
@@ -73,14 +73,19 @@ func (p *Project) GitURL(name, owner string, isSSH bool) (url string) {
...
-		url = fmt.Sprintf("git://%s/%s/%s.git", host, owner, name)
+		return fmt.Sprintf("https://%s/%s/%s.git", host, owner, name)
 	}
 
-	return url
+	if isSSH || preferredProtocol() == "ssh" {
+		return fmt.Sprintf("git@%s:%s/%s.git", host, owner, name)
+	}
+
+	if preferredProtocol() == "git" {
```
mislav: If someone specifies `git` as their preferred protocol, does that still use `ssh` for "writeable" repos, e.g. cloning a repo to which you have writeable permissions? (This is due `git` protocol not being pushable.)

This would be a way people could opt into the old behavior before this change, but this behavior doesn't yet seem exercised in depth by updates to tests.

commands/clone.go
darrenwee: Hi @mislav, sorry for the delay. School's been busy!

This would be taken care of by `Project.GitURL`. If the `hub.protocol` is set to `ssh` or `git`, this would override the defaulting-to-`https` and use the user-defined protocol. Removing this block of logic just prevents the URLs from defaulting to `ssh` even if the repo is private or you have push permission. I felt that this would be in line with the spirit of #1596 for first-time users of `hub` to be able to use their GCM without any surprises.

github/project.go
darrenwee: Nope, if `git` is specified as the preferred protocol, all repos cloned will use that protocol regardless of their repo privileges/privacy. Wouldn't it be appropriate to respect that setting and clone using the plain `git` URL?

> This would be a way people could opt into the old behavior before this change, but this behavior doesn't yet seem exercised in depth by updates to tests.

I'm not clear on what you mean. Are you referring to people who currently use `hub`, and currently set `hub.protocol` to `git`, and not surprising them with this change in behavior?

darrenwee: Hmm, okay, that sounds fair. This change should be made inside `clone.go`, right? Something like this?

```
				if !isSSH &&	
					args.Command != "submodule" &&	
					github.IsGitProtocol() {	
					isSSH = repo.Private || repo.Permissions.Push
```

Werneth57:
I know that sounds in a real world setting confusing to most computer geeks but actually to a accountant it makes great sense to me. It gives me the option to do so much more with my numbers at home because I only have one computer.  Unfortunately, I am new to Github and all the computer protocols so it makes it difficult for me to branch things correctly in a systematic way. But give me a number and I will blow your mind.

PR #1695, opened 2018-03-09
[RFC] Make fork --remote-name=origin rename origin

giuscri:
If user passes `origin` as the remote name
pointing to forked repository it's likely
she wants to rename origin as upstream.

### DIFF
--- a/commands/fork.go
+++ b/commands/fork.go
@@ -112,7 +112,7 @@ func fork(cmd *Command, args *Args) {
 		originURL := originRemote.URL.String()
 		url := forkProject.GitURL("", "", true)
 
-		// Check to see if the remote already exists and points to the fork.
+		// Check to see if the remote already exists.
 		currentRemote, err := localRepo.RemoteByName(newRemoteName)
 		if err == nil {
 			currentProject, err := currentRemote.Project()
@@ -120,6 +120,11 @@ func fork(cmd *Command, args *Args) {
 				if currentProject.SameAs(forkProject) {
 					ui.Printf("existing remote: %s\n", newRemoteName)
 					return
+				}
+				if newRemoteName == "origin" {
+					// Assume user wants to follow github guides for collaboration
+					ui.Printf("renaming existing \"origin\" remote to \"upstream\"\n")
+					args.Before("git", "remote", "rename", "origin", "upstream")
 				}
 			}
 		}
--- a/features/fork.feature
+++ b/features/fork.feature
@@ -34,12 +34,16 @@ Feature: hub fork
         json :name => 'dotfiles', :owner => { :login => 'mislav' }
       }
       """
-    And I successfully run `git remote rename origin upstream`
     When I successfully run `hub fork --remote-name=origin`
-    Then the output should contain exactly "new remote: origin\n"
+    Then the output should contain exactly:
+      """
+      renaming existing "origin" remote to "upstream"
+      new remote: origin\n
+      """
     And "git remote add -f origin git://github.com/evilchelu/dotfiles.git" should be run
     And "git remote set-url origin git@github.com:mislav/dotfiles.git" should be run
     And the url for "origin" should be "git@github.com:mislav/dotfiles.git"
+    And the url for "upstream" should be "git://github.com/evilchelu/dotfiles.git"
 
   Scenario: Fork the repository with redirect
     Given the GitHub API server:


commands/fork.go
```diff
@@ -112,16 +112,15 @@ func fork(cmd *Command, args *Args) {
 		originURL := originRemote.URL.String()
 		url := forkProject.GitURL("", "", true)
 
-		// Check to see if the remote already exists and points to the fork.
```
giuscri: What's the point of this check? Even if you have a remote called the same way of `newRemoteName` but pointing to a different project, `git` will later block you

giuscri:
Marking the pr as WIP as this is breaking a bunch of scenarios.

commands/fork.go
mislav: Not sure why this check is here, but I believe it was wanted behavior. Is it hard to preserve?

commands/fork.go
```diff
@@ -112,16 +112,15 @@ func fork(cmd *Command, args *Args) {
...
-					ui.Printf("existing remote: %s\n", newRemoteName)
-					return
-				}
+			if newRemoteName != "origin" {
+				ui.Printf("existing remote: %s\n", newRemoteName)
+				return
 			}
+			// Assume user wants to follow github guides for collaboration
+			ui.Printf("renaming %s as upstream\n", newRemoteName)
+			args.Before("git", "remote", "rename", newRemoteName, "upstream")
```
mislav: What if `upstream` already exists?

giuscri: hub will fail because of git: upstream already exists. Actually Printf of origin being renamed is there to allow user understand that potential git error.

commands/fork.go
giuscri: no, I've left split the change in two commits. Only applying the second will remove that behaviour.

giuscri:
@mislav is there anything i can change to make this pr merged?

features/fork.feature
```diff
@@ -167,10 +170,10 @@ Feature: hub fork
       }
       """
     When I run `hub fork`
-    Then the exit status should be 128
-    And the stderr should contain exactly:
+    Then the exit status should be 0
```
mislav: Why is the exit status success? Since `mislav` already exists as remote, `hub fork` should fail.

features/fork.feature
```diff
@@ -167,10 +170,10 @@ Feature: hub fork
...
+    And the stdout should contain exactly:
       """
-      fatal: remote mislav already exists.\n
+      existing remote: mislav\n
```
mislav: Why has the error message in this case changed?

features/fork.feature
```diff
@@ -34,9 +34,12 @@ Feature: hub fork
         json :name => 'dotfiles', :owner => { :login => 'mislav' }
       }
       """
-    And I successfully run `git remote rename origin upstream`
     When I successfully run `hub fork --remote-name=origin`
-    Then the output should contain exactly "new remote: origin\n"
+    Then the output should contain exactly:
+      """
+      renaming origin as upstream
```
mislav: Could the message say `renaming existing "origin" remote to "upstream"` so that it's more readable and clearer as to what's happening?

features/fork.feature
giuscri: because it's now coming from hub, not git. This unifies `Unrelated remote already exists` with other related scenarios like `Related fork and related remote already exist`, `Related fork and related remote, but with differing protocol, already exist`, ...

features/fork.feature
giuscri: This unifies `Unrelated remote already exists` with other related scenarios like `Related fork and related remote already exist`, `Related fork and related remote, but with differing protocol, already exist`, ...

mislav: I don't see this as a desired change. This PR was supposed to only introduce an extra feature: rename "origin" to "upstream" when attempting to add "origin" fork. Everything else should stay the same.

giuscri:
OK, I've dropped the last commit basically. This PR now provides only the renaming of origin to upstream.

I still think that
```
115                 // Check to see if the remote already exists.
116                 currentRemote, err := localRepo.RemoteByName(newRemoteName)
117                 if err == nil {
118                         currentProject, err := currentRemote.Project()
119                         if err == nil {
120                                 if currentProject.SameAs(forkProject) {
121                                         ui.Printf("existing remote: %s\n", newRemoteName)
122                                         return
123                                 }
```
should be partially removed: even when currentProject is _not_ the same of forkProject git will fail as it refuses to create two remotes with the same name.

btw.

features/fork.feature
```diff
@@ -34,9 +34,12 @@ Feature: hub fork
         json :name => 'dotfiles', :owner => { :login => 'mislav' }
       }
       """
-    And I successfully run `git remote rename origin upstream`
     When I successfully run `hub fork --remote-name=origin`
-    Then the output should contain exactly "new remote: origin\n"
+    Then the output should contain exactly:
+      """
+      renaming existing "origin" remote to "upstream"
```
mislav: This verifies command ouput, but can there be an additional step that verifies that the `upstream` remote now exist and that it points to the same thing `origin` used to point to?

giuscri: would
```
diff --git a/features/fork.feature b/features/fork.feature
index f5738cb..0afcb18 100644
--- a/features/fork.feature
+++ b/features/fork.feature
@@ -43,6 +43,7 @@ Feature: hub fork
     And "git remote add -f origin git://github.com/evilchelu/dotfiles.git" should be run
     And "git remote set-url origin git@github.com:mislav/dotfiles.git" should be run
     And the url for "origin" should be "git@github.com:mislav/dotfiles.git"
+    And the url for "upstream" should be "git@github.com:evilchelu/dotfiles.git"
 
   Scenario: Fork the repository with redirect
     Given the GitHub API server:
```
be enough?

mislav: Yep!

giuscri:
ping

mislav:
Thank you for your patience! 🙇

Issue #1674, opened 2018-02-08
hub ignore


mohsen1:
Add a command for initializing a `.gitignore` file. Using data in https://github.com/github/gitignore it should accept options for type of project. 

Something like 
```sh
hub ignore nodejs
```

mislav:
It's an interesting feature idea. Not sure if it fits on our roadmap, but I'll keep it open until we figure that out.

giuscri:
What should happen when there's already a .gitignore?
1. Refuse to continue
```
$ hub init-ignore nodejs
`.gitignore` already exits in this git repository. Remove it before
running `hub init-ignore nodejs` # <- shameless copy of `vagrant init` error
```
2. No problem as the command just prints Node.gitignore content on stdout
```
$ hub ignore nodejs >> .gitignore
```

sfdye:
Maybe this will be useful?
https://developer.github.com/v3/gitignore/

mislav:
After some consideration, I have decided that—for now—this is best implemented in user scripts, rather than maintained in hub core. Now that there is a new `hub api` command, I will close this feature request and leave implementing this as an exercise to the reader. https://github.com/github/hub/releases/tag/v2.8.3 https://github.com/github/hub/pull/2016

For example:
```sh
hub api gitignore/templates/Ruby | jq -r .source
```

Thank you for suggesting!

PR #1698, opened 2018-03-14
[WIP] hub ignore

giuscri:
See https://github.com/github/hub/issues/1674

Currently works only with `-l <LANGUAGE>`
option.

Missing error handling, list of all
available templates.


mislav:
Thank you for your contribution! As stated in the referenced issue, I'm not sure whether this falls into the feature set we want to see and maintain in hub. I don't see this being used often, and a new `hub ignore` command sounds ambiguous to me (it's not clear what will happen when you run it) and will shadow any custom `git-ignore` scripts that people might have on their systems.

In general, we are wary of creating new hub subcommands unless there is a compelling reason to do so.

giuscri:
should I stop working on this?

/cc @mohsen1

mislav:
Yes I think it's OK to leave this PR as-is and not to invest too much time polishing it up before we determine whether we want the feature or not.

PR #1705, opened 2018-03-18
Fix base branch detection during `hub pull-request`

### LINKED ISSUE #1538: pull-request is not setting base correctly
I have a repo that had (at one point) github configured to use a default branch _other than_ master. (It was an accident, but the old default was `FOUN-514-create-json-schema`)

GitHub has since been fixed, so the default is now `master` as it should be. (This was fixed a _long_ time ago.)

However, whenever hub begins a new PR, it defaults the base to that old default branch: `FOUN-514-create-json-schema`. I've checked the repo's git config, and all is as it should be (no straggling references to that old branch). I've checked in `~/.config/hub` but I don't see anything in there persisting the old default.

Does hub get this from the github api each time? Or is it persisted somewhere locally that I can flush?

It's worth noting that GitHub sets the base branch correctly when opening a PR via the UI (as confirmation that the default branch is set correctly).

PR message generated by hub:
```

# Requesting a pull to webanalytics:FOUN-514-create-json-schema from webanalytics:github-rename
#
# Write a message for this pull request. The first block
# of text is the title and the rest is the description.
```

```
$ git remote show origin 
* remote origin
  Fetch URL: git@github.shuttercorp.net:webanalytics/analytics-schema.git
  Push  URL: git@github.shuttercorp.net:webanalytics/analytics-schema.git
  HEAD branch: master
  Remote branches:
    github-rename     tracked
    master            tracked
    qa                tracked
    refactor-test     tracked
    refs/pull/1/head  tracked
    [snipped 'pull' refs]
  Local branches configured for 'git pull':
    github-rename merges with remote github-rename
    master        merges with remote master
  Local refs configured for 'git push':
    github-rename pushes to github-rename (up to date)
    master        pushes to master        (up to date)
```

mafredri:
Pull requests now use `default_branch` from the repository configuration (JSON) when no `.git/refs/remotes/origin/HEAD` is present. If the two differ, a warning is printed to the console and the local base branch is used.

This PR covers the following scenarios:

* There is no `.git/refs/remotes/origin/HEAD`, uses `default_branch` (from JSON)
* The branch in `.git/refs/remotes/origin/HEAD` differs from `default_branch`, uses `.git/refs/remotes/origin/HEAD` and prints a warning to the console (see below)

Example warning:

```console
$ hub pull-request
Warning: Local base branch differs from remote (got: developmen, want: development), to fix:
        git remote set-head origin --auto
```

I considered writing a test but the tests don't pass (even on `master`). Running macOS 10.13.3, Go 1.10, Git 2.16.2. The failing test output is available in [this gist](https://gist.github.com/mafredri/eaa34e30ac8022a348c19ab07edf4d95).

Running the tests also made changes to my personal git configuration (including alias modifications, not cool 😢). Lucky I've got it in version control.

Fixes #1538.

### DIFF
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -158,9 +158,22 @@ func pullRequest(cmd *Command, args *Args) {
 		flagPullRequestIssue = parsePullRequestIssueNumber(arg)
 	}
 
+	var defaultBase string
+	if headRepo, err := client.Repository(headProject); err == nil {
+		headProject.Owner = headRepo.Owner.Login
+		headProject.Name = headRepo.Name
+		defaultBase = headRepo.DefaultBranch
+	}
+
 	if base == "" {
-		masterBranch := localRepo.MasterBranch()
-		base = masterBranch.ShortName()
+		if masterBranch, ok := localRepo.RemoteMasterBranch(); ok {
+			base = masterBranch.ShortName()
+		} else {
+			base = defaultBase
+		}
+		if base != defaultBase {
+			fmt.Printf("Warning: Local base branch differs from remote (got: %s, want: %s), to fix:\n\tgit remote set-head origin --auto\n", base, defaultBase)
+		}
 	}
 
 	if head == "" && trackedBranch != nil {
@@ -183,11 +196,6 @@ func pullRequest(cmd *Command, args *Args) {
 		} else {
 			head = trackedBranch.ShortName()
 		}
-	}
-
-	if headRepo, err := client.Repository(headProject); err == nil {
-		headProject.Owner = headRepo.Owner.Login
-		headProject.Name = headRepo.Name
 	}
 
 	fullBase := fmt.Sprintf("%s:%s", baseProject.Owner, base)
--- a/github/localrepo.go
+++ b/github/localrepo.go
@@ -103,12 +103,7 @@ func (r *GitHubRepo) CurrentBranch() (branch *Branch, err error) {
 }
 
 func (r *GitHubRepo) MasterBranch() (branch *Branch) {
-	origin, e := r.RemoteByName("origin")
-	var name string
-	if e == nil {
-		name, _ = git.BranchAtRef("refs", "remotes", origin.Name, "HEAD")
-	}
-
+	name := r.remoteMasterBranchName()
 	if name == "" {
 		name = "refs/heads/master"
 	}
@@ -116,6 +111,23 @@ func (r *GitHubRepo) MasterBranch() (branch *Branch) {
 	branch = &Branch{r, name}
 
 	return
+}
+
+func (r *GitHubRepo) RemoteMasterBranch() (b *Branch, ok bool) {
+	name := r.remoteMasterBranchName()
+	if name == "" {
+		return nil, false
+	}
+	return &Branch{Repo: r, Name: name}, true
+}
+
+func (r *GitHubRepo) remoteMasterBranchName() string {
+	origin, e := r.RemoteByName("origin")
+	var name string
+	if e == nil {
+		name, _ = git.BranchAtRef("refs", "remotes", origin.Name, "HEAD")
+	}
+	return name
 }
 
 func (r *GitHubRepo) RemoteBranchAndProject(owner string, preferUpstream bool) (branch *Branch, project *Project, err error) {
--- a/github/client.go
+++ b/github/client.go
@@ -444,14 +444,15 @@ func (client *Client) FetchCIStatus(project *Project, sha string) (status *CISta
 }
 
 type Repository struct {
-	Name        string                 `json:"name"`
-	FullName    string                 `json:"full_name"`
-	Parent      *Repository            `json:"parent"`
-	Owner       *User                  `json:"owner"`
-	Private     bool                   `json:"private"`
-	HasWiki     bool                   `json:"has_wiki"`
-	Permissions *RepositoryPermissions `json:"permissions"`
-	HtmlUrl     string                 `json:"html_url"`
+	Name          string                 `json:"name"`
+	FullName      string                 `json:"full_name"`
+	Parent        *Repository            `json:"parent"`
+	Owner         *User                  `json:"owner"`
+	Private       bool                   `json:"private"`
+	HasWiki       bool                   `json:"has_wiki"`
+	Permissions   *RepositoryPermissions `json:"permissions"`
+	HtmlUrl       string                 `json:"html_url"`
+	DefaultBranch string                 `json:"default_branch"`
 }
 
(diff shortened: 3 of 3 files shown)


mafredri:
From the failing Travis test, I see that my assumption with regards to `default_branch` always being available is not true for the tests. Is this a bug in the test setup or something that should be handled?

mislav:
> I considered writing a test but the tests don't pass (even on `master`).

That's weird, since we have CI that verifies that tests pass on `master`. https://travis-ci.org/github/hub/branches

Could there be something about your location environment that our test suite doesn't properly insulate itself from?

> Running the tests also made changes to my personal git configuration (including alias modifications, not cool 😢). Lucky I've got it in version control so no harm done.

Oh no; this should't happen. Our tests change the environment variable `HOME` to a temporary location which should be completely separate from your usual `HOME` location. https://github.com/github/hub/blob/846f9944cd7b150f4b6e0f08aefb5b51b94419b7/features/support/env.rb#L38

Can you share with us how you ran the tests and which git config properties were changed?

mafredri:
> Could there be something about your location environment that our test suite doesn't properly insulate itself from?

Maybe? But I don't know what that would be. Testing isn't really documented, I ran `make test-all` and the output is in the gist posted above.

> Can you share with us how you ran the tests and which git config properties were changed?

Nothing special about how I ran the test. But I think the assumption about overriding `HOME` is insufficient, it only takes care of one use-case (when configuration is located in`~/.gitconfig`). I believe you also need to set (or unset) `XDG_CONFIG_HOME`, perhaps others. My configuration is located in `$XDG_CONFIG_HOME/git/config`. I can't recall the exact options that were changed, but it's not really relevant (some aliases were added, some overridden, etc).

mislav:
@mafredri Thanks for the tip; I've just pushed 839a5f3

commands/pull_request.go
```diff
@@ -158,9 +158,22 @@ func pullRequest(cmd *Command, args *Args) {
...
 	if base == "" {
-		masterBranch := localRepo.MasterBranch()
-		base = masterBranch.ShortName()
+		if masterBranch, ok := localRepo.RemoteMasterBranch(); ok {
+			base = masterBranch.ShortName()
+		} else {
+			base = defaultBase
+		}
+		if base != defaultBase {
+			fmt.Printf("Warning: Local base branch differs from remote (got: %s, want: %s), to fix:\n\tgit remote set-head origin --auto\n", base, defaultBase)
```
mislav: Could this print to stderr instead?

github/localrepo.go
```diff
@@ -118,6 +113,23 @@ func (r *GitHubRepo) MasterBranch() (branch *Branch) {
 	return
 }
 
+func (r *GitHubRepo) RemoteMasterBranch() (b *Branch, ok bool) {
```
mislav: 1. Could this simply return `error` instead of `ok bool`?
2. Why did the original `MasterBranch` needed to change? I don't see anything new in the updated logic. What was the rationale for adding `RemoteMasterBranch`?

mafredri: > 1. Could this simply return error instead of ok bool?

Sure, that would work as well. I guess the question is, what's the more correct signal?

> 2. Why did the original MasterBranch needed to change? I don't see anything new in the updated logic. What was the rationale for adding RemoteMasterBranch?

I wanted to differentiate between:

1. `git.BranchAtRef("refs", "remotes", origin.Name, "HEAD")` returns `refs/heads/master`
2. Above fails and name is set to `refs/heads/master` in `MasterBranch()`.

In the first case, when repo default branch differs, we show a warning. In the second case, we don't.

The new function was added to avoid changing behavior of `MasterBranch`, but I could of course either change the return signature of `MasterBranch` instead and update the code in all relevant locations or remove the fallback `refs/heads/master`. Thoughts?

commands/pull_request.go
mafredri: Sure, that's nicer, I'll fix.

github/localrepo.go
mislav: I see, thanks for the explanation. I'm fine with the current approach for now. I would love to see some tests!

mafredri:
@mislav thanks for pushing 839a5f38f221cd29e1bbf9c293d8b8639ca2c7e9, I tried it out but `make test-all` still makes changes to my git config 😕, here are some of the changes:

```diff
@@ -10,6 +12,8 @@

+       editor = $FOO \"${BAR}\"
+       commentchar = auto
 [alias]
@@ -62,3 +66,6 @@
        instructionFormat = (%an <%ae>) %s
+[hub]
+       test =
+       reportCrash = never

**Comment by mislav** [comment]
@mafredri Thanks for providing more info? Can you look at the output of `env` and share with us any environment configuration that could be affecting where git reads configuration from? We try to cancel out the `HOME` and `XDG_CONFIG_*` variables, but there might be something else. Do you have `GIT_CONFIG` set by any chance?

**Comment by mafredri** [comment]
AFAICT, there shouldn't be any other vars in my env that can affect this. So I tried making the following change:

```diff
diff --git a/features/support/env.rb b/features/support/env.rb
index 603bbad..9e891fd 100644
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -37,7 +37,7 @@ Before do
   # avoids reading from current user's "~/.gitconfig"
   set_env 'HOME', File.expand_path(File.join(current_dir, 'home'))
   # https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables
-  set_env 'XDG_CONFIG_HOME', nil
+  set_env 'XDG_CONFIG_HOME', File.expand_path(File.join(current_dir, 'home', '.config'))
   set_env 'XDG_CONFIG_DIRS', nil
   # used in fakebin/git
   set_env 'HUB_SYSTEM_GIT', system_git
```

As a result, the only change that is made to my git config is `hub.reportCrash = never`.

I believe this remaining change originates from [github/crash_report.go#L143-L149](https://github.com/github/hub/blob/98f153f3a639ceb871e5e2e61d3b294e58da430f/github/crash_report.go#L143-L149) invoked by [github/crash_report_test.go#L59](https://github.com/github/hub/blob/98f153f3a639ceb871e5e2e61d3b294e58da430f/github/crash_report_test.go#L59).

Tested by adding a panic to [github/crash_report_test.go#L38](https://github.com/github/hub/blob/98f153f3a639ceb871e5e2e61d3b294e58da430f/github/crash_report_test.go#L38) in which case `never` becomes `always` in my git config.

mislav:
@mafredri Thank you for providing extra information! Can you try f93043712d90c5608a8ad750586a3028cc353558 which I just pushed to master?

PR #1718, opened 2018-04-12
updated go version of delete repo

andreasbaumann:


github/client.go
```diff
@@ -247,6 +247,27 @@ func (client *Client) CreateRepository(project *Project, description, homepage s
 	return
 }
 
+func (client *Client) DeleteRepository(project *Project) (err error) {
+	var repoURL string
+	if project.Owner != client.Host.User {
```
mislav: This conditional seems unnecessary. This should work without any conditional:

```go
repoURL := fmt.Sprintf("/repos/%s/%s", project.Owner, project.Name)
```

github/client.go
```diff
@@ -247,6 +247,27 @@ func (client *Client) CreateRepository(project *Project, description, homepage s
...
+		repoURL = fmt.Sprintf("/repos/%s/%s", client.Host.User, project.Name)
+	}
+
+	api, err := client.simpleApi()
+	if err != nil {
+		return
+	}
+
+	res, err := api.Delete(repoURL)
+	if err = checkStatus(204, "deleting repository", res, err); err != nil {
```
mislav: The `if err != nil { return err }` conditional isn't necessary since the function already does `return err` in the end.

commands/delete.go
```diff
@@ -0,0 +1,118 @@
...
+	if strings.Contains(repoName, "/") {
+		split := strings.SplitN(repoName, "/", 2)
+		owner = split[0]
+		repoName = split[1]
+	}
+
+	project := github.NewProject(owner, repoName, host.Host)
+	gh := github.NewClient(project.Host)
+
+	if !gh.IsRepositoryExist(project) {
```
mislav: I don't think that checking for repository existing is necessary. Let's just try the delete operation and print the 404 message if the repository wasn't found.

commands/delete.go
```diff
@@ -0,0 +1,118 @@
...
+	"strings"
+
+	"github.com/github/hub/git"
+	"github.com/github/hub/github"
+	"github.com/github/hub/utils"
+)
+
+var cmdDelete = &Command{
+	Run:   delete,
+	Usage: "delete [[<ORGANIZATION>/]<NAME>]",
```
mislav: Instead of making `NAME` optional, how about requiring the argument? That would force the user to type the name of the repository to delete even if it's the current repository.

commands/delete.go
```diff
@@ -0,0 +1,118 @@
...
+	project := github.NewProject(owner, repoName, host.Host)
+	gh := github.NewClient(project.Host)
+
+	if !gh.IsRepositoryExist(project) {
+		fmt.Println("No such repository")
+		args.NoForward()
+		return
+	}
+
+	if os.Getenv("HUB_UNSAFE_DELETE") == "" {
```
mislav: Instead of supplying an environment variable to skip the prompt, can we add a flag such as `hub delete -y` instead?

commands/delete.go
```diff
@@ -0,0 +1,118 @@
...
+		fmt.Printf("Repository '%s' exists. Really delete it (y/N)?", repoName)
```
mislav: Instead of having the user first choose `y` and then type the name of the repository, how about asking them to do just one thing? I.e. just type the name of the repository in the first place, or just choose `y` and be done with it. But not both.

commands/delete.go
```diff
@@ -0,0 +1,118 @@
...
+
+		fmt.Println("Please write the name of the repository again (this operation can not be undone!): ")
+		_, err = fmt.Scan(&s)
+		if err != nil {
+			fmt.Println(err)
+			args.NoForward()
+			return
+		}
+		s = strings.TrimSpace(s)
+		if s != repoName {
```
mislav: What if `repoName` contains uppercase characters?

commands/delete.go
andreasbaumann: Yep, done.

commands/delete.go
andreasbaumann: Agreed, delete the current working repo is a little bit dangerous anyway. :-)

commands/delete.go
andreasbaumann: I was trying to mimick the becaviour in the web interface of github, but yes, just a 'y' is enough.
I was super-paranois while implementing the delete because there were strong oppositions to
give people a dangerous tool. Well, for the same reason we would have to remove 'rm' from 
'coreutils'. :-)

commands/delete.go
andreasbaumann: Obsolete: I'm not checking the reponame anymore.

commands/delete.go
andreasbaumann: Looks no like:

```
shell> hub delete xxx
Really delete repository 'xxx'(y/N)?y
Error deleting repository: Not Found (HTTP 404)
Not Found
```

I removed the check.

andreasbaumann:
Just forgot to test what happens if the delete_repo right is not in the access token:

```
abaumann@eeepc:~> hub delete cpp_starter_project
Really delete repository 'cpp_starter_project'(y/N)?y
Error deleting repository: Unauthorized (HTTP 401)
Bad credentials
```

Is this enough?

andreasbaumann:
Just forgot to test what happens if the delete_repo right is not in the access token:

```
shell:~> hub delete cpp_starter_project
Really delete repository 'cpp_starter_project'(y/N)?y
Error deleting repository: Forbidden (HTTP 403)
Must have admin rights to  @Repository.
```

Is this enough?

commands/delete.go
```diff
@@ -34,18 +36,22 @@ hub-init(1), hub(1)
...
 	CmdRunner.Use(cmdDelete)
 }
 
 func delete(command *Command, args *Args) {
 	var repoName string
 	if args.IsParamsEmpty() {
-		dirName, err := git.WorkdirName()
-		utils.Check(err)
-		repoName = github.SanitizeProjectName(dirName)
+		ui.Errorln("Expecting name of reposiroty to delete")
```
mislav: 1. Typo in `repository`
2. It would be great if the params empty case simply printed out the command's synopsis on stderr and exited with a non-zero status. https://github.com/github/hub/blob/da7348a83b41bca0b27b58ac356b34c6fc536c83/commands/commands.go#L95-L107

mislav:
> Just forgot to test what happens if the delete_repo right is not in the access token:

Could we perhaps detect a 403 and additionally print out something like:

```
Please edit the token used for hub at https://github.com/settings/tokens
and verify that the `delete_repo` scope is enabled.
```

commands/delete.go
```diff
@@ -2,22 +2,24 @@ package commands
...
 
 var cmdDelete = &Command{
 	Run:   delete,
-	Usage: "delete [[<ORGANIZATION>/]<NAME>]",
+	Usage: "delete [-y] [<ORGANIZATION>/]<NAME>",
 	Long: `Delete an existing repository on GitHub.
 
 ## Options:
 
+	-y
```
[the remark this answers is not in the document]
andreasbaumann: done

commands/delete.go
andreasbaumann: ok. done

mislav:
Thank you @andreasbaumann! I've added additional fixes on top of your branch.

Issue #1651, opened 2018-01-19
Expand <user/repo> when using git pull


ritiek:
Just like `hub clone <user/repo>` expands to `git clone git://github.com/<user/repo>`, it would be nice to have a similar thing for `hub pull <user/repo>`.

Tmplt:
Similar syntax would be nice for `hub checkout` when checking pull requests out. E.g. `hub checkout <user/repo>/pull/<pull-request>` instead of a URL, but this might be fit for another issue.

Or does this break compatibility with git?

mislav:
Thank you for the feature idea! I gave it a thought since you've opened this but I don't think it's worth to invest into creating a wrapper for yet another git command. `hub pull OWNER/REPO` is not a syntax that I imagine being used very often, except for pulling from forks, in which case there is already the `hub fetch OWNER && git merge OWNER/BRANCH` combo available. Also, lately with hub we want to focus more on creating extra commands (such as `pull-request`, `release`, and `issue`) and less on extending existing git commands (such as `clone` and `checkout`) since those were historically harder to maintain.

PR #1719, opened 2018-04-12
Expand <user/repo> when using hub pull

ritiek:
Addresses #1651. Since I pull stuff from GitHub quite often, this feature is useful to me.

This PR is kind of a copy of [clone.go](https://github.com/github/hub/blob/1e5bb959dcc11f515d8865ece27c5fc9dd3b2d39/commands/clone.go).

Example:
```
$ git clone https://github.com/github/hub
$ cd hub
$ git checkout code-coverage
$ hub pull github/hub
```

Feel free to close this PR if there is a more modular way to implement this!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ HELP_EXT = \
 	share/man/man1/hub-checkout.1 \
 	share/man/man1/hub-cherry-pick.1 \
 	share/man/man1/hub-clone.1 \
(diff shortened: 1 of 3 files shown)


mislav:
Thank you! Code looks good but this is not a feature we want right now. https://github.com/github/hub/issues/1651#issuecomment-395332443

Issue #1343, opened 2016-11-21
Support creation of milestones


cbeams:
Please consider a `milestone` subcommand for `hub` that would allow users to create and modify milestones. Looking at #1232, and searching through open issues, I haven't seen anything to this effect on the roadmap. Note that I've requested this same feature be implemented in `ghi` at https://github.com/stephencelis/ghi/issues/321.

mislav:
Thanks for the suggestion. Let's see how many users are interested in this feature— hopefully they will add their 👍 to your idea over time.

This isn't on my own personal roadmap for hub, since there are many tasks that have much higher priority for the next release, and I believe that project management features of GitHub (milestones, projects, editing labels) are best performed in the web interface rather than CLI.

cbeams:
Thanks for the quick response. My issue is that I manage a large number of fine-grained repositories that all share milestones and labels, so I have a real need to automate this stuff. I script things with `ghi` and `hub` where I can, and drop down to to the GitHub API where necessary. Milestone creation is the only thing that I have to resort to the API for now, and that's why I ask.

mislav:
Thanks for explaining your use-case!

For the longest time now, I've had this plan of creating a low-level `hub api` command that will let you make manual API requests for all functionality that is currently not covered with the rest of hub's built-in feature set. The `api` command would take care of authentication and JSON serialization for you, making it very easy to POST something to the `/milestones` endpoint of the current project. See https://github.com/github/hub/issues/1203 for example of another use-case.

However, considering that we're moving towards GraphQL at GitHub for the next generation of our API, I'm reluctant to ship the `api` command that would rely on the API v3 (our current, REST API). I would much rather explore how this could be possible using GraphQL.

In any case, there will always be people like you who want to do a thing with hub that is currently not implemented via any command, and it would be nice if we had a low-level scripting interface that would allow a more flexible access. If such a command existed right now, you could start using it right away in your scripts instead of waiting for a future release of hub that would implement some sort of `milestone create` command.

mislav:
After some consideration, I have decided that—for now—this is best implemented in user scripts, rather than maintained in hub core. Now that there is a new `hub api` command, I will close this feature request and leave implementing this as an exercise to the reader. https://github.com/github/hub/releases/tag/v2.8.3 https://github.com/github/hub/pull/2016

Thank you for suggesting!

Issue #1593, opened 2017-10-10
Milesones


luisbrandao:
Does hub have support to create milestones? If not, can this be added?

mislav:
No, milestone creation right now isn't supported nor is it planned. If we added milestone creation then we will have to add milestone editing, deletion and closing as well, and I consider this set of features to be better suited for the web interface right now, since I don't imagine the majority of our users ever managing milestones from the command-line. Thank you for your feature request, though!

PR #1720, opened 2018-04-12
WIP milestones

seemethere:
We utilize milestones pretty heavily internally and have had to build one off utilities in order to be able to do it automatically, it'd be great if we could just do this in `hub`.

The goal is basically to provide functionality for all of the API calls found at https://developer.github.com/v3/issues/milestones/

Relates to:

* https://github.com/github/hub/issues/1343

Work to be done:
* Add commands to:
  * [ ] create milestones
  * [ ] list milestones
  * [ ] update milestones
  * [ ] delete milestones
  * [ ] get a single milestone 

UX still needs to be fleshed out but I this right now is just a starting point

If this functionality is not desired let me know and I'll happily close the PR.

Signed-off-by: Eli Uriegas <eli.uriegas@docker.com>

### DIFF
--- a/commands/milestones.go
+++ b/commands/milestones.go
@@ -1,0 +1,71 @@
+package commands
+
+import (
+	"fmt"
+
+	"github.com/github/hub/github"
+	"github.com/github/hub/utils"
+)
+
+var (
+	cmdMilestone = &Command{
+		// TODO: Add usage information, eventually get rid of `Key: "milestone",`
+		Key: "milestone",
+		Run: listMilestones,
+	}
+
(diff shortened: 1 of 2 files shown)


mislav:
Hi, thanks for proposing and starting work on this! While I'm certainly feeling confident that you can fully flesh out this feature, milestones is not something we want to invest in hub right now, since there are many other features on our roadmap that take priority. Because every added feature requires further maintenance for many years to come, lately I'm being conservative with adding new features lately. See https://github.com/github/hub/issues/1593#issuecomment-336528950

PR #1727, opened 2018-04-24
Set slice length to known number of args

tcarrio:
*This is a majorly inconsequential change and saw this when I was looking through the source on my phone.*

The arguments slice is made with an initial length of 0 rather than the number of arguments that will be given, so it will be reprovisioned at least once by copying contents from the underlying array to a new one. This is work that doesn't need to be done, this PR makes the the args slice's initial length the exact number of arguments that will be appending to the slice. No array copies needed!

I could not find a contribution guide from the README, please let me know if there are any more steps through which this needs to go.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -97,7 +97,7 @@ func New(cmd string) *Cmd {
 	utils.Check(err)
 
 	name := cmds[0]
-	args := make([]string, 0)
+	args := make([]string, len(cmds)-1)
 	for _, arg := range cmds[1:] {
 		args = append(args, arg)
 	}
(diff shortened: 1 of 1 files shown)


tcarrio:
I'll have to look at the cmd_test, I didn't expect this to cause any issues.

tcarrio:
I'm so very dumb. I'll fix this later tonight.

tcarrio:
So I fixed this to create an array of the same size as the number of arguments, then used copy duplicate values to it. This should remove any dynamic resizing of the underlying array.

mislav:
Thank you!

PR #1740, opened 2018-05-16
Allow a user to provide a token instead of a password

llimllib:
* Before writing a test, I'd like to know if this approach seems sane to you?
* I've verified that it works locally and will figure out how to write a test if the general idea of this change seems sensible to you. (Any guidance on what I'd need to do appreciated, of course)

cc @mislav 

"fixes" #1067, although that ticket is questionably a bug, this is more of a feature

### DIFF
--- a/github/client.go
+++ b/github/client.go
@@ -683,13 +683,31 @@ type AuthorizationEntry struct {
 	Token string `json:"token"`
 }
 
+// Test if the provided password is actually valid oauth token
+func isPasswordAToken(api *simpleClient, user, password string) bool {
+	api.PrepareRequest = func(req *http.Request) {
+		req.Header.Set("Authorization", "token "+password)
+	}
+
+	res, _ := api.Get("user")
+	if res != nil && res.StatusCode == 200 {
+		return true
+	}
+	return false
+}
+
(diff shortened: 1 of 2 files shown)


github/client.go
```diff
@@ -723,6 +723,20 @@ func (client *Client) FindOrCreateToken(user, password, twoFactorCode string) (t
 		} else {
 			errInfo, e := res.ErrorInfo()
 			if e == nil {
+				// If we got a un/pw message the user may have provided an
+				// access token rather than a password. Test if it's valid,
+				// and return it if it is
+				if strings.TrimSpace(errInfo.Message) == "This API can only be accessed with username and password Basic Auth" {
```
mislav: Best not to tie this behavior to the exact language in the error message of the server response. That message could change.

How about a totally different approach? If a password of 40 or more characters was pasted, first check directly if it's a valid token (by requesting `api.github.com/user` or similar), otherwise fall back to the basic auth dance?

llimllib: I actually was going to check on any 40-character password until I noticed that github already throws this particular error only when you give it an oauth token. But you're right, it does suck to depend on a string, I wish there was a more reliable message in this case!

I'll change it to your idea.

mislav:
Thank you! This is excellent.

PR #1752, opened 2018-05-28
pull-request: Hide Signed-off-by

### LINKED ISSUE #1724: hub pull-request: Remove the "Signed-off-by: NAME <email@email.foo)" line
I frequently use `git commit -S && git push my-fork && hub pull-request` flow.

However the `hub pull-request` command copies all of the commit message to the pull request message editor. This includes the line like:

        Signed-off-by: Ahmet Alp Balkan <ahmetb@example.com>

And since my email is in there, it accidentally leaks to the PR description every now and then which makes it more susceptible to spam.

Would you be OK with having code that excludes the `^Signed-off-by\:` lines from the suggested PR description? I believe the PR description only goes to the GitHub UI and doesn't have much to do with the actual commit object.

ahmetb:
Fixes #1724.

This patch omits the "Signed-off-by:" line coming from pull requests created
out of single commit messages. This way the user does not need to manually
remove their email address from the PR description.

Not introducing a flag for this as it applies to only one case (single-commit
diff between heads, as well as no -m/-f/-i specified).

Meta: sending this PR with the patched build, my Signed-off-by line is removed
automatically from the PR message.

### DIFF
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -252,6 +252,9 @@ of text is the title and the rest is the description.`, fullBase, fullHead))
 		if len(commits) == 1 {
 			message, err = git.Show(commits[0])
 			utils.Check(err)
+
+			re := regexp.MustCompile(`(?m)\n^Signed-off-by:\s.*$`)
+			message = re.ReplaceAllString(message, "")
 		} else if len(commits) > 1 {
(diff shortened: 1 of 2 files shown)


mislav:
Thank you! 🙇

PR #1766, opened 2018-06-06
Add MacPorts installation method

grimreaper:


mislav:
Thanks, but I would rather not encourage the use of Macports. It causes compatibility issues with some other open source software that I maintain and as a result I don't have a good feeling about it. Macports users will just have to guess that hub is available as `hub` port, which is not too hard to guess.

grimreaper:
It's funny you say that because homebrew is generally known as the incompatible option. As the pithy saying goes: if you prefer popularity, use homebrew; correctness use macports.

Out of curiosity, to avoid FUD, what software have you seen that macports caused issues for?

mislav:
ruby-build is a tool to compile Ruby versions and [ocassional bug reports](https://github.com/rbenv/ruby-build/issues?utf8=%E2%9C%93&q=is%3Aissue+macports) with the build process whose cause was difficult to track down turned out to be caused my Macports-installed software.

Issue #1772, opened 2018-06-10
Avoid to pass `--cmd` to non-vim-s vi-s


iamleot:
Hello to the entire hub community,
AFAIK not every `vi`-s support the `--cmd` option and the current arguments
appended in [openTextEditor()](https://github.com/github/hub/blob/0ef312ec03553ba858da74fed2b4b98e50769cca/github/editor.go#L130).

At least with [vi(1)](http://man.NetBSD.org/cgi-bin/man-cgi?vi+1) on NetBSD doing `hub pull-request` fails with:

````
% hub pull-request
vi: unknown option -- -
usage: ex [-eFRrSsv] [-c command] [-t tag] [-w size] [file ...]
usage: vi [-eFlRrSv] [-c command] [-t tag] [-w size] [file ...]
error using text editor for pull request message
````

At the moment we've patched that in the [pkgsrc](https://www.pkgsrc.org/) [net/hub](https://pkgsrc.se/net/hub) package changing the regular expression in [openTextEditor()](https://github.com/github/hub/blob/0ef312ec03553ba858da74fed2b4b98e50769cca/github/editor.go#L132) to not match `vi` (i.e. `\b(?:[gm]?vim)(?:\.exe)?$`).

Could such patch be acceptable?
Please let me know and I will try to fill a pull request if needed.

Thank you!

mislav:
We would accept a patch to stop matching `vi` and just match `vim` instead! Thank you

Issue #1602, opened 2017-10-16
hub doesn't pass editor options to vi


patheticpat:
I just stumbled across this regex that detects if the users editor is some variant of vi:
https://github.com/github/hub/blob/cbcc349e39c359ea2bd1f3f0d2ad56463750a905/github/editor.go#L114
Unless I am missing something obvious, shouldn't this rather be something like
```golang
r := regexp.MustCompile(`\b(?:[gm]?vim|vi)\b`)
```
in order to also match vi, but not gvi or mvi?

mislav:
You're right. The `[m]` expression makes no sense since it would only ever match the literal `m` character. Maybe this was supposed to be `(m)?`? Care to submit a PR for this?

patheticpat:
Sure.

PR #1775, opened 2018-06-11
Do not pass `--cmd' to vi in openTextEditor()

iamleot:
As discussed in issue #1772 do not pass `--cmd` arguments to `vi` because they
are specific to vim.

(I have not deleted the non-grouping parentheses in the RE to help
readability.)

### DIFF
--- a/github/editor.go
+++ b/github/editor.go
@@ -129,7 +129,7 @@ func (e *Editor) readContent() (content []byte, err error) {
 
 func openTextEditor(program, file string) error {
 	editCmd := cmd.New(program)
-	r := regexp.MustCompile(`\b(?:[gm]?vim|vi)(?:\.exe)?$`)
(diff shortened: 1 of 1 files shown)


mislav:
This would undo #1603 prompted by #1602. @patheticpat: are you fine with that? It looks like not all vi variants support `--cmd`.

iamleot:
Hello Mislav,
(sorry for the delay)

Mislav Marohnić writes:
> This would undo #1603 prompted by #1602. @patheticpat: are you fine with that? It looks like not all vi variants support `--cmd`.

Whoops, nice catch!

Back to the original issue (I forgot to cite that) but also POSIX
vi(1) (<http://pubs.opengroup.org/onlinepubs/9699919799/utilities/vi.html>)
does not support any `--cmd' option and the command passed via
`--cmd' seems specific to just vim (and possible forks).

(But I guess that on several systems `vi' is just an alias to `vim'
and hence everything will work)

PR #1780, opened 2018-06-15
Migrate from gopkg.in/yaml.v1 to gopkg.in/yaml.v2

anthonyfok:
gopkg.in/yaml.v1 is no longer maintained since September 2014,
and was removed from Debian in July 2016 due to an unresolved RC bug,
see https://bugs.debian.org/832414.

Thankfully, yaml.v2 is largely compatible with yaml.v1, and a quick
check with https://blog.labix.org/2014/09/22/announcing-yaml-v2-for-go
reveals no necessary code change except an update of the import path.


Gopkg.toml
```diff
@@ -35,8 +35,8 @@
   name = "golang.org/x/crypto"
 
 [[constraint]]
-  branch = "v1"
-  name = "gopkg.in/yaml.v1"
```
mislav: To skip the middle-man, could we simply reference the `v2` branch of `github.com/go-yaml/yaml`?

anthonyfok: I suppose it is possible, but not recommended.  The README.md at https://github.com/go-yaml/yaml specifically says:

> The import path for the package is *gopkg.in/yaml.v2*.

And I haven't seen any packages deviates from the official import path.

For reference, Hugo is also using the `gopkg.in/yaml.v2` as import path, see https://github.com/gohugoio/hugo/blob/master/Gopkg.toml#L131-L133

But of course, hopefully this will probably become a moot point when vgo arrives, and perhaps the @go-yaml will rename the import path in the future.  Just my random speculation.  ;-)

Many thanks!

PR #1797, opened 2018-07-06
Fix example for fetch.

ehuss:


mislav:
Good catch thank you!

faronangus:
great i  think that worked....
 
               Thank You  xfaronang88

PR #1799, opened 2018-07-06
Add documentation link to Readme

joelanman:
I found it quite hard to find the documentation from the GitHub repo. I think the only link is in the header, and from that page I didn't find the link to documentation easily either. Might be worth adding a link in the Readme?


mislav:
Thank you!

Issue #1377, opened 2017-01-09
UX: hub pull-request strips Markdown headers


mgedmin:
I was creating a pull request for Ansible by running `hub pull-request`.  It downloaded Ansible's [PULL_REQUEST_TEMPLATE.md](https://github.com/ansible/ansible/blob/devel/.github/PULL_REQUEST_TEMPLATE.md) very felpfully and spawned Vim for me.  I edited the PR description, which looked something like

```
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
blah blah

##### ANSIBLE VERSION
ver ver

##### SUMMARY

bug bug bug
```

and then I quit vim and looked at the PR that `hub` created for me.

I discovered that `hub` had stripped off all the Markdown headers, resulting in a nonsensical-looking

>  - Bugfix Pull Request
>
> blah blah
> ver ver
> bug bug bug

This is rather unfriendly to projects that use the `#` header syntax in their `.github/PULL_REQUEST_TEMPLATE.md`.

(I was using `hub` version 2.3.0-pre4.)

mislav:
This is known, but thank you for opening the issue because I feel this is something that I need to fix. The `#` character is the default comment character for various git messages, such as commit subject & body, but this default doesn't work as well with Markdown. Maybe `hub pull-request` and `hub issue create` should use a different default comment character.

pirj:
@mislav I think [changing the default git comment char](http://stackoverflow.com/a/14931661/202914) may resolve this. It does not seem to work with an `auto` setting, however, it plays nicely with `core.commentchar=-`.

mgedmin:
FWIW `-` is often used for Markdown lists, and it'd be a shame it `hub` stripped those.  `;` sounds safer.

pirj:
You may prefix `-` in lists with a space and it does not count as a comment then, but in general yes, `;` is a better choice.

jasonkarns:
Another workaround is to use `====` and `----` underlining the headers, since that is also valid markdown for headers.

```md
A First Level Header
====================

A Second Level Header
---------------------
```

cben:
I've seen some projects switch from `##` in templates to `====` and `-----` due to this, but not all.  E.g. the OP example of Ansible wants tiny level-5 headers to can't use `-----`.

- hub could invoke git with different commentchar just for that command (to not mess with user settings otherwise)?
  wait, git is not used to open PRs!  Does hub just emulate same behavior?  If so we fully control it?

[...]

PR #1812, opened 2018-07-18
Update README.md

fjlopezs:
Added workaround for make pull requests.


mislav:
Thank you, but this workaround will be obsolete after we implement the `scissors` cleanup mode which will fix the problem with markdown headers getting stripped. Follow https://github.com/github/hub/issues/1377#issuecomment-394760333 for updates

PR #1814, opened 2018-07-19
fea(pull_request): short hand for `--no-edit`

JensRantil:
Happens so many times I am submitting a tiny pull request something in a
single commit that a shorthand parameter would be very useful.

### DIFF
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -17,7 +17,7 @@ import (
 var cmdPullRequest = &Command{
 	Run: pullRequest,
 	Usage: `
-pull-request [-focp] [-b <BASE>] [-h <HEAD>] [-r <REVIEWERS> ] [-a <ASSIGNEES>] [-M <MILESTONE>] [-l <LABELS>]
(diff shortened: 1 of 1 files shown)


JensRantil:
I'm a bit unsure why the tests fail for one go version but not the others.

mislav:
Thank you for this idea! I'm not so sure whether this would be useful to the majority of hub users. Most often, I think people want to review what they're submitting and write more for the body of the PR (people mostly leave bodies of commit messages blank).

The `--no-edit` flag was added to support those workflows where you explicitly choose to use only the commit message for PR title and body, but I don't want to make this flag easier to use through a shorthand. I think that, if you really use this often, you can create an alias: `alias prn='hub pull-request --no-edit`. What do you think?

> I'm a bit unsure why the tests fail for one go version but not the others.

That was an intermittent test failure; not your fault. I've restarted the job.

JensRantil:
I'm fine with making this an alias. Future visitors: Please ping in thread if you'd find this useful!

PR #1825, opened 2018-08-06
Update link to code of conduct

anand-sundaram-zocdoc:
The contributing doc mentions a code of conduct but this links to a nonexistent URL that used to host a code of conduct that has since been abandoned. I updated it to point to the in-repo code of conduct which seems to be kept up to date.


mislav:
Thank you!

PR #1838, opened 2018-08-18
Create '--no-track' option for hub pull-request

HaraldNordgren:
In my daily work, I always run

    git branch -u origin/master

on my feature branches to keep track of the master branch which I will eventually merge in to. It bothers me that `hub` resets my tracking branch after creating the PR, so that I have to re-run `git branch -u origin/master`.

We could also discuss changing the default behavior of hub to opt-in to `--set-upstream` rather than to opt-out. But for now I am happy as long as I get this option, so that Hub can co-operate smoothly with my workflow.

### DIFF
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -54,6 +54,9 @@ pull-request -i <ISSUE>
 	-p, --push
 		Push the current branch to <HEAD> before creating the pull request.
 
+	--no-track
+		When using "--push", do not change the upstream tracking branch.
+
 	-b, --base <BASE>
 		The base branch in "[OWNER:]BRANCH" format. Defaults to the default branch
 		(usually "master").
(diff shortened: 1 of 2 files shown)


HaraldNordgren:
@mislav Merge this?

HaraldNordgren:
Thanks for you response!

I think an argument for adding this is that `--set-upstream` is not the default behavior of Git. I would generally expect Hub to be as transparent as possible with its interaction with git. Which is why I would optimally see this being an opt-in rather than an opt-out.

Pushing the branch manually like

    git push origin $(git rev-parse --abbrev-ref HEAD) && hub pull-request

kind of defeats the purpose of using a tool, wouldn't you agree? 😊

mislav:
Yeah I see your point. I'll think about removing `--set-upstream` in a future release of hub that might depart from perfect backwards-compatibility.

Protip: this also works in git for pushing the same-named branch: `git push origin HEAD`

HaraldNordgren:
@mislav Merge this or close it please :)

PR #1845, opened 2018-08-23
Add unix socket support

apakulov-stripe:
Bringing **Unix domain socket** support to the `Hub`.

If `unix_socket` configuration is available for a given host in `~/.config/hub`, it will configure `http.Transport` that will work for `http` and `https` over **Unix domain socket**

### DIFF
--- a/fixtures/test_configs.go
+++ b/fixtures/test_configs.go
@@ -21,6 +21,20 @@ func SetupTomlTestConfig() *TestConfigs {
   host = "github.com"
   user = "jingweno"
   access_token = "123"
+  protocol = "http"`
+	ioutil.WriteFile(file.Name(), []byte(content), os.ModePerm)
+	os.Setenv("HUB_CONFIG", file.Name())
+
+	return &TestConfigs{file.Name()}
+}
+
+func SetupTomlTestConfigWithUnixSocket() *TestConfigs {
+	file, _ := ioutil.TempFile("", "test-gh-config-")
+
+	content := `[[hosts]]
+  host = "github.com"
+  user = "jingweno"
+  access_token = "123"
   protocol = "http"
   unix_socket = "/tmp/go.sock"`
 	ioutil.WriteFile(file.Name(), []byte(content), os.ModePerm)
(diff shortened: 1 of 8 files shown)


fixtures/test_configs.go
```diff
@@ -21,7 +21,8 @@ func SetupTomlTestConfig() *TestConfigs {
   host = "github.com"
   user = "jingweno"
   access_token = "123"
-  protocol = "http"`
+  protocol = "http"
+  unix_socket = "/tmp/go.sock"`
```
mislav: This adds `unix_socket` to default test configs. Is this desired, or should we create separate test cases which exercise the unix socket functionality?

github/client.go
```diff
@@ -829,7 +829,8 @@ func (client *Client) simpleApi() (c *simpleClient, err error) {
 }
 
 func (client *Client) apiClient() *simpleClient {
-	httpClient := newHttpClient(os.Getenv("HUB_TEST_HOST"), os.Getenv("HUB_VERBOSE") != "")
+	unixSocket := os.ExpandEnv(client.Host.UnixSocket)
```
mislav: Why is `ExpandEnv` necessary here?

github/config_service_test.go
```diff
@@ -74,7 +77,8 @@ func TestConfigService_TomlSave(t *testing.T) {
   host = "github.com"
   user = "jingweno"
   access_token = "123"
-  protocol = "https"`
+  protocol = "https"
+  unix_socket = "/tmp/go.sock"`
```
mislav: These tests also congfigure `unix_socket`. It would be good if some tests didn't have this configuration, so they could keep exercising old logic.

github/client.go
apakulov-stripe: It resolves\expands a path with environment variables, e.g.: `$HOME = /Users/apakulov`
If we have a value `$HOME/go.sock` it should result into `/Users/apakulov/go.sock`

github/config_service_test.go
apakulov-stripe: Split into separated tests in 0eb76ff

fixtures/test_configs.go
apakulov-stripe: Added a separate test config in 0eb76ff

apakulov-stripe:
Thank you for the quick review turnaround.  
Unix domain socket will primarily be useful for other GHE users in case they have extra security layer. 
We're running a sidecar-proxy that exposes `unix-socket` to connect to GHE.

apakulov-stripe:
@mislav yeah, I was able to communicate to our internal GHE using this branch.

apakulov-stripe:
@mislav Any ETAs on when we can get it released?

apakulov-stripe:
@mislav gentle ping :)

apakulov-stripe:
Hello @mislav, do you have any updates?

mislav:
Sorry for the wait!

PR #1850, opened 2018-08-25
Add MacPorts installation instructions

ylluminarious:
Just thought it would be nice to let people know that `hub` is installable via MacPorts as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,17 @@ git version 1.7.6
 hub version 2.2.3
 ```
 
+### MacPorts
+
+`hub` can also be installed through [MacPorts](https://www.macports.org/):
+
+```sh
+$ sudo port install hub
+$ hub version
+git version 2.18.0
+hub version 2.5.0
+```
+
 #### Windows
 
 `hub` can be installed through [Scoop](http://scoop.sh/) on Windows:


mislav:
https://github.com/github/hub/pull/1766#issuecomment-395211543
https://github.com/github/hub/pull/1608#issuecomment-339950502
https://github.com/github/hub/pull/276#issuecomment-11842559
https://github.com/github/hub/pull/211#issuecomment-6999507

ylluminarious:
@mislav That's a shame... I don't see how adding instructions could hurt anything. Everyone knows that it's best not to mix package managers. You leave MacPorts users with the impression that there is no way to install on macOS except via Homebrew. Btw, I'm a long-time Homebrew user and I've found MacPorts to be superior in many ways. I do not think that your bias towards Homebrew should prevent the addition of objective instructions for a package manager which has very valid use-cases and which is superior to Homebrew in many respects.

ylluminate:
Yes, @ylluminarious is exactly right. We're really finding MacPorts to be much superior to Homebrew now and we've all moved or are in the process of moving over to it. Homebrew is definitely not the shining star that folks want to portray it to be. Definitely need to facilitate this @mislav.

mislav:
I hear you, but I do prefer Homebrew as installation method on macOS, and the omission of MacPorts is intentional and reflects that. macOS users do have other installation options: they can download a precompiled binary for Darwin, they can compile from source themselves and, if they use MacPorts, they can use its command-line API to search for a formula related to this project.

I maintain the Homebrew formula for hub by keeping it up to date immediately upon every release of hub. The port for hub https://github.com/macports/macports-ports/blob/master/devel/hub/Portfile is unofficial and, depending on the responsiveness of the maintainer, can fall out of date. We've seen that happening with Chocolatey package manager for Windows and with community-maintained PPAs for Ubuntu/Debian.

This decision is not blocking MacPorts users from installing hub. It's simply avoiding endorsing MacPorts as installation method in general.

ylluminate:
<sigh> feels collusive and like "good ol' boy" syndrome to me at this point vs giving other options fair shakes. We're see too many cases of pandering to preference that's not based on actual superiority and it's creating a very negative feeling towards GitHub and others doing these things. We've even been moving off of GitHub to GitLab because of some of this behavior.  

Thank you, though, for explaining some of your thoughts on the matter.  Definitely not trying to be negative in an unfounded way here.  Things just aren't adding up, especially when others seem to be willing to pitch in.

mislav:
Sure, thank you for providing your point of view! Since this isn't the first request for this kind of change to our installation docs, it gives me some bit of an insight into MacPorts is still being used more than I thought.

PR #1855, opened 2018-08-28
Prevent setting a public upstream when creating a private repo.

### LINKED ISSUE #1854: `hub create -p` can unexpectedly set the upstream to a public repo
When I use `hub create -p` (specifying to create a private repo), there should be no situation where hub sets the remote to a public repository. This behaviour makes it trivially easy to accidently push private code to a new public repository.

In the event that a repo of the same name already exists, and is public, `hub` will detect the existing repo and set the upstream. *It gives no clear indication that the upstream registered is not a private repository*.

For example, the following results in a public repo being set as `origin` as the result of a command to create a private repository:
```
test ❯ git init
Initialized empty Git repository in /workspace/env/test/.git/
test (master) ❯ hub create
Updating origin
https://github.com/nrmitchi/test
test (master) ❯ rm -rf .git
test ❯ git init
Initialized empty Git repository in /workspace/env/test/.git/
test (master) ❯ hub create -p
Existing repository detected. Updating git remote
Updating origin
https://github.com/nrmitchi/test
```

While there is a message that an existing repository was found, it seems to me that `hub` should exit with a failure if the named repository already exists, but doesn't match the privacy settings that were specified.

nrmitchi:
Fixes #1854

### DIFF
--- a/commands/create.go
+++ b/commands/create.go
@@ -109,9 +109,9 @@ func create(command *Command, args *Args) {
 	project := github.NewProject(owner, newRepoName, host.Host)
 	gh := github.NewClient(project.Host)
 
-	repo, _ := gh.Repository(project)
+	repo, err := gh.Repository(project)
 
-	if repo != nil {
+	if err == nil {
 		if !repo.Private && flagCreatePrivate {
 			err = fmt.Errorf("Repository '%s' already exists and is public", repo.FullName)
 			utils.Check(err)
--- a/commands/create.go
+++ b/commands/create.go
@@ -110,7 +110,18 @@ func create(command *Command, args *Args) {
 	gh := github.NewClient(project.Host)
 
 	if gh.IsRepositoryExist(project) {
-		ui.Errorln("Existing repository detected. Updating git remote")
+
+		repo, err := gh.Repository(project)
+		if err != nil {
+			utils.Check(err)
+		}
+
+		if !repo.Private && flagCreatePrivate {
+			err = fmt.Errorf("Existing public repository detected. git remote not updated")
+			utils.Check(err)
+		} else {
+			ui.Errorln("Existing repository detected. Updating git remote")
+		}
 	} else {
 		if !args.Noop {
 			repo, err := gh.CreateRepository(project, flagCreateDescription, flagCreateHomepage, flagCreatePrivate)
--- a/commands/create.go
+++ b/commands/create.go
@@ -109,15 +109,14 @@ func create(command *Command, args *Args) {
 	project := github.NewProject(owner, newRepoName, host.Host)
 	gh := github.NewClient(project.Host)
 
-	if gh.IsRepositoryExist(project) {
+	repo, err := gh.Repository(project)
+	if err != nil {
+		utils.Check(err)
+	}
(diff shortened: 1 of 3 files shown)


commands/create.go
```diff
@@ -110,7 +110,18 @@ func create(command *Command, args *Args) {
 	gh := github.NewClient(project.Host)
 
 	if gh.IsRepositoryExist(project) {
-		ui.Errorln("Existing repository detected. Updating git remote")
+
+		repo, err := gh.Repository(project)
```
mislav: This queries `gh.Repository()` info again after `gh.IsRepositoryExist()` already queried it (but only returned a boolean). If you need to query whether a repository exist *and* access its info, it would be best to call `gh.Repository()` once initially and base all subsequent checks on that result.

commands/create.go
```diff
@@ -110,7 +110,18 @@ func create(command *Command, args *Args) {
...
+		if err != nil {
+			utils.Check(err)
+		}
+
+		if !repo.Private && flagCreatePrivate {
+			err = fmt.Errorf("Existing public repository detected. git remote not updated")
```
mislav: I would like to see a clearer error message, like:

```go
fmt.Errorf("Repository `%s' already exists and is public", repo.FullName)
```

nrmitchi: I wasn't quite sure about the messaging, I'll go with your suggestion

commands/create.go
nrmitchi: Good catch. Seeing that the `IsRepositoryExist` function is not used elsewhere, do you have objections to removing it?

commands/create.go
```diff
@@ -109,8 +109,18 @@ func create(command *Command, args *Args) {
 	project := github.NewProject(owner, newRepoName, host.Host)
 	gh := github.NewClient(project.Host)
 
-	if gh.IsRepositoryExist(project) {
-		ui.Errorln("Existing repository detected. Updating git remote")
+	repo, err := gh.Repository(project)
+	if err != nil {
+		utils.Check(err)
```
mislav: I think this breaks current CI checks. Keep in mind that `err` will be present if the repo doesn't exist. This whole conditional can go away.

commands/create.go
```diff
@@ -109,8 +109,18 @@ func create(command *Command, args *Args) {
...
+	}
+
+	if repo != nil {
```
mislav: This can be `err == nil`. In that case you can be sure that `repo` is non-nil. No other checks are necessary

commands/create.go
nrmitchi: Yupp, this took me a bit to figure out why all the tests were breaking. It did occur to me that we don't currently differentiate between a "the repo doesn't exist" error, and a _real_ error. In that situation a duplicate error wouldn't be created (assuming the github api handles the conflict properly), however it could be dealt with better in hub itself.

It's not really a concern here though, and is outside the scope of what I'm trying to accomplish with this PR

mislav:
Thank  you!! ✨

nrmitchi:
No problem! Thanks for getting this merged in so quickly

PR #1864, opened 2018-09-04
Add Go module

shuLhan:
The list of required directive is automatically detected by "go mod" from
Gopkg.lock file.


go.mod
```diff
@@ -0,0 +1,17 @@
+module github.com/github/hub
```
mislav: Do we need to add `/v2` here?

mislav:
Perhaps we should have a CI build step for Go version 1.11 (and above) that test with `GO111MODULE=on` so that both modules and legacy `vendor/` approach are exercised in tests

go.mod
shuLhan: Since this repository is a program, I don't think its needed.

shuLhan:
Hey @mislav , thank you reviewing.

> Is there anything else that we need to change?

As you already comment above, maybe add test to build without GOPATH in CI.

> How would people use this module support with hub?

Since this repository is already use semver, user that "go get" specific tag will use the go.mod in the commit.

go.mod
mislav: https://github.com/golang/go/wiki/Modules#semantic-import-versioning:

> If the module is version v2 or higher, the major version of the module _must_ be included in both the module path in the go.mod file (e.g., `module example.com/my/mod/v2`) and the package import path (e.g., `import "example.com/my/mod/v2/foo"`).

shuLhan: Done.

mislav:
This was superseded by https://github.com/github/hub/pull/2138 which doesn't break the CI build.

PR #1900, opened 2018-10-14
fix `compare` command if not on any branch

### LINKED ISSUE #1584: hub compare does not work when head is detached, even if a full range is specified
The following sequence of commands make `hub` fail:
```
$ git checkout --detach
HEAD is now at 978a908... <redacted>
$ hub compare release..master
Aborted: not currently on any branch.
```
This would be understandable if I would have omitted the upper boundary, in which case according to `--help` the current branch would be used, but I would expect it to work correctly when full ranges are specified.

Crunch09:
There is no need to detect the current branch if a range is provided as an argument.

Fixes #1584.

### DIFF
--- a/commands/compare.go
+++ b/commands/compare.go
@@ -72,14 +72,14 @@ func compare(command *Command, args *Args) {
 		r       string
 	)
 
-	branch, project, err = localRepo.RemoteBranchAndProject("", false)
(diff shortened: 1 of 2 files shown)


mislav:
Looks great; thank you! 💫

PR #1903, opened 2018-10-17
Have the -a flag request a review instead of assigning the pull request

dannygrant-zp:
We've been using the reviewer feature since it was released, and the -a flag was really helpful when you knew who you wanted a review from. This changes it to request a review instead of setting the assignee. I kept it as the `-a` option because it's familiar and because `-r` is already taken

API documentation on creating requesting reviews on a pull request:
https://developer.github.com/v3/pulls/review_requests/#create-a-review-request


dannygrant-zp:
whoops. this was meant to be merged into a fork

PR #1920, opened 2018-10-28
Adding unit tests for util methods

tariq1890:
Making the TimeAgo method testable


mislav:
Thank you! More tests is always great 👍

Issue #1913, opened 2018-10-23
ability to list reviewers with the --format flag


whatsadebugger:
%as lists assignees
It would be nice to see reviewers from other prs so you can copy them over to your own PRs

mislav:
Interesting use-case! Agreed.

PR #1924, opened 2018-10-29
Add ability to list requested reviewers with `%rs`

Crunch09:
Not sure if I understood @whatsadebugger correctly in #1913.
If instead we want to list all reviewers (not just requested ones) we
would need to make additional API requests I think.


mislav:
Thank you for the feature!

I've added a fix that includes requested team names in the list as well. ✨

Crunch09:
@mislav Great, thank you! And cheers for adding to the star wars theme in the test! 😅

whatsadebugger:
You got it right. Just requested reviewers. Thanks a lot 👍

PR #1926, opened 2018-10-31
Add licensing information to README

karlcoelho:


mislav:
Sorry, but we want the people to actually consult the local LICENSE file instead of following the online link.

If we change the license in the future, but somebody has a hub version downloaded before that change, when this person looks up the license for the project, we want them to read the local LICENSE file instead of following this link. If they followed it, they would get the license for the latest master branch of the project, which might not match the license of the currently downloaded version that they have.

PR #1931, opened 2018-11-02
Add a new command via you can close pr

darland:
Hi guys. What do you think about that?


mislav:
Hi thank you for your contribution and sorry for the late review!

After some consideration, I have decided that—for now—this is best implemented in user scripts, rather than maintained in hub core. Now that there is a new `hub api` command, I will close this feature request and leave implementing this as an exercise to the reader. https://github.com/github/hub/releases/tag/v2.8.3 https://github.com/github/hub/pull/2016

In the future there will be a command that updates existing pull requests. That command will also have the ability to close PRs.

PR #1916, opened 2018-10-24
 Use 'scissors' to separate comments from PR messages 

jonnystoten:
This fixes #1377.

Before this fix, Markdown headers in PULL_REQUEST_TEMPLATE.md would be ignored as # is the default comment character.

PR #1939, opened 2018-11-12
add pullrequest vim syntax.

dsifford:
See #1916. ([discussion starts here](https://github.com/github/hub/pull/1916#issuecomment-434759694))

This _should_ install fine on MacOS and Linux. Windows install steps still need to be added. I don't own a windows machine so any tips there would be appreciated.

One thing I'm noticing right now as I write this is that on MacOS the pullrequest template doesn't render with any whitespace between the title line and the scissors line, so the syntax currently causes the first line of the scissors line to be highlighted with an error group.

I think it would be best to keep that way and instead just put some whitespace in the template if it doesn't already exist.

Let me know if you need anything else. Thanks!

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -16,7 +16,7 @@ fi
 prefix="${PREFIX:-$prefix}"
 prefix="${prefix:-/usr/local}"
 
-for src in bin/hub share/man/*/*.1; do
+for src in bin/hub share/man/*/*.1 share/vim/vimfiles/*/*.vim; do
   dest="${DESTDIR}${prefix}/${src}"
   mkdir -p "${dest%/*}"
   [[ $src == share/* ]] && mode="644" || mode=755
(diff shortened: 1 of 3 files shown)


mislav:
Thank you; this looks great!

jasonkarns:
@dsifford awesome!

Will this require an update to the homebrew formula for hub to ensure the syntax gets installed? It doesn't _appear_ that the homebrew formula runs `install.sh` so I think this will need duplicated over there.

jasonkarns:
Or maybe not? Looks like it symlinked correctly?

```
$ ls -al /usr/local/share/vim/vimfiles
lrwxr-xr-x  1 jasonkarns  admin    48B Dec 21 09:23 /usr/local/share/vim/vimfiles@ -> ../../Cellar/hub/HEAD-cc38879/share/vim/vimfiles
```

mislav:
@jasonkarns The homebrew formula runs `make install`.

PR #1945, opened 2018-11-21
Add Support indentation YAML

inokappa:
Hi, 

My environment is as follows.

```sh
$ ruby --version
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin17]

$ bundle exec gem list

*** LOCAL GEMS ***

bundler (1.16.2)
byebug (10.0.2)
hub (1.12.4)
minitest (5.11.3)
```

When using the following YAML in $HOME/.config/hub, I got an error(`NoMethodError (undefined method []=' for nil:NilClass)`)  .

```yaml
$ cat $HOME/.config/hub
---
github.com:
  - user: test
    oauth_token: TEST
```

I modified to be able to parse either YAML indentation or not.
Please confirm.

Regards.

Yohei

### DIFF
--- a/lib/hub/github_api.rb
+++ b/lib/hub/github_api.rb
@@ -507,7 +507,8 @@ module Hub
             host = hash[$1] = []
           when /^([- ]) (.+?): (.+)/
             key, value = $2, $3
-            host << {} if $1 == '-'
+            host << {} if $1 == '-' or $2 =~ /^\s*-\s*/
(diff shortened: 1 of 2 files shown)


mislav:
Hi thank you for the fix! But, I'm not maintaining the 1.x line anymore. The Ruby implementation of hub is ancent and slow and you would do much better to upgrade to the latest 2.6 version.

Issue #1287, opened 2016-09-29
pull-request breaks with multiple labels


aesnyder:
```
hub pull-request -l foo, bar
Error updating issue: Unprocessable Entity (HTTP 422)
Missing field: "name"
```

```
hub pull-request -l foo
https://github.com/aesnyder/dotfiles/pull/6
```

Any suggestions?

aesnyder:
Make sure there aren't spaces between the labels:

`hub pull-request -l foo, bar`

PR #1952, opened 2018-11-28
Docs: mention that comma-separated lists must not have spaces

OliverJAsh:
This was tripping me up for so long. I eventually discovered the problem was my spaces, re. https://github.com/github/hub/issues/1287#issuecomment-250454584.

### DIFF
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -66,17 +66,17 @@ pull-request -i <ISSUE>
 		The head branch in "[OWNER:]BRANCH" format. Defaults to the current branch.
 
 	-r, --reviewer <USERS>
-		A comma-separated list of GitHub handles to request a review from.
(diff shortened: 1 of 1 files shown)


mislav:
Thanks for suggesting this!

1. How did spaces end up in your list in the first place? If you did `-a foo, bar` in your shell, that's passing 3 separate arguments: `-a`, `foo,` and `bar`. The final argument is going to get processed as having no relation to `-a` at all.

2. Could we just trim spaces around the comma in code instead? I see no negative consequences of doing that

mislav:
Thank you for explaining. I've asked because if you intend to use it as `-l "Status: Ready for Review", "Type: Bug"`, then trimming by itself wouldn't work— we would have to change the flag processing library to consider that the 3rd argument `"Type: Bug"` is in relation to `-l`.

PR #1958, opened 2018-12-01
Remove instructions for adding a git remote

KevinSjoberg:
The `hub create` command will automatically add a `origin` remote if not
present.


mislav:
Sorry, but this example shows what would be *run* when the hub command executes, not what the user should run. This distinction might not be clear immediately, but that was the intent. The rest of the documentation within this project follows that pattern.

```
$ hub command
> git command that would be run as result
> another git command that would be run
```

KevinSjoberg:
I see, makes sense when I think about it. 🙂

PR #1959, opened 2018-12-03
minor refactor of git methods

tariq1890:
A minor refactor in methods of `git.go`. This removes some code duplication and improves code re-use :)

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -215,10 +215,8 @@ func CommentChar(text string) (string, error) {
 }
 
 func Show(sha string) (string, error) {
-	cmd := cmd.New("git")
-	cmd.WithArg("-c").WithArg("log.showSignature=false")
-	cmd.WithArg("show").WithArg("-s").WithArg("--format=%s%n%+b").WithArg(sha)
-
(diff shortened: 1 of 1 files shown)


tariq1890:
@mislav requesting your review on this :)

mislav:
Hi thank you for your contribution! However, due to this PR having a subtle and possibly accidental functional change, and doesn't gain is much in terms of code readability, I'm going to make the call to close this.

The functional change is: `c := cmd.New("git"); c.WithArg("a")` is **not** the same as `gitCmd("a")` due to the fact that `gitCmd` also applies GlobalFlags, which the former approach doesn't do. This might not have any negative effect in most of cases, but could have hard-to-debug consequences in some rare cases. I would therefore vote to avoid making the functional change.

Also, repeating `c.WithArg("a").WithArg("b")` is not so bad, even though it looks repetitive at first. The code in question doesn't change often and is not necessarily difficult to read. [Not all code duplication is bad](https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction).

PR #1962, opened 2018-12-06
Improve Zsh completion installation docs

### LINKED ISSUE #1956: Suggestion: add docs for zsh completions
Out of the box, Zsh completions don't work as you would expect.

Given a fresh install:

``` bash
brew install zsh
brew install git
brew install hub
alias git=hub
```

When I type `git pull-<tab>`, it successfully completes to `git pull-request`.

However, completion does not work for subcommand arguments. For example, when I type `git pull-request -<tab>`, the arguments do not complete.

After some digging, I discovered that I had two sets of Zsh completions for Git (using the setup above, which I believe is the route most people will take):

- Git's official completions (e.g. `/usr/local/share/zsh/site-functions/_git`)
- Zsh's official Git completions (e.g. `/usr/local/Cellar/zsh/5.6.2_1/share/zsh/functions/_git`)

Both of these were on my `$fpath`.

I then discovered that the author of the Hub Zsh completions seems to advise the usage of Zsh's official Git completions, rather than Git's official completions: https://github.com/github/hub/pull/295#issuecomment-14324233

I removed Git's official completions with `rm /usr/local/share/zsh/site-functions/_git` and then Hub Zsh completions seemed to work as expected. That is, when I type `git pull-request -<tab>`, arguments *do* complete.

I would like to suggest that we clearly document this caveat—Hub Zsh completions only seem to fully work when using Zsh's official Git completions, rather than Git's official completions.

OliverJAsh:
Fixes #1956

### DIFF
--- a/etc/README.md
+++ b/etc/README.md
@@ -1,8 +1,31 @@
 # Installation instructions
 
-## bash + Homebrew
+## Homebrew
 
-If you're using Homebrew, just run `brew install hub` and you should be all set with auto-completion.
+If you're using Homebrew, just run `brew install hub` and you should be all set with auto-completion (for both Bash and Zsh).
+
+### Zsh
+
+hub's Zsh completions are [designed to work with Zsh's official Git completions](https://github.com/github/hub/pull/295#issuecomment-14324233), *not* Git's official Zsh completions.
+
+If you installed Zsh and Git via Homebrew, you will need to delete Git's official Zsh completions to allow Zsh's official Git completions to work, which in turn will allow hub's Zsh completions to work.
+
+You can do this deletion in `.zshrc`, which ensures the file remains deleted even after upgrades of Git.
+
+``` sh
+# Delete Git's official completions to allow Zsh's official Git completions to work.
+# This is also necessary for hub's Zsh completions to work:
+# https://github.com/github/hub/issues/1956.
+function () {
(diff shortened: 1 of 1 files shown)


apjanke:
+1. I like this change. It'd clear up some confusion for me – it wasn't clear to me that you need to delete the Git-provided `_git` completion to get things working best, until I found this PR. (I'd go so far as to add a "Yes, really: delete the Git-provided `_git` completion definition." to this doc.)

How about adding a link from the main README.m to this etc/README.md doc, too? Would make it easier to find.

OliverJAsh:
@apjanke Good idea. I'm happy to make that change. Is there anything else I need to change to get this approved/merged?

etc/README.md
```diff
@@ -1,8 +1,31 @@
 # Installation instructions
 
-## bash + Homebrew
+## Homebrew
 
-If you're using Homebrew, just run `brew install hub` and you should be all set with auto-completion.
+If you're using Homebrew, just run `brew install hub` and you should be all set with auto-completion (for both Bash and Zsh).
```
mislav: This is not true (for zsh). The reader might think that they are done at this point. However, if they also installed git with Homebrew, then their hub (when aliased as "git") completions are busted.

This sentence should add that you're all set **for bash**, but that you should read on for zsh instructions.

PR #1964, opened 2018-12-10
Restore pre-2.20.0 `git help` output

### LINKED ISSUE #1963: Incompatible with git 2.20.0
## My system config
```
$ git --version
git version 2.20.0
hub version 2.6.0
```

## The problem

After upgraded to git 2.20.0, running `hub <alias>` would freeze the command line prompt.

```
$ \git st # running git directly is fine
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean

$ hub st # running hub results to hang up
^C
```

## The cause

Running `ps aux | grep git` shows that the hub is stuck at running `git help -a`. The related codes are:

https://github.com/github/hub/blob/f4252d55ecfa76aa4db50458faab4a269dada014/git/git.go#L359-L374

Running `git help -a` manually, I found that since 2.20.0 git starts using page (e.g. `less`) to display help message.  As a result, it causes hub hang up when it invokes `git help -a` in the codes mentioned above.

jacobbednarz:
After upgrading to Git 2.20.0, all operations that relied on `hub alias`
would indefinitely hang. It can be reproduced using the following:

1. Create a new alias such as `st` for `git status`.
2. Run `hub <alias>` in a git repository.

  ```
  $ hub st

  ... hangs forever
  ```

Inspecting the running process shows that it is executing `git help -a`
and not able to proceed. Digging into the `hub` shows that the `help`
command is called within `IsBuiltInGitCommand`. Once this entrypoint was
found, I started bisecting and found this works in 2.19.x but fails in
2.20.x. Sure enough, the Git 2.20.0 release notes [mention a
corresponding change](https://github.com/git/git/blob/master/Documentation/RelNotes/2.20.0.txt#L17-L18).

>  * "git help -a" now gives verbose output (same as "git help -av").
>    Those who want the old output may say "git help --no-verbose -a".

It was even nice enough to include the backwards compatibility fix :)

To fix the hanging, I've restored the existing behaviour by including
the `--no-verbose` flag that mimics the 2.19.x behaviour.

Fixes #1963

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -357,7 +357,7 @@ func gitCmd(args ...string) *cmd.Cmd {
 }
 
 func IsBuiltInGitCommand(command string) bool {
-	helpCommandOutput, err := gitOutput("help", "-a")
+	helpCommandOutput, err := gitOutput("help", "--no-verbose", "-a")
 	if err != nil {
 		return false
 	}
--- a/git/git.go
+++ b/git/git.go
@@ -359,6 +359,10 @@ func gitCmd(args ...string) *cmd.Cmd {
 func IsBuiltInGitCommand(command string) bool {
 	helpCommandOutput, err := gitOutput("help", "--no-verbose", "-a")
 	if err != nil {
+		// support git versions that don't recognize --no-verbose
+		helpCommandOutput, err = gitOutput("help", "-a")
+	}
+	if err != nil {
 		return false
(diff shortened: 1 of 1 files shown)


jacobbednarz:
Thanks for the additional help on this one @mislav, you rock! :tada:

Issue #1961, opened 2018-12-05
Change CLI processing so that `--flag value` is equivalent to `--flag=value`


mislav:
This has been tripping up people (and ocassionally me) for a while. Ideally our flag processing would work identically to how git handles it.

https://github.com/ogier/pflag
https://github.com/github/hub/issues/1819#issuecomment-444577641

mislav:
@da-edra Please do! 🙏

Issue #1819, opened 2018-07-26
`hub pull-request --message <msg>` fails validation when valid


philwinder:
`hub pull-request --message <msg>` doesn't work as expected. When passing `--message "hi there"` it fails with `flag needs an argument`.

To recreate run the following dockerfile:

```Dockerfile
FROM alpine:3.8

# Add testing repo for hub
RUN echo "http://dl-cdn.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories

# Install utilities
RUN apk add --no-cache \
    git=2.18.0-r0 \
    bash=4.4.19-r1 \
    hub=2.4.0-r0

RUN hub version

# This work (but obviously errors because we don't have a repo)
RUN hub pull-request -m "hi there" || true

# This does not
RUN hub pull-request --message "hi there"
```

Which when building produces:

```shell
$ docker build -t temp .
Sending build context to Docker daemon  2.048kB
Step 1/6 : FROM alpine:3.8
 ---> 11cd0b38bc3c
Step 2/6 : RUN echo "http://dl-cdn.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories
 ---> Using cache
 ---> 3e02c5483900
Step 3/6 : RUN apk add --no-cache     git=2.18.0-r0     bash=4.4.19-r1     hub=2.4.0-r0
 ---> Using cache
 ---> 8a8c8ef68e7a
Step 4/6 : RUN hub version
 ---> Using cache
 ---> 8fe076fd3dfd
Step 5/6 : RUN hub pull-request -m "hi there" || true
 ---> Running in 6997405a8032
fatal: Not a git repository
Removing intermediate container 6997405a8032
 ---> 34ce91dbe0e0
Step 6/6 : RUN hub pull-request --message "hi there"
 ---> Running in 8b08dc0fe7d5
flag needs an argument: --message

Usage: hub pull-request [-focp] [-b <BASE>] [-h <HEAD>] [-r <REVIEWERS> ] [-a <ASSIGNEES>] [-M <MILESTONE>] [-l <LABELS>]
       hub pull-request -m <MESSAGE> [--edit]
       hub pull-request -F <FILE> [--edit]
       hub pull-request -i <ISSUE>
The command '/bin/sh -c hub pull-request --message "hi there"' returned a non-zero code: 1
```

mislav:
[...]

PR #1971, opened 2018-12-18
Clarify long-form CLI options

mislav:
Indicate that long-form CLI flags are used like `--message=VALUE`, and other tweaks.

Ref. https://github.com/github/hub/issues/1819 https://github.com/github/hub/issues/1961

/cc @philwinder @devisnik @adamkaplan


adamkaplan:
Awesome, thank you!

PR #1974, opened 2018-12-20
Capitalize GitHub in doco

apjanke:
There's a couple places in the man pages where "GitHub" is mis-capitalized as "Github". This PR corrects them.


mislav:
Good catch; thank you!

PR #1977, opened 2018-12-21
Automatically add the `upstream` remote when cloning a fork

mislav:
```sh
$ hub clone my-fork
> origin:   git://github.com/MYUSER/my-fork.git
> upstream: git://github.com/OWNER/original-project.git

$ hub clone --origin=upstream my-fork
> upstream: git://github.com/MYUSER/my-fork.git
> origin:   git://github.com/OWNER/original-project.git
```

Ref. https://github.com/github/hub/issues/348

/cc @apjanke

### DIFF
--- a/commands/clone.go
+++ b/commands/clone.go
@@ -5,7 +5,9 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/github/hub/cmd"
 	"github.com/github/hub/github"
+	"github.com/github/hub/ui"
 	"github.com/github/hub/utils"
 )
 
@@ -56,15 +58,32 @@ func transformCloneArgs(args *Args) {
 	isSSH := parseClonePrivateFlag(args)
 	hasValueRegexp := regexp.MustCompile("^(--(upload-pack|template|depth|origin|branch|reference|name)|-[ubo])$")
 	nameWithOwnerRegexp := regexp.MustCompile(NameWithOwnerRe)
+	var targetDir string
+	var originName string
+	quiet := false
 	for i := 0; i < args.ParamsSize(); i++ {
 		a := args.Params[i]
 
 		if strings.HasPrefix(a, "-") {
+			if a == "--origin" || a == "-o" {
+				if i+1 < args.ParamsSize() {
+					originName = args.Params[i+1]
+				}
+			} else if strings.HasPrefix(a, "--origin=") {
+				originName = strings.TrimPrefix(a, "--origin=")
+			} else if strings.HasPrefix(a, "-o") {
+				originName = strings.TrimPrefix(a, "-o")
+			} else if a == "--quiet" || a == "-q" {
+				quiet = true
+			}
 			if hasValueRegexp.MatchString(a) {
 				i++
 			}
 		} else {
-			if nameWithOwnerRegexp.MatchString(a) && !isCloneable(a) {
+			if targetDir != "" {
+				targetDir = a
+				break
+			} else if nameWithOwnerRegexp.MatchString(a) && !isCloneable(a) {
 				name, owner := parseCloneNameAndOwner(a)
 				var host *github.Host
 				if owner == "" {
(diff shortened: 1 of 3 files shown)


commands/clone.go
```diff
@@ -116,9 +135,32 @@ func transformCloneArgs(args *Args) {
...
+						addRemote.WithArgs("remote", "add", "-f", upstreamName, upstreamUrl)
+						if !quiet {
+							ui.Errorf("Adding remote '%s' for the '%s/%s' repo\n",
+								upstreamName, repo.Parent.Owner.Login, repo.Parent.Name)
+						}
+						output, err := addRemote.CombinedOutput()
+						if err != nil {
+							ui.Errorln(output)
+						}
+						return err
```
apjanke: Maybe it should do a `git fetch` for the added remote at this point? The normal case for doing a `git clone` fetches the full history for the cloned repo; since we're basically cloning two repos at once here, maybe we should do fetch so we have both repos' history.

apjanke:
I like it!

I'm a little confused by the `--origin=upstream` scenario. In this case, is `MYUSER/my-fork` the fork root, or are we just using different names for the remotes of GitHub repos with the same fork relationship? If it's the latter, maybe the remote names should be `origin` and `<MYUSER>` instead of `origin` and `upstream`; I think that would match hub's existing naming conventions better: for the remote naming style that doesn't use `upstream`, generally `origin` is the fork tree root, and all other repos are named using the GitHub user name, including for the current user.

commands/clone.go
mislav: Notice that we're doing `git remote add -f`. The extra flag automatically initiates a fetch.

mislav:
> I'm a little confused by the `--origin=upstream` scenario.

Thanks for weighing in. I wasn't sure that my approach was the best solution. Primarily, I was just trying to avoid the `fatal: remote upstream already exists` error. Perhaps when the user does `--origin=upstream` explicitly, maybe we should skip adding an extra remote for the parent repo altogether?

commands/clone.go
apjanke: Gotcha. Missed that.

Kampfkarren:
What's with the holdup on a merge?

mislav:
@Kampfkarren The PR is ready, but I wanted to hold off until I can ensure (by adding tests) that other commands (e.g. `hub pull-request`) will also handle the triangular workflow well.

friederbluemle:
Regarding `--origin=upstream` I'm also wondering if we can find a better terminology. Both `origin` _and_ `upstream` are _remote names_. The only difference is that `origin` is the Git default, and `upstream` is a convention that some people prefer to use. Other than that there is nothing special about either name. At first glance it's not clear what it means to set them "equal"..

I also agree 100% with @apjanke regarding remote naming:

> generally `origin` is the fork tree root, and all other repos are named using the GitHub user name, including for the current user.

Further, since many people have `git` aliased as `hub` it may be better to preserve the **default** behavior of the `clone` command to only create one `origin` remote. This of course can be changed by command line arguments or even set permanently as config options when using `hub`.

Have you considered something like this:

```
$ hub clone my-fork
> origin:   git://github.com/<myuser>/my-fork.git

$ hub clone --with-root-remote my-fork
> origin:   git://github.com/<owner>/original-project.git
> myuser:   git://github.com/<myuser>/my-fork.git

$ hub clone --with-root-remote=upstream my-fork
> origin:   git://github.com/<myuser>/my-fork.git
> upstream: git://github.com/<owner>/original-project.git
```

mislav:
> generally `origin` is the fork tree root, and all other repos are named using the GitHub user name, including for the current user.

This was the main convention for as long as hub existed, but I feel the community has outgrown that convention. GitHub's own help docs now suggest a triangular git workflow where the canonical remote is called "upstream" and your own fork called "origin".

I want to adopt this in hub because it would help with scripting: it's way easier for everyone if `hub fork` sets up an "origin" remote that points to your newly created fork by default rather than naming it by your username, since then you can continue to interact with that remote right away in the script. If the remote name is username-based, then you first need to read the newly created remote name before you can start using it, which can be awkward to do.

Of course, no convention will ever handle everyone's use-case. Hub doesn't and will not try to support different popular git workflows out there; it will only handle the one we (GitHub) generally recommend and believe in.

nt4f04uNd:
@mislav can you revive this PR, this would be very useful

PR #1978, opened 2018-12-22
Tab completion for repo names in `hub clone`

apjanke:
How about some tab completion for the shorthand repo names in the `hub clone` command?

This PR adds support for tab completion of repo names in `git clone <repo>`. It supports both un-qualified repo names for the current user, and `<user>/<repo>` qualified names.

Examples:

```
hub clone <TAB>
```

will list all the current user's own repos.

```
hub clone foo<TAB>
```

will complete this user's repos starting with "foo".

```
hub clone defunkt/<TAB>
```

will list all repos for user `defunkt`.

screen shot 2018-12-22 at 12 58 46 am

screen shot 2018-12-22 at 12 59 07 am

### Limitations:

Does not complete user names, because there's just too darn many of them. (Maybe if the GitHub API exposed a prefix-filtered user listing endpoint, we could do that too.)

It's pretty slow right now. If this PR is accepted, I'll add some caching to it.

### DIFF
--- a/commands/internal_list_repos.go
+++ b/commands/internal_list_repos.go
@@ -48,10 +48,6 @@ func internalListRepos(cmd *Command, args *Args) {
 	var ownerName string
 	if len(words) > 0 {
 		ownerName = words[0]
-	} else {
-		user, err := gh.CurrentUser()
-		utils.Check(err)
-		ownerName = user.Login
 	}
 
 	repos, err := gh.FetchRepositories(ownerName)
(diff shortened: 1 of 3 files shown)


mislav:
@apjanke There is a `hub api` command now for low-level API access! You could consider re-implementing this using GraphQL calls (they're going to be faster) instead of adding a new internal endpoint. https://github.com/github/hub/releases/tag/v2.8.3 https://github.com/github/hub/pull/2016

apjanke:
Yep, that's a good idea. Don't pull this; I'll rewrite.

PR #1979, opened 2018-12-22
256-color terminal support

### LINKED ISSUE #1560: Label colors don't render even close to correctly in Terminal.app
With the `hub isssue -l config` command, the output looks like this:
image

The same repo with `ghi list --label config` looks like this:
image

apjanke:
Fixes #1560.

Here's a PR to add 256-color support for terminals like Terminal.app, which don't support 24-bit ("true") color.

This is a work in progress: the logic is there, but the colors aren't coming out right. They're too blue.

I added a test command `hub internal-print-color-test` so you can see a swatch of the color cube it's working with. Suggestions welcome on what I'm doing wrong here.

screen shot 2018-12-22 at 4 55 19 am

screen shot 2018-12-22 at 4 46 41 am

### DIFF
--- a/commands/internal-print-color-test.go
+++ b/commands/internal-print-color-test.go
@@ -1,0 +1,18 @@
+package commands
+
+import "github.com/github/hub/utils"
+
+var cmdInternalPrintColorTest = &Command{
+	Run:   internalPrintColorTest,
+	Usage: "internal-print-color-test",
+	Long: "Print a terminal color test swatch",
+}
+
+func init() {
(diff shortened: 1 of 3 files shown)


apjanke:
Ah, figured out the color problem: I was using `^` instead of `math.Pow` for exponentiation, but that's bitwise AND. Oops.

Fixed and amended this PR. Looks better now:

screen shot 2018-12-22 at 5 05 51 am

mislav:
Awesome work!! Thank you

PR #1980, opened 2018-12-22
Improve label contrast ratios

apjanke:
Fixes #1960.
Depends on and includes #1979.

This PR improves the contrast ratios of displayed label text by using a more sophisticated algorithm for picking foreground colors to go with the background issue colors.

I'm not sure I got the color candidate calculation right, because I'm not sure I understood the use of the terms "lightness" and "brightness" [in this comment](https://github.com/github/hub/issues/1960#issuecomment-443743138)

PR #1981, opened 2018-12-22
Change Color to use uint8 RGB components

apjanke:
Is there a particular reason that `Color` is using `int64` values for its R/G/B components instead of `uint8`? Since we're dealing with 24-bit color, valid R/G/B values are alway going to be 0-255 and fit in a `uint8`.

You don't need to worry about saving memory here, but a `uint8` seems like a more natural fit of the data structure to the model. Plus it would eliminate some possible out-of-range errors, and remove the need for explicit `float64` casts in some cases.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -84,29 +84,29 @@ func IsOption(confirm, short, long string) bool {
 }
 
 type Color struct {
-	Red   int64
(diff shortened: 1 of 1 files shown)


mislav:
Now bundled in #1980

PR #1982, opened 2018-12-22
ui Printf functions: handle errors by dying

apjanke:
The `ui` `Printf` and related functions return an `error` to indicate their status. But none of the call sites that are calling them are catching that returned error. Go linters see this as an un-handled error and complain about it.

What do you think about locally handling errors on stdout/stderr by just panicking and dying, removing the returned `error` from their signatures? This will get rid of the go linter warnings, and seems a little cleaner to me.


mislav:
Makes sense, thank you!

PR #1983, opened 2018-12-22
Replace deprecated Dial with DialContext

apjanke:
Dial is deprecated as of Go 1.7, replaced with DialContext.

This PR updates it to avoid the deprecation warning.


mislav:
Thank you!

PR #1984, opened 2018-12-22
Rename local variables to avoid collisions with package names

apjanke:
golint is issuing some warnings about local variable names colliding with imported package names. Are you interested in getting rid of those warnings by renaming the local variables?

(Or maybe rename the `cmd` package to the longer `command`, and preserve `cmd` as a local variable name?)

### DIFF
--- a/commands/compare.go
+++ b/commands/compare.go
@@ -120,10 +120,10 @@ func compare(command *Command, args *Args) {
 	}
 
 	subpage := utils.ConcatPaths("compare", rangeQueryEscape(r))
-	url := project.WebURL("", "", subpage)
+	myUrl := project.WebURL("", "", subpage)
 
 	args.NoForward()
-	printBrowseOrCopy(args, url, !flagCompareURLOnly && !flagCompareCopy, flagCompareCopy)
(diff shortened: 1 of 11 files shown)


apjanke:
Yeah, golint (with offenses causing a CI error) in the CI sounds like a good idea.

Would you like me to look at adding it to the Travis CI configuration as part of this PR?

apjanke:
Well, shucks. From the [golint README](https://github.com/golang/lint/blob/master/README.md):

> Golint is not perfect, and has both false positives and false negatives. Do not treat its output as a gold standard. We will not be adding pragmas or other knobs to suppress specific warnings, so do not expect or require code to be completely "lint-free". In short, this tool is not, and will never be, trustworthy enough for its suggestions to be enforced automatically, for example as part of a build process.

So we probably won't be able to get to a fully-clean baseline. That probably rules out using it as a CI check.

apjanke:
I think doing golint as part of CI is no-go, due to the above. Are you still interested in merging this, or should I just close it out?

mislav:
Since we won't use golint in CI, and golint doesn't even show these warnings for me in development, there doesn't seem to be a path forward to resist future variables shadowing packages, so I'm going to close this out. Thank you for looking into this!

PR #1992, opened 2018-12-30
build: embed reproducible timestamp in manpages

eli-schwartz:
If the environment variable SOURCE_DATE_EPOCH is set in the build environment, prefer it over the current date when determining the correct timestamp to use as the date the manpages were built.

See https://reproducible-builds.org/specs/source-date-epoch/


mislav:
This is great; thank you!! TIL

PR #1994, opened 2018-12-31
build: respect environment LDFLAGS and strip the build path

eli-schwartz:
golang does not natively respect LDFLAGS, but you can pass them on the command line using -ldflags=-extldflags=...
This is important for distributions, in order to provide common functionality such as hardening flags.

Also strip the prefixed root source directory from the embedded source file paths. This is not important information for the debugger, which should only care about paths relative to $GOPATH, and results in less build environment metadata leaking into the final binary. (This also aids in reproducible builds when using different build directories, see e.g. https://github.com/golang/go/issues/16860)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,9 @@ SOURCES = $(shell script/build files)
 SOURCE_DATE_EPOCH ?= $(shell date +%s)
 BUILD_DATE = $(shell date -u -d "@$(SOURCE_DATE_EPOCH)" '+%d %b %Y' 2>/dev/null || date -u -r "$(SOURCE_DATE_EPOCH)" '+%d %b %Y')
 HUB_VERSION = $(shell hub version | tail -1)
+export LDFLAGS := -extldflags=$(LDFLAGS)
+export GCFLAGS := all=-trimpath=$(PWD)
+export ASMFLAGS := all=-trimpath=$(PWD)
 
 MIN_COVERAGE = 89.4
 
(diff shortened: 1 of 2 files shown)


eli-schwartz:
Before:

```
$ checksec -f bin/hub
RELRO           STACK CANARY      NX            PIE             RPATH      RUNPATH	Symbols		FORTIFY	Fortified	Fortifiable  FILE
Partial RELRO   Canary found      NX enabled    PIE enabled     No RPATH   No RUNPATH   No Symbols       Yes	0		2	bin/hub
$ strings bin/hub| grep $USER
/home/eschwartz/git/hub/main.go
/home/eschwartz/git/hub/.gopath/src/github.com/github/hub/commands/version.go
/home/eschwartz/git/hub/.gopath/src/github.com/github/hub/commands/utils.go
/home/eschwartz/git/hub/.gopath/src/github.com/github/hub/commands/sync.go
[...]
```

After, with `LDFLAGS="-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now"` (note the presence of relro)

```
$ checksec -f bin/hub
RELRO           STACK CANARY      NX            PIE             RPATH      RUNPATH	Symbols		FORTIFY	Fortified	Fortifiable  FILE
Full RELRO      Canary found      NX enabled    PIE enabled     No RPATH   No RUNPATH   No Symbols       Yes	0		2	bin/hub
$ strings bin/hub| grep $USER
$ 
```

eli-schwartz:
I've been trying to think what to do for go 1.9, the problem is that `-gcflags=all=` on go>=1.10 is equivalent to `-gcflags=` on go 1.9 and using the latter everywhere causes newer versions of go to not trim all paths (only some).

From the release notes:

>  The go build -asmflags, -gcflags, -gccgoflags, and -ldflags options now apply by default only to the packages listed directly on the command line. For example, go build -gcflags=-m mypkg passes the compiler the -m flag when building mypkg but not its dependencies. The new, more general form -asmflags=pattern=flags (and similarly for the others) applies the flags only to the packages matching the pattern. For example: go install -ldflags=cmd/gofmt=-X=main.version=1.2.3 cmd/... installs all the commands matching cmd/... but only applies the -X option to the linker flags for cmd/gofmt. For more details, see go help build.

I think the solution is to do string comparison on the output of `go version` to choose between the two.

PR #2017, opened 2019-01-27
Makefile: fix arguments to go, not matching what go expects

eli-schwartz:
This caused gcc to receive the actual argument "''" or for example "'-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now'" and then error with
gcc: error: '-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now': No such file or directory

Furthermore, the trimpath arguments tried to trim an actual path resembling "'$PWD'", which did nothing as the literal single quotes were not valid in this context.

The former issue means it is down to the go compiler to interpret -ldflags properly in order to break up the (multiple) flag values and correctly parse the extldflags, possible spaces included.

In the latter case, trimpath only accepts one argument and the shellscript already correctly double-quotes this in order to pass it as one argv[3] to the compiler:
{ "go", "build", "-gcflags", "all=-trimpath=/path/to/pwd" }

And from there the compiler should be able to interpret paths with spaces on its own.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,9 +3,9 @@ SOURCE_DATE_EPOCH ?= $(shell date +%s)
 BUILD_DATE = $(shell date -u -d "@$(SOURCE_DATE_EPOCH)" '+%d %b %Y' 2>/dev/null || date -u -r "$(SOURCE_DATE_EPOCH)" '+%d %b %Y')
 HUB_VERSION = $(shell bin/hub version | tail -1)
 FLAGS_ALL = $(shell go version | grep -q 'go1.[89]' || echo 'all=')
-export LDFLAGS := -extldflags='$(LDFLAGS)'
-export GCFLAGS := $(FLAGS_ALL)-trimpath='$(PWD)'
-export ASMFLAGS := $(FLAGS_ALL)-trimpath='$(PWD)'
+export LDFLAGS := -extldflags '$(LDFLAGS)'
+export GCFLAGS := $(FLAGS_ALL)-trimpath '$(PWD)'
+export ASMFLAGS := $(FLAGS_ALL)-trimpath '$(PWD)'
 
 MIN_COVERAGE = 89.4
 


eli-schwartz:
Hmm, according to strace, when invoked as `-ldflags "-extldflags='-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now'"` this gets passed on as-is to `go tool link "-extldflags='-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now'"` which is behavior I did not really expect, but when you remove the "=" from the extldflags, it gets passed as two arguments and the second argument no longer contains the single-quote literals.

Makefile
```diff
@@ -3,9 +3,9 @@ SOURCE_DATE_EPOCH ?= $(shell date +%s)
 BUILD_DATE = $(shell date -u -d "@$(SOURCE_DATE_EPOCH)" '+%d %b %Y' 2>/dev/null || date -u -r "$(SOURCE_DATE_EPOCH)" '+%d %b %Y')
 HUB_VERSION = $(shell bin/hub version | tail -1)
 FLAGS_ALL = $(shell go version | grep -q 'go1.[89]' || echo 'all=')
-export LDFLAGS := -extldflags='$(LDFLAGS)'
-export GCFLAGS := $(FLAGS_ALL)-trimpath='$(PWD)'
-export ASMFLAGS := $(FLAGS_ALL)-trimpath='$(PWD)'
+export LDFLAGS := -extldflags '$(LDFLAGS)'
```
mislav: Let's remove the quotes instead and put back the `=` and have people write to us if they experience problems with values that include spaces.

eli-schwartz: Assuming the rationale is because you're not sure either, how the arguments are meant to be passed in such cases?

Done.

mislav: Oh I thought my addition of the quotes was the problem? And that removing them will solve it. But you're saying that doesn't completely solve it? I'm not sure about the proper solution, since I don't use these compiler flags at all, so I'll defer to your judgement.

Issue #2028, opened 2019-02-05
hub api to take a raw JSON body


mlk:
I'd like to be able to interact with [branch protection API](https://developer.github.com/v3/repos/branches/#update-branch-protection). However using the current API it does not look like it supports sending arrays for example.

I'd like a way of just sending a pre-formed JSON body to the API.
i.e.
```

hub api --raw-body '{ "enforce_admins": false, "required_pull_request_reviews": { "dismissal_restrictions": null, "dismiss_stale_reviews": false, "require_code_owner_reviews": false, "required_approving_review_count": 1 } , "restrictions": null, "required_status_checks": { "status": false, "context": [] }  }' -X PUT repos/{owner}/{repo}/branches/master/protection
```

mislav:
Thanks for writing in. Yes, unfortunately we don't support passing array/object fields yet. Your idea to pass raw body in is sound. In addition, I'd like to also support passing array/objects with `--field`, but I don't have an idea yet of how the syntax would look like. 🤔

fsouza:
@mislav I _just_ opened this PR (coincidence, really lol): https://github.com/github/hub/pull/2029

But after reading this issue, I believe that supporting raw JSON fields (in addition to raw JSON body) may be a better solution?

mlk:
Awesome work. Thank you.

PR #2029, opened 2019-02-05
commands/api: add support for reading array fields

fsouza:
Not sure if the syntax is the best? I just hit this as part of some integration work with deployments and thought that maybe this could be an acceptable solution for taking arrays of strings (for ``required_contexts``).

Later I saw #2028, so maybe we need something that will work for more cases than just a slice of strings (or we could make it ``[]interface{}`` and call ``magicValue`` for each item in the array, don't know).

### DIFF
--- a/commands/api.go
+++ b/commands/api.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"encoding/csv"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -248,12 +249,16 @@ func readFile(file string) (content []byte) {
 
 func readArray(input string) []interface{} {
 	input = strings.TrimRight(strings.TrimLeft(input, "["), "]")
-	values := []interface{}{}
-	for _, v := range strings.Split(input, ",") {
-		v = strings.TrimSpace(v)
-		if v != "" {
-			values = append(values, magicValue(v))
-		}
+	r := csv.NewReader(strings.NewReader(input))
+	r.TrimLeadingSpace = true
+	r.LazyQuotes = true
+	parts, err := r.Read()
+	if err != nil && err != io.EOF {
+		utils.Check(fmt.Errorf("invalid array value %q: %v", input, err))
+	}
+	values := make([]interface{}, len(parts))
+	for i, v := range parts {
+		values[i] = magicValue(v)
 	}
 	return values
 }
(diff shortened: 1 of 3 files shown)


fsouza:
Yeah I thought about the first question too and I think it may be easier to support raw JSON fields? What do you think?

If not, we can make this work by interpreting the contents of ``[...]`` as a CSV line.

Question 2 is easy, we can convert this into ``[]interface{}`` and call ``magicValue``  for each item.

fsouza:
@mislav implemented support for multiple types and strings that contain commas. Let me know if that makes sense.

TBH, for my use case, being able to specify raw JSON would suffice :)

mislav:
> Question 2 is easy, we can convert this into `[]interface{}` and call `magicValue` for each item.

That works for getting numbers/booleans/null. However, now you have another problem: how do you pass an array of strings `[true,12,null]` and have them remain strings?

To achieve that for regular fields, we have `--raw-field`.

fsouza:
@mislav I though the expectation is that if you're using ``--fields`` you understand that the values that are passed are interpreted, no? x)

We'd probably need to come up with a solution for passing arrays to ``--raw-field`` (which is really ``--string-field``).

Another alternative, for arrays, is to make ``hub api`` behave like curl and allow users to repeat keys when they want to pass multiple values (``-F required_contexts=v1 -F required_context=v2``). The flags are already inspired by curl, you might as well make them behave the same.

The same could also apply to headers. Then you can use ``--field`` or ``--raw-field`` as you wish. This is way easier with ``application/x-www-form-urlencoded`` than with JSON, but can still be done.

mislav:
> Another alternative, for arrays, is to make `hub api` behave like curl and allow users to repeat keys when they want to pass multiple values (`-F required_contexts=v1 -F required_context=v2`)

That's not a bad idea. I was thinking about something like supporting the syntax `-F required_contexts[]=v1 -F required_context[]=v2` to build out arrays. Then a person could continue to use `-F` for magic values and `-f` for string values. However, that still leaves features to be desired:

1. How would a person pass in an empty array?
2. How would a person pass in an array of hashes?
3. How would someone specify a nested hash?

Because it doesn't seem that we can come up with an elegant approach that handles all of these use-cases, I think simply supporting raw input and having the user construct the payload outside of `hub` seems like the best approach for now.

Thank you for all your contribution and thoughts so far!

fsouza:
@mislav oh, that's a great point, tots missed it. Thank _you_ for the all feedback and patience, and for actually implementing the solution 😁 

Closing this in favor of #2031.

PR #2031, opened 2019-02-07
Allow raw request body via `hub api --input <FILE>`

### LINKED ISSUE #2028: hub api to take a raw JSON body

mislav:
This is to enable nested JSON arrays/hashes.

Closes #2029, fixes #2028
/cc @mlk @fsouza

### DIFF
--- a/commands/api.go
+++ b/commands/api.go
@@ -16,7 +16,7 @@ import (
 
 var cmdApi = &Command{
 	Run:   apiCommand,
-	Usage: "api [-t] [-X <METHOD>] [--cache <TTL>] <ENDPOINT> [-F <KEY-VALUE>]",
+	Usage: "api [-it] [-X <METHOD>] [-H <HEADER>] [--cache <TTL>] <ENDPOINT> [-F <FIELD>|--input <FILE>]",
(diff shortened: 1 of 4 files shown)


fsouza:
@mislav I actually take my approval back lol Why not build on the standard of ``@filename``? It works like that for graphql requests (I don't imagine anyone will ever write the gql query in the command line, but they technically could).

So one could do ``--input '{"field1": "value1", "arr": [1, 2, 3]}'`` or ``--input @body`` or ``--input @-``. What do you think?

Issue #2025, opened 2019-02-01
Ability to create remotes that aren't named origin


pgierz:
Hi guys,

The hub plugin makes my life *much* easier, thanks!!
A feature request: when I use hub create, the remote name is automatically set to origin. While this is certainly a fine default, would there be an easy way to add a command line option for something else? e.g. work, private

This was apparently already asked before, but I can't seem to tell how it was solved.
See also:
#158 #381

Cheers,
Paul

mislav:
There is `hub fork --remote-name work`, for example. Maybe the same flag could exist for `hub create`?

Until then, you can create a script that runs `hub create && git remote rename origin work`

pgierz:
@mislav do you know where this needs to be done? I'd be happy to give this a try but I don't have an overview of the codebase...

mislav:
@pgierz This is where "origin" is hardcoded within the `create` command: https://github.com/github/hub/blob/08004b921836a3fa8ae6054ab273cdbf6da39f8c/commands/create.go#L131

This is how the `fork` command reads the optional `--remote-name` flag: https://github.com/github/hub/blob/08004b921836a3fa8ae6054ab273cdbf6da39f8c/commands/fork.go#L67-L69

PR #2034, opened 2019-02-09
remote name can be specified in the create command

pgierz:
Hi @mislav, I've never written any go before, but I gave it a shot! Could you see if this works for [github#2025](https://github.com/github/hub/issues/2025)?

### DIFF
--- a/commands/create.go
+++ b/commands/create.go
@@ -128,7 +128,13 @@ func create(command *Command, args *Args) {
 	localRepo, err := github.LocalRepo()
 	utils.Check(err)
 
-	originName := "origin"
+        var originName string
+        if flagOriginRemoteName := args.Flag.Value("--remote-name"); flagOriginRemoteName != "" {
+                originName = flagOriginRemoteName
+        } else {
+                originName = "origin"
+        }
+
 	if originRemote, err := localRepo.RemoteByName(originName); err == nil {
 		originProject, err := originRemote.Project()
 		if err != nil || !originProject.SameAs(project) {
--- a/features/create.feature
+++ b/features/create.feature
@@ -27,6 +27,17 @@ Feature: hub create
       """
     When I successfully run `hub create -p`
     Then the url for "origin" should be "git@github.com:mislav/dotfiles.git"
+  Scenario: Create repo with new remote name specified
+    Given the GitHub API server:
+    """
+    post('/user/repos') {
+      assert :private => false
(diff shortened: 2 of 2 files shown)


features/create.feature
```diff
@@ -27,6 +27,17 @@ Feature: hub create
       """
     When I successfully run `hub create -p`
     Then the url for "origin" should be "git@github.com:mislav/dotfiles.git"
+  Scenario: Create repo with new remote name specified
+    Given the GitHub API server:
+    """
+    post('/user/repos') {
+      assert :private => false
```
mislav: This assertion is not necessary in this test

features/create.feature
```diff
@@ -27,6 +27,17 @@ Feature: hub create
       """
     When I successfully run `hub create -p`
     Then the url for "origin" should be "git@github.com:mislav/dotfiles.git"
+  Scenario: Create repo with new remote name specified
+    Given the GitHub API server:
+    """
```
mislav: Minor: the `"""` lines are usually indented one level under the `Given` clause (see other tests)

features/create.feature
```diff
@@ -27,6 +27,17 @@ Feature: hub create
...
+    post('/user/repos') {
+      assert :private => false
+      status 201
+      json :full_name => 'mislav/dotfiles'
+    }
+    """
+    When I sucessfully run `hub create --remote-name=work`
+    Then the url for "work" should be "git@github.com:mislav/dotfiles.git"
```
mislav: This looks good, perhaps there could be an additional assertion that says `And there should be no "origin" remote`

commands/create.go
```diff
@@ -128,7 +128,13 @@ func create(command *Command, args *Args) {
 	localRepo, err := github.LocalRepo()
 	utils.Check(err)
 
-	originName := "origin"
+        var originName string
+        if flagOriginRemoteName := args.Flag.Value("--remote-name"); flagOriginRemoteName != "" {
```
mislav: This looks good so far. You also have to add the `--remote-name` flag to the flags documentation near the start this file for the flag to start working. Something like:

```
	--remote-name <REMOTE>
		Set the name for the new git remote (default: "origin").
```

The Go code that you wrote here looks great, but you can also write it simpler:

```go
originName := args.Flag.Value("--remote-name")
if originName == "" {
  originName = "origin"
}
```

PR #2039, opened 2019-02-15
Add FreeBSD installation instructions

0mp:


README.md
```diff
@@ -68,6 +68,15 @@ On Arch Linux you can install `hub` from official repository:
 $ sudo pacman -S hub
 ```
 
+#### FreeBSD
+
+On FreeBSD you can install a prebuilt `hub` package with
+[pkg(8)](http://man.freebsd.org/pkg/8):
+
+```console
+# pkg install hub
```
mislav: Why does the command start with `#`?

0mp: Well, sudo is not a part of the FreeBSD base system and pkg has to be run as root. I've used `#` instead of `$` to show that root access is required. :smiley: 

Cheers!

mislav:
Thank you!

PR #2056, opened 2019-02-22
about.html

RepoPOD202:
This is a good idea and a replacement for the old README.md


mislav:
No sorry, I don't see how this extra file would be more useful than the existing documentation.

PR #2062, opened 2019-03-02
check head branch exists before creating PR

### LINKED ISSUE #1930: Warn about an unpushed branch when opening a pull-request
I think this was previously the behaviour (before the change from `#` to `>8` for the `PULLREQ_EDITMSG`).

Basically I'd like to see the commit messages in the text by default, so I can easily move them above the line if I want. Having the default Pull Request template text there as well also makes sense, maybe template above the scissors by default and commit messages below would make sense.

Basically something like this:

```diff
+<Github PR template>

# ------------------------ >8 ------------------------
# Do not modify or remove the line above.
# Everything below it will be ignored.

-<Github PR template>

+<commit messages>
```

Swapping "commit messages" and "Github PR template" would be fine too I guess.

Crunch09:
Previously hub checked that there were [no unpushed commits](https://github.com/github/hub/blob/b3b3ebb8e517acd269489b9bd163de2f8fd309fc/commands/pull_request.go#L219-L223) but this check doesn't work when there is not even a remote branch.

**Before**
screenshot 2019-03-02 at 22 58 28

**After**
screenshot 2019-03-02 at 23 00 22

This fixes #1930.

### DIFF
--- a/commands/pr.go
+++ b/commands/pr.go
@@ -344,18 +344,29 @@ func findCurrentPullRequest(localRepo *github.GitHubRepo, baseProject *github.Pr
 	}
 }
 
-func findPushTarget(branch *github.Branch) (*github.Branch, *github.Project, error) {
-	branchRemote, _ := git.Config(fmt.Sprintf("branch.%s.remote", branch.ShortName()))
-	if branchRemote == "" || branchRemote == "." {
-		return nil, nil, fmt.Errorf("branch has no upstream configuration")
+func branchTrackingInformation(branch *github.Branch) (string, *github.Branch, error) {
+	branchRemote, err := git.Config(fmt.Sprintf("branch.%s.remote", branch.ShortName()))
+	if branchRemote == "." {
+		err = fmt.Errorf("branch is tracking another local branch")
+	}
+	if err != nil {
+		return "", nil, err
 	}
 	branchMerge, err := git.Config(fmt.Sprintf("branch.%s.merge", branch.ShortName()))
 	if err != nil {
-		return nil, nil, err
+		return "", nil, err
 	}
-	headBranch := &github.Branch{
+	trackingBranch := &github.Branch{
 		Repo: branch.Repo,
 		Name: branchMerge,
+	}
+	return branchRemote, trackingBranch, nil
+}
+
+func findPushTarget(branch *github.Branch) (*github.Branch, *github.Project, error) {
+	branchRemote, headBranch, err := branchTrackingInformation(branch)
+	if err != nil {
+		return nil, nil, err
 	}
 
 	if headRemote, err := branch.Repo.RemoteByName(branchRemote); err == nil {
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -176,10 +176,21 @@ func pullRequest(cmd *Command, args *Args) {
 	}
 
 	force := args.Flag.Bool("--force")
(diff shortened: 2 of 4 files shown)


github/client.go
```diff
@@ -215,6 +215,17 @@ func (client *Client) Repository(project *Project) (repo *Repository, err error)
 	return
 }
 
+func (repo *Repository) HasBranch(client *Client, branch string) bool {
+	api, err := client.simpleApi()
+	if err != nil {
+		return false
+	}
+
+	res, err := api.Get(fmt.Sprintf("/repos/%s/%s/branches/%s", repo.Owner.Login, repo.Name, branch))
```
mislav: Instead of fetching branch info, and thus generating 1 extra API request, how about we simply improve the `invalid value for "head"` error message and replace it with the error message you propose in this PR? That way, the user gets the desired outcome (a more helpful error message), but we don't have to actually change any of the mechanism of opening a PR.

Crunch09: I'm happy to change the PR to do that 👍 My thinking was just that this way the user gets an error message before typing a description of the PR (in an enditor not just passing the `-m` argument) while if we change just the error message the user writes a description for a PR that can't be created.

mislav: Ah I see; thanks for explaining.

Another alternative way of implementing this that avoids an API request would be to check for whether there is a remote tracking branch for the current branch. If there is none, we can assume that the branch wasn't pushed yet. This check should only ever kick in if `-h` *wasn't* passed. Does this sound ok?

Crunch09: Sorry for my late reply and thanks for your review! Yes, this sounds good to me. I will try to update the PR.

Crunch09: After looking into the code it seems like at the moment we default to the current branch name if there is no head and we don't have a tracking branch:
https://github.com/github/hub/blob/94a6cf45588fb7ae3b695731a518924d75e8348c/commands/pull_request.go#L172-L174

Just wanted to check whether we should still go ahead with this alternative then as it might break people's usage?

mislav: There are two ways we can check that the current branch was pushed somewhere:
1. the current branch has tracking information: `git config --show-origin --get-regexp 'branch\.BRANCH\.*`; or
2. there is a `.git/refs/remotes/REMOTE/BRANCH` file to indicate that there exists a remote tracking branch.

Of course, it's still possible that the branch was somehow pushed and both checks (1) and (2) still fail, but I think that would be a rare case and the user should be able to work around it by using `-f / --force`.

Does this make sense?

Crunch09: Yeah, this makes it clear @mislav , thanks! 👍 I will add those checks instead of the additional API call and assume the branch has not been pushed if both of these fail.

cmcaine:
This bit me recently. Thank you for making a PR about it!

If you find you don't have the time for this, please ping me and I'll see if I can finish it up for you :)

Crunch09:
@cmcaine 👍  Thanks for the reminder about this :) I'll try to work on the update tomorrow but if I don't manage to get it done I'll ping you.

Crunch09:
@mislav Sorry for the delay. I now updated the PR to remove the new API call and instead check if the current branch is tracking a branch (with `git rev-parse --verify @{u}`) or there is a `refs/remotes` file.
From what I read the current version should be sufficient but please let me know if I should use `git config --show-origin --get-regexp 'branch\.BRANCH\.*` instead.

PR #2098, opened 2019-03-29
Open draft pull requests with `-d` / `--draft`

### LINKED ISSUE #2038: [Feature] Create a "Draft" pull request
What better time to request a feature in `hub` than moments after a feature is released in the GitHub UI! 😁 I love not having to go to the UI for common tasks, and I can see this new feature being included in my workflow for discussing a PR prior to wanting to merge it.

Reference: https://github.blog/2019-02-14-introducing-draft-pull-requests/

A cursory look at the API docs doesn't seem to have this available yet (I do see a "locked" flag), so this issue may hang out for a while.

mislav:
Thank you @thehunt33r for the initial implementation!

Fixes #2038

### DIFF
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -16,7 +16,7 @@ import (
 var cmdPullRequest = &Command{
 	Run: pullRequest,
 	Usage: `
-pull-request [-focp] [-b <BASE>] [-h <HEAD>] [-r <REVIEWERS> ] [-a <ASSIGNEES>] [-M <MILESTONE>] [-l <LABELS>] [--draft]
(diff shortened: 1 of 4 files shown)


jasonkarns:
Woot!

How easy would it be to add some mention of the PR being a draft to the footer help at the bottom of the PR message?

I _would_ have just suggested changing `Requesting a pull` to `Drafting a pull` but it seems the vim syntax is matching on `"^Requesting a pull"`. Perhaps "Requesting a pull (draft)"  or something?

mislav:
> How easy would it be to add some mention of the PR being a draft to the footer help at the bottom of the PR message?

I don't think that is necessary right now. People have explicitly chosen `-d/--draft` every time, so they will know that the PR they're opening is a draft. The message in the editor serves mostly to tell you which head/base branches will be used, which are determined by magic by default.

PR #2113, opened 2019-04-15
Support asset arguments after tag argument in release create command

hnakamur:
This pull request add support for optional asset `<FILE>` arguments after the `<TAG>` argument in `hub release create` command.

I think it is easier to use asset arguments than asset options because you can specify glob patterns like `*.tar.xz` if the shell support expanding glob patterns.

Also add a note that `-a` option can be specified multiple times in the long help of `hub release create command`.


hnakamur:
I'm closing this as I can now use [gh release create | GitHub CLI](https://cli.github.com/manual/gh_release_create) instead.

PR #2120, opened 2019-04-23
Added Chocolatey as Windows install method in the readme

richardballard:
Updated the readme to also include Chocolatey as an install method for Windows, instead of only listing Scoop.


mislav:
Thank you! Ref. e3bf3bc9712b12e9222faafc94df3fd285c26bf3

Issue #2134, opened 2019-05-07
hub breaks git when GIT_TRACE is used


eli-schwartz:
```
[eschwartz@arch ~/git/projectname]$ /usr/bin/git config --get alias.new
log --pretty=fuller -p @{1}..@{0}
[eschwartz@arch ~/git/projectname]$ GIT_TRACE=1 HUB_VERBOSE=1 hub new >/dev/null
$ git config alias.new
$ git help --no-verbose -a
$ git 18:48:21.913636 git.c:419 trace: built-in: git config alias.new
18:48:21.924560 git.c:670               trace: exec: git-18:48:21.913636 git.c:419 trace: built-in: git config alias.new
18:48:21.924676 run-command.c:643       trace: run_command: git-18:48:21.913636 git.c:419 trace: built-in: git config alias.new
git: '18:48:21.913636' is not a git command. See 'git --help'.
```

The heuristic parsing which hub is doing, has grabbed git's internal trace messages and inserted it into the attempted command, causing everything to become invalid.

eli-schwartz:
Note that this only seems to affect alias expansion, not builtin commands.

PR #2135, opened 2019-05-08
Enable GIT_TRACE

mislav:
Only process stdout when shelling out to git instead of combined stdout + stderr. In most cases, we want to silence stderr because we attach a custom error message to failure scenarios. In some other cases, have stderr be attached to stderr of parent process.

Fixes https://github.com/github/hub/issues/2134


eli-schwartz:
Thanks!

PR #2138, opened 2019-05-15
migrate to go modules from dep

tariq1890:
`dep` is on its way to get phased out by `go mod`.


go.mod
```diff
@@ -0,0 +1,20 @@
+module github.com/github/hub
```
mislav: Should we append `/v2` here? See discussion at https://github.com/github/hub/pull/1864#discussion_r214792768

tariq1890:
I don't think so. I don't see how this will be a breaking change for it to necessitate a change in the import path. Do we have many projects depending on hub as a library?

go.mod
tariq1890: I don't think so. I don't see how this will be a breaking change for it to necessitate a change in the import path. Do we have many projects depending on hub as a library?

PR #2139, opened 2019-05-17
set new remote as default for pushing when forked.

ahmubashshir:


mislav:
Thank you for your contribution, but I believe this feature would be disruptive to our users. The `remote.pushDefault` config overrides all other push configuration, which people might not realize.

I prefer to choose which remote to push to whenever I publish a branch. Example:

```console
git push -u <remote> HEAD
```

The `-u` option makes sure that my push configuration is saved so I don't have to provide any arguments to `git pull` or `git push` for that branch in the future.

PR #2141, opened 2019-05-18
[pr] add ability to open Pull Request in the browser

ashemesh:
add the ability to open a PR page in browser (also copy & URL print to screen).
ability to open the forked PR by pass the `head` flag.
Screenshot 2019-05-18 at 22 41 31
in the example above `hub` is a fork of `github/hub`.

### DIFF
--- a/commands/pr.go
+++ b/commands/pr.go
@@ -5,6 +5,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/github/hub/git"
 	"github.com/github/hub/github"
 	"github.com/github/hub/ui"
 	"github.com/github/hub/utils"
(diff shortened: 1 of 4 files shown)


ashemesh:
@mislav in the above scenario, if `origin` has PR with `feature` has the head, we query should `origin` and as a "fallback" check `upsteam` for a PR with `owner:feature` as head, do you agree?

ashemesh:
@mislav is this cover the scenario you wrote above?
Screenshot 2019-05-25 at 22 04 57
I'm not so happy with the amount of `if \ else` in the function code, but I can't think how to support all the possible combinations, so maybe it is better not to support the `head` flag, what do you think?

mislav:
@ashemesh I've cleaned up code & tests, added more tests, and added the `hub pr show <NUMBER>` functionality as per @whatsadebugger. Please review!

ashemesh:
@mislav sorry only saw it now! thank you!

mislav:
@ashemesh No worries; thank you for your contribution! 🎉

PR #2142, opened 2019-05-21
build: pass mod vendoring flag on newer versions of go

eli-schwartz:
When go autodetects that it is being run as a go mod, and that there is a vendor directory, it will still try to redownload all sources over the network, unless you use -mod=vendor to tell it to use that. Additionally, when using -mod=vendor the compiler will nicely avoid messing with $GOPATH at all, since it can operate in a completely self-contained manner.

Take advantage of this, when the detected go version is at least 1.11 (when the -mod flag was introduced).

### DIFF
--- a/script/build
+++ b/script/build
@@ -14,6 +14,7 @@ find_source_files() {
 build_hub() {
   mkdir -p "$(dirname "$1")"
   go build \
+	  $MOD_VENDOR_ARG \
 	  -ldflags "-X github.com/github/hub/version.Version=`./script/version` $LDFLAGS" \
 	  -gcflags "$GCFLAGS" \
 	  -asmflags "$ASMFLAGS" \
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,10 @@ export LDFLAGS := -extldflags '$(LDFLAGS)'
 export GCFLAGS := $(FLAGS_ALL)-trimpath '$(PWD)'
 export ASMFLAGS := $(FLAGS_ALL)-trimpath '$(PWD)'
 
+ifneq ($(MOD_VENDOR_ARG),)
+	unexport GOPATH
+endif
+
 MIN_COVERAGE = 89.4
 
 HELP_CMD = \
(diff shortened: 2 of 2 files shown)


eli-schwartz:
BTW the use of modules is super nifty and with its use, plus this, I can now remove a bunch of golang hacks from my distro packaging. Thanks!

eli-schwartz:
Ok -- this is entirely magical, everything errors out with "build flag -mod=vendor only valid when using modules" -- but it is using modules??? And it works on my machine and not in travis.

script/build
```diff
@@ -14,6 +14,7 @@ find_source_files() {
 build_hub() {
   mkdir -p "$(dirname "$1")"
   go build \
+      "$MOD_VENDOR_ARG" \
```
mislav: This should not have quotes. If MOD_VENDOR_ARG is empty, you do not want an empty string argument to be passed. You want no argument to be passed.

eli-schwartz: It also used space indents instead of tabs, that is fixed too.

eli-schwartz:
OK -- that mostly just told me what I already knew...

But local testing helped me figure out something else, which is that `make` will also fail locally for me... *iff* I set $GOPATH, *and* the hub repository is cloned to the $GOPATH. Simply building the repository with an externally defined $GOPATH is not sufficient to trigger this issue.

It seems that Travis' language support will clone the repository to /home/travis/gopath/src/github.com/github/hub, at which point this fails. Forcing -mod=vendor for a build in which you are `cd`ed to $GOPATH is apparently not something the go tools support.

Thoughts on fixing this?

mislav:
@eli-schwartz Thanks for digging into! Maybe we can clear GOPATH before running the build script in Travis? That way, even if the repo was initially cloned to GOPATH, with the environment variable being unset, it might cease to interfere with the module-based approach.

You are welcome to tweak `.travis.yml` in this PR and see if it helps. Thank you!

eli-schwartz:
I made that change in the Makefile itself. CI is now green.

mislav:
Thank you! 🙇

PR #2153, opened 2019-05-31
Update x/sys module to support Risc-V

carlosedp:
Updating projects using libraries that were recently updated to support Risc-V.

Go upstream work is tracked on: golang/go#27532

Risc-V software support tracker on https://github.com/carlosedp/riscv-bringup


mislav:
Thank you. I'm not familiar with Risc-V. What is the benefit for the hub end-user?

carlosedp:
Hi @mislav, Risc-V is an open-source architecture that is gaining traction and having new software for it boosts it's popularity. There is already one board with a Risc-V SOC, the SiFive Unleashed and more will come.

I use Hub daily in place of git and developing on and for Risc-V with it improves my productivity.

mislav:
Thank you for the explanation!

If this helps you compile hub on this new architecture, then I'm all for it. I don't see how this could have adverse affect on other platforms.

PR #2156, opened 2019-06-02
Move commitLogs declaration inside multiple commits block

danielcompton:
This makes it clearer that commits are only added as a commented section in the case when there are multiple commits on a branch.

I'm a very new go developer, so apologies if this is a non-idiomatic change.

### DIFF
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -86,7 +86,7 @@ pull-request -i <ISSUE>
 	-l, --labels <LABELS>
 		Add a comma-separated list of labels to this pull request. Labels will be
 		created if they do not already exist.
-	
+
 	-d, --draft
 		Create the pull request as a draft.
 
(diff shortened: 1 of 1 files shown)


commands/pull_request.go
```diff
@@ -267,12 +266,13 @@ of text is the title and the rest is the description.`, fullBase, fullHead))
 			re := regexp.MustCompile(`\nSigned-off-by:\s.*$`)
 			message = re.ReplaceAllString(message, "")
 		} else if len(commits) > 1 {
+			commitLogs := ""
```
mislav: Instead of adding this line, can you initialize the `commitLogs` variable by turning the `=` on the next line into `:=` instead?

danielcompton: Thanks, have updated this.

danielcompton:
This is ready for review again. Thanks!

PR #2158, opened 2019-06-04
Ignore VSCode

sgharms:
I'm not ready to give up vim, but sometimes I like trying other editors and git caught this in my dirty state. Seems wise to guard against?


mislav:
Sure; thank you!

PR #2161, opened 2019-06-06
Add STDIN support as -r / --REMOTE_URL argument for batch processing

sgharms:
## Use Case

My team uses `hub` to do issue reporting. The collection of repositories stays pretty static. By leveraging #2159 as the "innards" of a loop of a list of repo URLs delivered via STDIN, we generate a nice report.

## Example Use

`cat list_of_prework_urls.txt |bin/hub issue -r STDIN --include-pulls -f '%U%n' |tee  issues`

## Presentation Note

This PR is built off of #2159, so this particular diff is a bit hairy pending that landing to `master`. The diffset can be more easily grokked [here](https://github.com/sgharms/hub/compare/add-repository-flag-to-issues..add-stdin-support).

### DIFF
--- a/commands/issue.go
+++ b/commands/issue.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"bufio"
 	"fmt"
 	"net/url"
 	"os"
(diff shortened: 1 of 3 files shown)


mislav:
Thank you for your contribution, but I don't see this feature as something that we'd like to have or maintain in hub.

I fully agree with your other PR to add `-r` to `hub issue`. Ideally, I want to see `-r` supported in every hub command. When I find time, I will review and help with your other PR to get merged. Once that is done, you will be able to achieve this with a loop:

```sh
while read remote_url; do
  bin/hub issue -r "$remote_url" --include-pulls -f '%U%n'
done < list_of_prework_urls.txt | tee issues
```

Issue #2155, opened 2019-06-02
Adding a milestone to issue


daniel-vera-g:
From the man page I get:

`  hub issue create [-oc] [-m MESSAGE|-F FILE] [--edit] [-a USERS] [-M MILESTONE] [-l LABELS]`

When I try to open an issue like:

`hub issue create -m "Test" --edit -M v0.0.1 -l feature`

Or:

`hub issue create -m "Test" --edit -M "v0.0.1" -l feature`

with the milestone existing, it creates the issue with the label but without the Milestone.

I am using: 

```
git version 2.21.0 
hub version 2.11.2 
```

Did I miss something?

mislav:
It seems like `hub issue create -M` expects to receive a numeric ID for a milestone rather than its title suh as `v0.0.1` https://github.com/github/hub/blob/d76e3e8ebf5a37eb13dc820dc42a3262f9be6db3/commands/issue.go#L577-L579

`hub pull-request` is better because it also accepts milestone titles https://github.com/github/hub/blob/d76e3e8ebf5a37eb13dc820dc42a3262f9be6db3/commands/pull_request.go#L305-L315

I wasn't aware that these commands treat milestones differently. Would you like to submit a PR that extends the same approach that `hub pull-request` has to `hub issue create`?

daniel-vera-g:
Ah ok.

I am not really fluent in go, but will try to submit the PR in the course of this week.

mislav:
Fixed in https://github.com/github/hub/pull/2166

PR #2166, opened 2019-06-12
Add a milestone name instead of an id, when creating 

daniel-vera-g:
I adapted the suggestion in https://github.com/github/hub/issues/2155#issuecomment-498217736 to pass a name to the milestone instead of an id.

### DIFF
--- a/commands/issue.go
+++ b/commands/issue.go
@@ -169,7 +169,7 @@ hub-pr(1), hub(1)
 		-a, --assignee USER
 		-s, --state STATE
 		-f, --format FMT
-		-M, --milestone M
+		-M, --milestone NAME
 		-c, --creator USER
 		-@, --mentioned USER
 		-l, --labels LIST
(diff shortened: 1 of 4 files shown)


mislav:
@daniel-vera-g Thank you for your contribution! Your code was a good start, but the implementation was lacking: it only handled milestones specified by name, but not ones specified by number (keeping backwards-compatibility). This was evident by a test in CI failing.

I have added tests and code to fix this. Please review 🙇

daniel-vera-g:
> @daniel-vera-g Thank you for your contribution! Your code was a good start, but the implementation was lacking: it only handled milestones specified by name, but not ones specified by number (keeping backwards-compatibility). This was evident by a test in CI failing.
> 
> I have added tests and code to fix this. Please review

Oh, I am sorry. I probably overlooked the broken CI.
Nevertheless, thanks for the help. It was my first time working with go so it was a bit of unfamiliar.
Will keep the recommendations in mind for the next time ;)

PR #2198, opened 2019-07-19
Allow hub used in Actions by specifying GITHUB_USER

### LINKED ISSUE #2149: Using hub from GitHub Actions
When `hub api` is called from a GitHub Action with a `GITHUB_TOKEN` available without any additional configuration it fails with "Bad credentials".

Command:
```
HUB_VERBOSE=true hub api /repos/rtimush/test/git/refs
```

Output:
```
$ git rev-parse -q --git-dir
$ git remote -v
$ git config --get-all hub.host
> GET https://api.github.com/user
> Authorization: token [REDACTED]
> Accept: application/vnd.github.v3+json;charset=utf-8
< HTTP 403
{"message":"Resource not accessible by integration","documentation_url":"https://developer.github.com/v3/users/#get-the-authenticated-user"}
> GET https://api.github.com/repos/rtimush/test/git/refs
> Authorization: token 
> Accept: application/vnd.github.v3+json;charset=utf-8
< HTTP 401
{"message":"Bad credentials","documentation_url":"https://developer.github.com/v3"}
{"message":"Bad credentials","documentation_url":"https://developer.github.com/v3"}
```

mislav:
When supplied with GITHUB_TOKEN, hub will attempt to request the `user` API resource which isn't available in the context of Actions.

To work around this, scripts may now set GITHUB_USER to avoid hitting the unavailable API resource.

The actual value of GITHUB_USER is not used in most commands; therefore, it can be a name of an organization or even a dummy value.

Fixes #2149

### DIFF
--- a/features/authentication.feature
+++ b/features/authentication.feature
@@ -261,8 +261,33 @@ Feature: OAuth authentication
     And the stderr should contain exactly:
       """
       Error getting current user: Forbidden (HTTP 403)
-      Resource not accessible by integration\n
+      Resource not accessible by integration
+      You must specify GITHUB_USER via environment variable.\n
       """
+
+  Scenario: Credentials from GITHUB_TOKEN and GITHUB_USER
+    Given I am in "git://github.com/monalisa/playground.git" git repo
+    Given the GitHub API server:
+      """
+      get('/user') {
+        status 403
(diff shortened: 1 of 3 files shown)


aliashori75:
cmd/gh/main.go

PR #2205, opened 2019-07-23
Simplify code

muesli:
- Use bytes.Equal instead of bytes.Compare
- Use time.Since instead of manual timestamp subtracting
- Use sort.Strings instead of sort.StringSlice
- No need to specify the len of the slice


mislav:
Excellent finds; thank you!

PR #2207, opened 2019-07-23
Trap syscall.SIGTERM instead of os.Kill

muesli:
os.Kill can not be signal trapped. I assume we want to trap syscall.SIGTERM
instead?

### DIFF
--- a/github/config.go
+++ b/github/config.go
@@ -192,7 +192,7 @@ func getPassword() (string, error) {
 	}
 
 	c := make(chan os.Signal)
-	signal.Notify(c, os.Interrupt, os.Kill)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
 	go func() {
 		s := <-c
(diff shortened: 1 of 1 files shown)


github/config.go
```diff
@@ -192,7 +192,7 @@ func getPassword() (string, error) {
 	}
 
 	c := make(chan os.Signal)
-	signal.Notify(c, os.Interrupt, os.Kill)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
```
mislav: Basically, we want to restore terminal state if <kbd>Ctrl-C</kbd> or <kbd>Ctrl-D</kbd> are used. Since <kbd>Ctrl-C</kbd> is already handled as `os.Interrupt`, I guess it's okay to handle the additional TERM signal (you're right, Kill can't be handled).

mislav:
Good catch; thank you!

PR #2211, opened 2019-07-26
Tweak Makefile to support older Go versions' backported module support

vilmibm:
`go` supports environmentally-toggled support for the module system introduced in `1.11`.

Currently the Makefile checks for go > `1.10` to decide whether or not to use the `-mod` flag; for certain `go` installations, though, this flag is unsupported without setting `GO111MODULE=on` in the environment. 

Compounding this is having a module that depends on go >= `1.12`. If modules are disabled, everything builds fine; if they are enabled any Go version under `1.12` (even if it has backported module support) will fail.

This PR:

- only enables module use for `go` >= `1.12`
- sets `GO111MODULE=on` if modules are enabled

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,12 +3,13 @@ SOURCE_DATE_EPOCH ?= $(shell date +%s)
 BUILD_DATE = $(shell date -u -d "@$(SOURCE_DATE_EPOCH)" '+%d %b %Y' 2>/dev/null || date -u -r "$(SOURCE_DATE_EPOCH)" '+%d %b %Y')
 HUB_VERSION = $(shell bin/hub version | tail -1)
 FLAGS_ALL = $(shell go version | grep -q 'go1.[89]' || echo 'all=')
-export MOD_VENDOR_ARG := $(shell go version | grep -q 'go1.1[^0]' && echo '-mod=vendor')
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -115,7 +115,14 @@ make install prefix=/usr/local
 Prerequisites for compilation are:
 
 * `make`
-* [Go 1.9+](http://golang.org/doc/install)
+* [Go 1.11+](http://golang.org/doc/install)
```
mislav: This project builds on Go 1.9 just fine; we use it in CI.

README.md
```diff
@@ -115,7 +115,14 @@ make install prefix=/usr/local
...
+
+Depending on your environment, you may encounter a `build flag -mod=vendor only valid when using
```
mislav: Instead of providing build instructions like “sometimes you need this; sometimes you need that”, how about we simply change our build step (i.e. `make`) to work no matter the environment?

Ref. https://github.com/github/hub/pull/2142

vilmibm: 👍 makes sense to me

README.md
vilmibm: ah, interesting -- i made this change since i was unable to build without module support and saw that that was introduced after 1.9 but i see it was backported 👍

mislav:
Excellent; love the new approach! ❤️

Issue #1359, opened 2016-12-09
hub sync isn't very effective when squashing and merging commits


seanabraham:
I want to use `hub sync` to update local branches but more importantly delete branches that have been merged already via the Github UI. The problem is, when running `hub sync` I get many warnings like the following:
`warning: 'sean/branch_name' was deleted on origin, but appears not merged into master`.

I'm assuming this is because the PR opened for this branch was merged using the "Squash and Merge" option so the SHA no longer matches. Is there a way to force `hub sync` to still delete branches in this case?

mislav:
Sorry, `hub sync` still doesn't recognize squash-merged branches! I'm not sure what would be the best way to support this, since it's not trivial to detect just by looking at git objects.

If anyone has any ideas, I'm all ears.

chufucious:
Is there a command I can run to get rid of that issue?

kevinkjt2000:
I copied a complicated git alias from somewhere that helps clean up squashed branches.  It may or may not work for your use case:

```
[alias]
  cleanup = "!git remote prune origin && git checkout -q master && git for-each-ref refs/heads/ '--format=%(refname:short)' | while read branch; do mergeBase=$(git merge-base master $branch) && [[ $(git cherry master $(git commit-tree $(git rev-parse $branch^{tree}) -p $mergeBase -m _)) == '-'* ]] && git branch -D $branch; done"
```

In my case, I run `hub sync` followed by `git cleanup`.

I am not familiar at all with hub's code, but perhaps this logic of looping over refs and comparing merges could be integrated into the `sync` command?

PR #2213, opened 2019-08-01
Sync branch deletions after a squash merge

awarrenlove:
This is my thought on a possible solution to https://github.com/github/hub/issues/1359.

It provides a `--delete-all` option to the sync command that will attempt to delete all branches that were deleted on the remote, even those that do not appear merged. This will let us delete branches that were merged through a squash commit. By itself, the option prompts for confirmation on every branch similar to the `delete` command with the same override option `--yes` that will confirm all branches. I'm open to suggestions on better naming for that option; I'm not a big fan of it but haven't thought of anything better.

This also refactors the confirmation logic into the `utils` in order to allow sharing the same approach across both commands. Because in my opinion the `(yes/N)` piece of the prompt really belongs with the confirmation itself rather than the caller prompt, the order of the output switched slightly in the `delete` command and its test.

Side note: I'm not a Go developer so please feel free to be pedantic with any comments or suggestions here.


awarrenlove:
Please let me know if I can provide any more context. Like I said in the description, I don't particularly like the `--delete-all` option name either, so if you have an idea for that I would love to hear it.

jeffg-hpe:
`--squashed` or `--include-squashed` might make sense

awarrenlove:
@jeffg-hpe I thought about something specific to squashed, but technically this deletes all branches that don't look merged, not only those that were squashed, so I believe it would also include all branches deleted on the remote that really weren't merged too. That's why I went with "all," since it is deleting all of them to match the remote.

doronbehar:
So maybe `--delete-stray`? Inspired by [git-trim](https://github.com/foriequal0/git-trim).

PR #2214, opened 2019-08-01
Pass $EDITOR to the shell

### LINKED ISSUE #1482: Complex git core.editor setting fails with hub
I have my `core.editor` set to something complex (to get it to work cross machine):

```
[core]
    editor = "which mvim > /dev/null 2>&1 && mvim --remote-tab-wait-silent $@ || vim"
```

When trying `hub create issue`, I get:

```
/Users/ehesselink/.local/bin/mvim
/Users/ehesselink/.local/bin/mvim
/usr/bin/vim
error using text editor for issue message
```

The setting works with plain git though. I see this was discussed before in #881, but that was marked as fixed.

```
$ hub version
git version 2.9.2
hub version 2.2.9
```

bk2204:
Several attempts have been made to provide better handling of the EDITOR and VISUAL environment variables. However, because these commands can contain arbitrary shell code, it isn't possible to provide correct behavior without actually invoking the shell.

Create a new function that invokes the shell with the proper arguments. For all editor invocations, invoke the shell in the way that Git does, even though this is incompatible with the way that other software behaves. For example, Git allows the use of "$@" directly in the command, while most other programs, including Debian's `sensible-editor` and Git LFS, require the user to wrap such commands in a shell function. It isn't possible to make it work in both cases, but we're trying to be like Git, so do what Git does.

Rewrite the command line in verbose output so we don't show the user the doubled argument we need to pass for things to work correctly. Also, stop expanding environment variables in the editor code path, since the shell will do that for us.

I'm not 100% certain that this will catch all Windows Git installations; it's possible that there are other custom ones that won't install things the same way. However, since Git doesn't tell us what the path is for its `sh`, we can't do anything but guess. This does work fine on macOS and Linux, and should be fine on all other Unix systems as well.

Fixes #1482.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"syscall"
@@ -17,6 +18,7 @@ type Cmd struct {
 	Stdin  *os.File
 	Stdout *os.File
 	Stderr *os.File
+	Shell  bool
 }
 
 func (cmd Cmd) String() string {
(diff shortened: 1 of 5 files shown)


mislav:
Excellent explanation and work; thank you. Please give me some more time to review and test; I'm just back from vacation and catching up. 🙇

PR #2216, opened 2019-08-03
README: Replace "Linuxbrew" with "Homebrew on macOS or Linux"

issyl0:
- We're (slowly) moving away from the Linuxbrew name, now that it's officially a part of Homebrew.


mislav:
Thank you!

PR #2218, opened 2019-08-03
Fix error message on hub --paginate

tpope:
Calling hub with one or more options but no subcommand resulted in the
following error message trying to expand an empty string as an alias:

	error: key does not contain variable name: alias.

This fixes the root cause: attempting to retrieve an invalid config key.


mislav:
Thank you!

PR #2230, opened 2019-08-17
Introduce flag to prepare pull request message from commit

mathewcohle:
Inspired by "git commit --reuse-message", it might be useful to use
particular commit message when opening new pull request.

Introduced flag also respects pull request template if it is defined in
the repository.

------

Sorry that I've not created an issue first but I'm actually willing to keep
this patch around even if it will not be merged for some reason as I find this
feature quite useful for my current workflow.

While writing it I've noticed that it should be possible to improve the code
around creating pull request message as it bit inconsistent right now (e.g.
using `--no-edit` will behave similar to introduced flag if branch contains
just one commit) but I guess it is better to discuss these firstly.


commands/pull_request.go
```diff
@@ -35,6 +36,9 @@ pull-request -i <ISSUE>
 		If multiple <MESSAGE> options are given, their values are concatenated as
 		separate paragraphs.
 
+	-C, --commit <COMMIT>
```
mislav: Perhaps the expanded flag name could also be `--reuse-message` to match `git commit`?

mathewcohle:
@mislav Thanks for review:

* `--commit` -> `--reuse-message`
* fixed failing test (actually it catched the change of output if `pull_request_template` was not available), good job having this covered
* I've added test in `features/pull_request.feature` covering the introduced flag works as expected

Let me know if there is anything else to update.

features/pull_request.feature
```diff
@@ -538,6 +538,42 @@ Feature: hub pull-request
     Then the output should contain exactly "https://github.com/mislav/coral/pull/12\n"
     And the file ".git/PULLREQ_EDITMSG" should not exist
 
+  Scenario: Title and body from commit
```
mislav: Can you please rename this scenario "Title and body from commit when PR template exists" and create a copy of this scenario, but a simpler version of it where the PR template doesn't exist in the first place and the message is simply taken from the commit message? Thank you!

mathewcohle: Done :)

FYI it seems that CI tests are actually non-deterministic. I've encountered random failed test of untouched part of the codebase which were resolved in another run - see https://travis-ci.org/github/hub/builds/574154296 and https://travis-ci.org/github/hub/builds/574155640 (the code was exactly the same, I've just amend commit msg & push to trigger CI)

commands/pull_request.go
```diff
@@ -467,6 +472,33 @@ func parsePullRequestIssueNumber(url string) string {
 	return ""
 }
 
+func prepareMessageFromCommit(commit string) (string, error) {
+	message, err := git.Show(commit)
+	if err != nil {
+		return "", err
+	}
+	message = strings.Replace(string(message), "\r\n", "\n", -1)
```
mislav: Nit: perhaps you can use `ReplaceAll()`

commands/pull_request.go
```diff
@@ -467,6 +472,33 @@ func parsePullRequestIssueNumber(url string) string {
...
+
+	message = appendPullRequestTemplate(message)
```
mislav: I just realized that we probably don't want to be using a pull request template here:

1. If `--edit` wasn't passed (off by default), the person would not even have the chance to edit the template. Submitting the contents of a non-edited template is most likely wrong.
2. If `--edit` was passed, I still don't think we want to apply the template, because the subject and the body of the given commit acts as a "template" already.

Could you amend the behavior and tests please if this makes sense to you?

mathewcohle: Well, one of the motivation behind the implementation is (as it's probably most often the case) to scratch my pain point. Currently, most of the time when I'm opening PR I need:

* to use HEAD or some commit in branch as "base" for PR title and body (as I'm used to write "meaningful" commits so I can reuse them for this purpose - as I want reasonable starting point when describing the pull request)
* add information which we have in PR template (i.e. I want each PR to follow the structure as defined in the template)

I'm doing it right now by running something like `my_custom_script > file && hub pull-request -F file --edit`, where `my_custom_script`:

* takes HEAD commit and add it at the beginning of file
* parse and update PR template from repo

and I just update details if something is missing/redundant in manual edit. 

So I was thinking at the beginning about adding `-C` to `hub pull-request -F` but as I've found  that the behavior is bit inconsistent, it looks to me easier to introduce it separately and eventually in another PR unify the behavior across different flags. 

One possible solution I see here - which would keep around adding PR template to PR message - might be introducing another flag like `[no]template` so even if the user does not pass `--edit`, she might decide to skip the template. I am not sure what is preferable default behavior so let me know what you would prefer. 

Anyway, in case that this solution would not work for you, I will finish the PR as you propose but it would be nice to think about a way how it would be possible to implement the features which would allow to use `hub` in use case described above. I'm definitely open to implementing them in another PRs :)

mislav: > introducing another flag like `[no]template` so even if the user does not pass `--edit`, she might decide to skip the template. I am not sure what is preferable default behavior so let me know what you would prefer.

Thanks for explaining your use-case. I'm wary of adding more flags and configuration options such as `--[no]template`. From my standpoint, when someone's workflow becomes too specific to be sufficiently covered by hub core behavior, they should resort to using scripts to encapsulate their workflow and use `hub pull-request -F <file>` to pass contents in programmatically. You have already gone down that route, it seems.

It's also nice that some of these features would be available in core without the necessity of scripting them outside of hub, but I'm not sure what the right behavior should be here. Maybe we can add the template when `--edit` was present but skip it without. (I still think that posting a template without editing is never desireable.) Maybe I need to think about this for a while.

mathewcohle: > I'm wary of adding more flags and configuration options such as --[no]template

completely understand that, it's always a pain reading help/man with never ending list of flags

Agree with you on both points (specific behavior -> use custom script vs feature in core), I actually cannot say what is the right way here; it seems to me that somehow using commit message when creating PR is reasonable (as github is doing the same when you open PR using frontend).

Regarding edit/noedit behavior, personally I cannot see when/why you want to run `hub` without `--edit` (yes, you can always edit it once PR is created but) - if nothing else I would just skim through the title/body to check if it makes sense - whatever hub workflow you use. But that's obviously just my personal POV.

 > Maybe we can add the template when --edit was present but skip it without.

This would definitely works for me, but I somehow feel like if we are not making things "too complicated" / difficult to understand from user perspective (unless she reads the actual implementation).

> Maybe I need to think about this for a while.

Sure, np, let me know once you have some idea how to continue with it - I will finish PR then :)

Thanks for all the insights.

Issue #2234, opened 2019-08-19
`hub api` should replace all instance strings


OliverJAsh:
I am trying to write a script that gets the PR number for the current branch. This should work:

``` bash
#!/bin/bash

REPO_PATH="repos/{owner}/{repo}"
BRANCH=`git branch --show-current`
hub api "$REPO_PATH/pulls?head={owner}:$BRANCH" | jq '.[0].number'
```

However it doesn't, and I believe that's because the second instance of the `{owner}` literal string is not replaced. I feel like it should be?

Secondly, it would be useful to have a `--verbose` flag for this command, so we can see what the generated request URL looks like.

PR #2238, opened 2019-08-21
hub api url generation fixes

whuang8:
Fixes https://github.com/github/hub/issues/2234

- [x] Replace all occurrences of `{owner}` and `{repo}` in api paths and GraphQL queries
- [x] Add verbose mode to `hub api` to see generated URL

### DIFF
--- a/commands/api.go
+++ b/commands/api.go
@@ -193,8 +193,8 @@ func apiCommand(cmd *Command, args *Args) {
 	isGraphQL := path == "graphql"
 	if isGraphQL && params["query"] != nil {
 		query := params["query"].(string)
-		query = strings.Replace(query, quote("{owner}"), quote(owner), 1)
-		query = strings.Replace(query, quote("{repo}"), quote(repo), 1)
+		query = strings.Replace(query, "{owner}", owner, -1)
+		query = strings.Replace(query, "{repo}", repo, -1)
 
 		variables := make(map[string]interface{})
 		for key, value := range params {
@@ -209,8 +209,8 @@ func apiCommand(cmd *Command, args *Args) {
 
 		params["query"] = query
 	} else {
-		path = strings.Replace(path, "{owner}", owner, 1)
-		path = strings.Replace(path, "{repo}", repo, 1)
+		path = strings.Replace(path, "{owner}", owner, -1)
+		path = strings.Replace(path, "{repo}", repo, -1)
 	}
 
 	var body interface{}
@@ -329,8 +329,4 @@ func readFile(file string) (content []byte) {
 	}
 	utils.Check(err)
 	return
-}
-
-func quote(s string) string {
-	return fmt.Sprintf("%q", s)
 }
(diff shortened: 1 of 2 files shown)


commands/api.go
```diff
@@ -193,8 +196,8 @@ func apiCommand(cmd *Command, args *Args) {
 	isGraphQL := path == "graphql"
 	if isGraphQL && params["query"] != nil {
 		query := params["query"].(string)
-		query = strings.Replace(query, quote("{owner}"), quote(owner), 1)
```
whuang8: What is the purpose of using `quote()` here? Isn't replacing a word with quotes with another word with quotes the same as just replacing a word with another word?

I am not familiar with GraphQL syntax so I might be missing something here :)

whuang8: It looks like these 2 lines are the only places invoking `quote()`. So if this conversation is resolved with the decision to not use it, I can remove the function definition at the bottom of the file.

commands/api.go
```diff
@@ -209,8 +212,8 @@ func apiCommand(cmd *Command, args *Args) {
 
 		params["query"] = query
 	} else {
-		path = strings.Replace(path, "{owner}", owner, 1)
-		path = strings.Replace(path, "{repo}", repo, 1)
+		path = strings.Replace(path, "{owner}", owner, -1)
+		path = strings.Replace(path, "{repo}", repo, -1)
```
mislav: Could we use ReplaceAll instead?

commands/api.go
mislav: The reason I chose to replace only the quoted `"{owner}"` and `"{repo}"` placeholders  was to indicate that these end up being hardcoded string values within the GraphQL query, rather than accidentally replacing the non-quoted placeholder `{owner}` somewhere where it's not its own string literal. Additionally, if the expanded values contained a literal quote `"` themselves, this would escape it. (It's impossible for either owner handle or repo name to contain special characters such as a quote, though.)

I'm not sure whether this guard or limitation actually pays off, and it definitely prevents a person from using both `{owner}` and `{repo}` within the same string literal. Maybe you're right and we should go the simpler way and drop the `quote()` for now.

commands/api.go
```diff
@@ -67,6 +67,9 @@ var cmdApi = &Command{
 		Parse response JSON and output the data in a line-based key-value format
 		suitable for use in shell scripts.
 
+	-v, --verbose
```
mislav: Let's not add `-v` for now. Hub already has a verbose mode:

```
HUB_VERBOSE=1 hub api ...
```

If we added a flag to enable this mode instead of the environment variable, we would have to support this flag across all commands, not just this one.

commands/api.go
whuang8: `ReplaceAll` only works with Go 1.12. I initially implemented this using ReplaceAll but tests running on older versions of Go broke in Travis.

whuang8:
@mislav Review comments addressed. Please take another look when time allows. Thanks!

mislav:
Thank you!

Issue #2249, opened 2019-08-30
Document how to use hub CLI on GitHub Actions


fregante:
Apparently it should work now but it's not clear how to get there: https://github.com/github/hub/issues/2149

Can you document it somewhere?

mislav:
Good idea. I haven't used hub from within Actions yet, so I'm summoning @bdougie @Geertvdc @rtimush @CLowenthal-zmi @LRWeber in case any of you have anything to add (but feel free to ignore). Off the top of my head, these are the steps that we'd need to document:

1. First, your action needs to obtain the `hub` executable somehow in its environment.

   In your Dockerfile you could use a package manager such as `apt-get install` for Debian, but that is likely to get you an outdated version of hub.

   Another approach is to vendor hub in your project and set it up manually like demonstrated here https://github.com/LLNL/llnl.github.io/commit/2046d13c6321d40725fc8c1f5c33ee65ddcc9933. 

   There is also this, which looks promising but that I still need to look into: https://github.com/marketplace/actions/setup-hub

2. Second, hub needs the GITHUB_TOKEN environment variable. GitHub can generate and provide this token for your actions, but you need to pull in the secret manually in your steps: https://help.github.com/en/articles/virtual-environments-for-github-actions#github_token-secret

3. Third, you should set the GITHUB_USER environment variable to reference the actor for this action. It can simply be set to the owner of the current repo.

   This step is awkward and, even though it's relatively simple, it can be tedious to remember and repeat all the time. It was introduced as a workaround for the fact that hub will try to access the `/user` API endpoint to identify the current actor, but this endpoint is not available to actions.

   I came to understand that during actions runs there is also [a default environment variable](https://help.github.com/en/articles/virtual-environments-for-github-actions#environment-variables) called "GITHUB_REPOSITORY". We could infer GITHUB_USER from the first part of GITHUB_REPOSITORY and thus free people from having to set this value. I will try to make this addition to hub. It would be great if there was no step 3!

Geertvdc:
Hi @mislav, tnx for the mention. I'm the creator of the setup-hub action which installs the hub cli on your worker.

https://github.com/marketplace/actions/setup-hub

In that Readme it explains how to set it up. Basically you run the action to setup hub once once and afterwards you can just use it in your script steps.

If any questions on this action let me know (it was the first github action I created)

LRWeber:
[...]

PR #1585, opened 2017-10-08
add retriever.sh, #987: cross-platform download script

javier-lopez:
Hello,

This is my attempt to provide a cross-platform download script for hub, it works like this:

```
$ sh ./retriever.sh 
hub-linux-amd64-2.2.9.tgz
hub-linux-amd64-2.2.9/
```

It should work on windows/macos/linux/freebsd but I only tested on Linux, Ubuntu 16.06

Bests

PR #2259, opened 2019-09-09
Add install script to use hub more easily in github actions

vibou:
This script is directly inspired by https://github.com/helm/helm/blob/master/scripts/get

You can use it to install hub in command line easily using the following command:

`curl -qs https://raw.githubusercontent.com/vibou/hub/master/install.sh | env DESIRED_VERSION=2.12.3 bash -`

 If no version is specified the latest is used

`curl -qs https://raw.githubusercontent.com/vibou/hub/master/install.sh | bash -`

The script automatically detects the OS and the ARCH to download the correct release asset.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,212 @@
+#!/usr/bin/env bash
+
+PROJECT_NAME="hub"
+
+: ${USE_SUDO:="true"}
+: ${HUB_INSTALL_DIR:="/usr/local/bin"}
+
+# initArch discovers the architecture for this system.
+initArch() {
+  ARCH=$(uname -m)
+  case $ARCH in
+    armv5*) ARCH="armv5";;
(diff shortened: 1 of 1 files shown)


shouze:
deno has made a dedicated repository recently for this kind of scripts https://github.com/denoland/deno_install WDYT? 🤔

mislav:
@vibou Thank you for the script!

I have a sense that the script might be a bit too long for the purpose of just making hub available in the current environment, usually Actions. Since hub in CI is only used in one-off jobs, I don't think it makes sense to install it system-wide. Instead, downloading just the `bin/hub` executable would do.

What do you think of an approach like in https://github.com/github/hub/issues/2249#issuecomment-548008147?

mislav:
See also https://github.com/github/hub/pull/1585

PR #2260, opened 2019-09-09
add git log entries to pr edit mode with template

eputnam:
currently, when I use --file and --edit to open a pull request, I don't
see the commits contained in the PR like I would without --file/--edit.
I'd really like to have those as a reference for filling out my
template, so this change adds the git log entires into the messageBuilder even when --file has been received

### DIFF
--- a/commands/pull_request.go
+++ b/commands/pull_request.go
@@ -257,6 +257,18 @@ of text is the title and the rest is the description.`, fullBase, fullHead))
 	} else if args.Flag.HasReceived("--file") {
 		messageBuilder.Message, err = msgFromFile(args.Flag.Value("--file"))
 		utils.Check(err)
+		headForMessage := headTracking
+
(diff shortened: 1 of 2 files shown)


eputnam:
@mislav any thoughts on this?

mislav:
Thank you for your contribution and sorry for the late reply! 

I see `--file` as a lower-level interface, meaning that I prefer that it has precise behavior and not much magic. If you are passing a `--file`, it should ideally already be authored correctly. The `--edit` flag is here for the last-minute changes, but unlike the case where we completely author the message in the text editor from scratch (no `-m` nor `--file` were passed), in the case of `--file --edit` I don't agree that we should pull up the commit log.

Also, if we were to go this route, then we'd have to extend the same behavior to `-m msg --edit` as well, and I think that would lead to this whole logic getting more complicated than it already is.

Issue #2226, opened 2019-08-15
general gist support


jaymzh:
I know one can do janky jq stuff as mentioned in https://github.com/github/hub/issues/2116 but it'd be great to be able to create and grab gists nicely and easily with `hub` directly:

```
cat myfile | hub gist --public
hub gist [gist_url] > myfile
``

**Comment by mislav** [comment]
You could also use https://github.com/mislav/hub-api-utils/blob/master/bin/hub-gist.

From my comment on https://github.com/github/hub/issues/2116#issuecomment-485529449:

> Creating or editing gists is not anywhere near on our roadmap for hub so I'm going to close this for now.

**Comment by jaymzh** [comment]
@mislav - your example does not work, when you run it you get

```
{"message":"Not Found","documentation_url":"https://developer.github.com/v3/gists/#create-a-gist"}
```

However the JSON looks like what the documentation claims it should look like. :/

**Comment by mislav** [comment]
@jaymzh I just ran my `hub-gist` command linked above and it works. Can you check that your hub token has "gist" among scopes? https://github.com/settings/tokens

**Comment by jaymzh** [comment]
@mislav - ah I didn't know that was a thing. I added gist, and now I get:

```
$ /tmp/hub-gist /etc/hosts
{"message":"Validation Failed","errors":[{"resource":"Gist","code":"invalid","field":"files"},{"resource":"Gist","code":"custom","field":"Whoops,","message":"Whoops, files can't contain malformed path components."}],"documentation_url":"https://developer.github.com/v3/gists/#create-a-gist"}
```

jaymzh:
Not sure why `hub-gist` doesn't work, but I've sent a PR to add nice gist support to hub directly that now works with both creation and retrieval of gists: https://github.com/github/hub/pull/2269

Issue #2116, opened 2019-04-21
Gist upload 


gokceneraslan:
It'd be great to be able to upload a file as public or private Gist using hub. When used together with nbviewer it'd be a super convenient way to share Jupyter notebooks.

us:
Good idea! 

Please give a `future` label.

mislav:
You can already relatively easily create a gist using `hub api`. The only tricky part is that the JSON payload needs to be manually constructed. In this example I do that with jq:

```sh
gist() {
  local filename
  local public=false
  if [ "${1:-}" = "--public" ]; then
    public=true
    shift 1
  fi
  filename="${1?}"
  jq -nR --arg filename "$filename" --argjson public "$public" '
    {"files":{($filename):{"content":input}},public:$public}
  ' "$filename" | hub api gists --input - | jq -r '.html_url'
}

gist myfile.txt
gist --public myfile.txt
```

Creating or editing gists is not anywhere near on our roadmap for hub so I'm going to close this for now.

theduke:
Shame, this would be really useful functionality.

PR #2269, opened 2019-09-14
Make a gist command

jaymzh:
This addresses #2226 and #2116 by creating a `gist` subcommand.

This allows you to both create and retrieve gists. When creating a gist it can be public or private.

In order to do this, the `gist` scope is added to the token we request.

Closes: #2226
Closes: #2116

Signed-off-by: Phil Dibowitz <phil@ipom.com>

### DIFF
--- a/commands/gist.go
+++ b/commands/gist.go
@@ -19,14 +19,6 @@ cat <file> | gist create [--public]
 `,
 		Long: `Create a GitHub Gist
 
-Can both create and retrieve gists. With no arguements, it takes a file on
-stdin and creates a gist. With multiple '--file' arguments, will create a
-mult-file gist.
-
-If gistid is passed in, if there is only one file in the gist, it will be
-printed, otherwise will error if a specific file is not requested. However, if
-'--json' is used, then multiple files can be retreived.
-
 ## Commands:
 
     * _show_:
--- a/commands/gist.go
+++ b/commands/gist.go
@@ -2,7 +2,7 @@ package commands
 
 import (
 	"fmt"
-	"os"
+	"sort"
 
 	"github.com/github/hub/github"
 	"github.com/github/hub/ui"
@@ -80,33 +80,42 @@ func init() {
 	CmdRunner.Use(cmdGist)
 }
 
-func getGist(gh *github.Client, id string, filename string, no_headers bool) {
-	gist, err := gh.FetchGist(id, filename)
-	utils.Check(err)
+func getGist(gh *github.Client, id string, filename string, no_headers bool) error {
+	gist, err := gh.FetchGist(id)
+	if err != nil {
+		return err
+	}
+
 	if filename != "" {
-		if val, err := gist.Files[filename]; err {
-			fmt.Printf("%s\n", val.Content)
+		if val, ok := gist.Files[filename]; ok {
+			ui.Println(val.Content)
 		} else {
-			fmt.Printf("No such file in that Gist\n")
-			os.Exit(1)
+			return fmt.Errorf("no such file in gist")
 		}
 	} else {
 		print_hdrs := len(gist.Files) != 1 && !no_headers
 		if print_hdrs {
-			fmt.Printf("GIST: %s (%s)\n\n", gist.Description, gist.Id)
+			ui.Printf("GIST: %s (%s)\n\n", gist.Description, gist.Id)
 		}
-		for name, file := range gist.Files {
+
+		filenames := []string{}
+		for name := range gist.Files {
+			filenames = append(filenames, name)
+		}
+		sort.Strings(filenames)
+
+		for _, name := range filenames {
+			file := gist.Files[name]
 			if print_hdrs {
-				fmt.Printf("==== BEGIN " + name + " ====>\n")
+				ui.Printf("==== BEGIN %s ====>\n", name)
 			}
-			// yes, the printf is necessary to prevent it interpolating
-			// '%'s in the string as part of a format string.
-			fmt.Printf("%s\n", file.Content)
+			ui.Println(file.Content)
 			if print_hdrs {
-				fmt.Printf("<=== END " + name + " =======\n")
+				ui.Printf("<=== END %s =======\n", name)
 			}
 		}
 	}
+	return nil
 }
 
 func gist(cmd *Command, args *Args) {
--- a/commands/gist.go
+++ b/commands/gist.go
@@ -1,8 +1,8 @@
 package commands
 
 import (
-	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/github/hub/github"
 	"github.com/github/hub/ui"
@@ -14,7 +14,7 @@ var (
 		Run: printGistHelp,
 		Usage: `
 gist create [--public] [<FILES>...]
-gist show [--json] <ID> [<FILENAME>]
+gist show <ID> [<FILENAME>]
 `,
 		Long: `Create and print GitHub Gists
 
@@ -26,16 +26,12 @@ gist show [--json] <ID> [<FILENAME>]
 
     * _show_:
 		Print the contents of a gist. If the gist contains multiple files, the
-		operation will error out unless either <FILENAME> is specified or the
-		'--json' flag is used.
+		operation will error out unless <FILENAME> is specified.
 
 ## Options:
 
     --public
         Make the new gist public (default: false).
-
-    --json
-        Print all files in the gist and emit them in JSON.
 
 ## Examples:
 
@@ -56,7 +52,7 @@ hub(1), hub-api(1)
 	cmdShowGist = &Command{
 		Key:        "show",
 		Run:        showGist,
-		KnownFlags: "--json",
+		KnownFlags: "\n",
 	}
 
 	cmdCreateGist = &Command{
@@ -72,33 +68,27 @@ func init() {
 	CmdRunner.Use(cmdGist)
 }
 
-func getGist(gh *github.Client, id string, filename string, emitJson bool) error {
+func getGist(gh *github.Client, id string, filename string) error {
 	gist, err := gh.FetchGist(id)
 	if err != nil {
 		return err
 	}
 
-	if len(gist.Files) > 1 && !emitJson && filename == "" {
-		return fmt.Errorf("There are multiple files, you must specify one, or use --json")
+	if len(gist.Files) > 1 && filename == "" {
+		filenames := []string{}
+		for name := range gist.Files {
+			filenames = append(filenames, name)
+		}
+		return fmt.Errorf("the gist contains multiple files, you must specify one:\n%s", strings.Join(filenames, "\n"))
 	}
 
-	if emitJson {
-		data, err := json.Marshal(gist.Files)
-		if err != nil {
-			return err
-		}
-		fmt.Printf("%s\n", data)
-	} else if filename != "" {
+	if filename != "" {
 		if val, ok := gist.Files[filename]; ok {
 			ui.Println(val.Content)
 		} else {
 			return fmt.Errorf("no such file in gist")
 		}
 	} else {
-		/*
-		 * There's only one, but we don't know the name so a
-		 * loop us fine.
-		 */
 		for name := range gist.Files {
 			file := gist.Files[name]
 			ui.Println(file.Content)
@@ -131,16 +121,18 @@ func createGist(cmd *Command, args *Args) {
 
 func showGist(cmd *Command, args *Args) {
 	args.NoForward()
-
-	host, err := github.CurrentConfig().DefaultHostNoPrompt()
-	utils.Check(err)
-	gh := github.NewClient(host.Host)
-
+	if args.ParamsSize() < 1 {
+		utils.Check(cmd.UsageError("you must specify a gist ID"))
+	}
 	id := args.GetParam(0)
 	filename := ""
 	if args.ParamsSize() > 1 {
 		filename = args.GetParam(1)
 	}
-	err = getGist(gh, id, filename, args.Flag.Bool("--json"))
+
+	host, err := github.CurrentConfig().DefaultHostNoPrompt()
+	utils.Check(err)
+	gh := github.NewClient(host.Host)
+	err = getGist(gh, id, filename)
 	utils.Check(err)
 }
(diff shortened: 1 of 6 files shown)


jaymzh:
I'm very new to golang, so I'm sure there's lots of cleanups I can do - just let me know.

jaymzh:
the only test that's failing is test coverage dropped below a certain threshold. I'm not really clear on how best to write tests for this, so some guidance would be appreciated.

jaymzh:
OK now that I've added the right scopes, I can create gists - but there's a few bugs in gist creation. Updated PR coming soon.

jaymzh:
OK, this PR now works for creating and getting gists, and can do both public and private. It also adds the `gist` scope to our token. It's fully functional as far as I can tell.

As mentioned above, the only tests failing are test-coverage ones, and I'm happy to write some tests if someone gives me some guidance on what units they think are nicely testable here or points me to an example that might be relevant.

mislav:
This is the way we typically author tests: https://github.com/github/hub/blob/4aba5854179b8e381bebd08ec9b2d49a3613eb15/features/release.feature#L370-L388

You will notice you'll have to stub a minimal stub of a GitHub API server since we don't communicate with real GitHub API when running tests.

You can observe other scenarios for examples how to e.g. initialize a file that you might pass to `hub gist` command: https://github.com/github/hub/blob/4aba5854179b8e381bebd08ec9b2d49a3613eb15/features/api.feature#L243-L246

Good luck!

jaymzh:
Thanks @mislav ! There's now pretty thorough test coverage and all tests pass.

jaymzh:
Any concerns with this? I'd love to get this feature merged now that there's tests.

commands/gist.go
```diff
@@ -0,0 +1,201 @@
...
+type GistResponse struct {
+	HtmlUrl string `json:"html_url"`
+}
+
+func getGist(gh *github.Client, id string, filename string, no_headers bool) {
+	path := "gists/" + id
+	var headers map[string]string
+	cacheTTL := 0
+
+	response, err := gh.GenericAPIRequest("GET", path, "", headers, cacheTTL)
```
mislav: You should not use `GenericAPIRequest`. That method is intended for the `hub api` command. Instead, use:

```go
	res, err := api.Get(fmt.Sprintf("gists/%s", id))
	if err = checkStatus(200, "getting gist", res, err); err != nil {
		return
	}
```

See other methods in `github/client.go` for inspiration

commands/gist.go
```diff
@@ -0,0 +1,201 @@
...
+	"io/ioutil"
+	"os"
+	"path"
+	"strings"
+)
+
+var cmdGist = &Command{
+	Run: gist,
+	Usage: `
+gist <GistID> [--no-headers] [<filename>]
```
mislav: I think printing the contents of an existing gist might not be the functionality that I'd be interested in having in hub right now. I simply don't think it would get much use when compared to creating gists. Furthermore, printing gists with multiple files is awkward, as evident by the need for header lines (which look like git conflicts) and the extra `--no-headers` and `<filename>` arguments.

Would you be open to cutting down the functionality of this command to only support creating gists?

commands/gist.go
```diff
@@ -0,0 +1,201 @@
...
+gist [--public] --file <file>
```
mislav: When creating gists, I think this should be the preferred usage:

```
hub gist <file1> <file2...>
```

i.e. files are explicitly listed as arguments, without needing the `--file` flag. And if there are no files listed as arguments, stdin is used. What do you think?

github/client.go
```diff
@@ -873,7 +873,7 @@ func (client *Client) FindOrCreateToken(user, password, twoFactorCode string) (t
 	}
 
 	params := map[string]interface{}{
-		"scopes":   []string{"repo"},
+		"scopes":   []string{"repo", "gist"},
```
mislav: What about current hub users who already have authenticated and have a token, but their token doesn't have the "gist" scope? For them, all calls to `hub gist` will fail with an HTTP error. Maybe we could handle that error and suggests that they open `github.com/settings/token` and edit the scopes to include "gist".

commands/gist.go
jaymzh: The use-cases I'm going here pretty directly need that. Let me lay them out a bit:

* user says "thing isn't working" and does `./my_command | hub gist` and hand that URL to an engineer who can then do `hub gist <thing>`. This is incredibly useful.
* engineer says "hey check out this thing I'm trying to do" and does `cat foo.sh | hub gist` and engineer2 does `hub gist <thing> > /tmp/foo.sh` and runs `foo.sh` to help debug.

The ability to get these back out on the terminal is really really important to the point of writing this module.

I'm not sure why options seem like a downside... configurability seems better to me.

commands/gist.go
jaymzh: that only works if we drop the ability to fetch gists, which, as described above, defeats the purpose.

I'm OK if we want to say `--files <file1> <file2>` - That said in all my time using GH I've never ever seen anyone use more than one file in a Gist... I didn't even register it was possible until I started writing this... it's pastebin for output and code.

github/client.go
jaymzh: I'd love to do that! As you probably saw I spent a lot of time not able to understand why this wasn't working and thought the API was broken. I'll figure out how to do that.

jaymzh:
@mislav I either updated the code or answered your questions for each of your requests.

mislav:
Pushed some improvements to your branch. Hope that's ok! Let me know what you think.

> The ability to get these back out on the terminal is really really important to the point of writing this module.

Thanks for outlining your use cases for gist fetch functionality. It sounds like most of the cases you're describing involve single-file gists. I'm still convinced that printing multi-file gists will be awkward, since the headers included in the output are gnarly and hardly machine-readable, and with `--no-headers` the output is basically useless because multiple (potentially unrelated, potentially in different languages) files would be concatenated together. I also don't see much use-case for the `hub gist <id> <filename>`-style invocation because that would require you to know the filename you want to fetch beforehand, and in most cases the person would not know the filename.

How about:
- we only ever print single-file gists,
- we error out on multi-file gists,
- we support something like a `--json` flag to get raw JSON output which you can pipe to your own script that would handle multi-file.

github/client.go
```diff
@@ -1010,41 +1005,44 @@ func (client *Client) FetchGist(id string, filename string) (gist Gist, err erro
...
+	files := map[string]GistFile{}
+	files[basename] = gf
+	g := Gist{
+		Files:  files,
+		Public: public,
+	}
+
+	res, err := api.PostJSON("gists", &g)
+	if err = checkStatus(201, "creating gist", res, err); err != nil {
+		if res != nil && res.StatusCode == 404 && !strings.Contains(res.Header.Get("x-oauth-scopes"), "gist") {
```
jaymzh: ohhh that's neat.

jaymzh:
> Pushed some improvements to your branch. Hope that's ok! Let me know what you think.

Yup! I didn't know that was possible if the PR was based on a fork, neat.

> How about:
> 
>  we only ever print single-file gists,
>  we error out on multi-file gists,
> we support something like a --json flag to get raw JSON output which you can pipe to your own 
> script that would handle multi-file.

I mostly agree - only one small disagreement. Yes, it's almost entirely single-file gists, so orienting around that makes sense. However, if someone sends me a multi-file gist where one file is the script and another is the output, and I want to get the script into a file, I'd *really* like to be able to do that without copy-pasting. So the use case of "I look at the webui and decide OK, I want to do `hub gist <id> script.sh` or whatever. I also don't want to arbitrarily limit other people just because I happen to be a single-file-gist user.

So how about this:

* If you request a multi-file gist without a filename, we error out
* If you request a single-file gist, it just works
* We have a `--json` option, as you suggested.

 I think that addresses your concern with the ugly output on multi-file output while also providing maximum usefulness to everyone involved. Work for you?

github/client.go
```diff
@@ -1010,41 +1005,44 @@ func (client *Client) FetchGist(id string, filename string) (gist Gist, err erro
...
+			err = fmt.Errorf("%s\nGo to https://%s/settings/tokens and enable the 'gist' scope for hub", err, client.Host.Host)
```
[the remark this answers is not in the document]
mislav: Good question! `fmt.Errorf` is weird in it that it doesn't write to stdout nor stderr, but creates an Error object instead with the given message.

mislav:
> I didn't know that was possible if the PR was based on a fork

It's enabled by default, but you can disable it at the time of creating a new PR.

> * If you request a multi-file gist without a filename, we error out

Agreed: this is a good compromise.

I would still like to enable creating multi-file gists like so: `hub gist <file1> <file2...>`. I realize that this clashes with the `hub gist <id> <filename>`-style invocation for fetching existing gists. I wonder how we could un-clash them:
* `hub gist create <file...>` & `hub gist show <id> [<filename>]` - requiring a subcommand is an easy workaround, but now the user is required to type a lot more
* `hub gist --file <file1> --file <file2>` - multiple upload flags could be allowed
* …something else I'm not seeing yet?

jaymzh:
I think `create` and `show` are the cleanest... and people can always make aliases for them. That's easier than making an alias for `--file <file1> --file <file2>`. I'll update the PR this week when I have some time.

jaymzh:
OK @mislav - I *think* that's all of the things we discussed. LMK if I missed anything.

mislav:
@jaymzh Great work; thank you. I've pushed some fixes and tweak, please review! The most notable change is the dropping of `--json` flag; see the commit message for rationale and let me know if that makes sense. 🙇

commands/gist.go
```diff
@@ -13,62 +13,38 @@ var (
...
-    https://gist.github.com/bdf551042f77bb8431b99f13c1105168
+	$ echo hello | hub gist create --public
 
-    # Or with multiple files
-    $ hub gist create /tmp/testfile /tmp/testfile2
-    https://gist.github.com/bdf551042f77bb8431b99f13c1105168
+	$ hub gist create <file1> <file2>
 
+    # print a specific file within a gist:
+    $ hub gist show <ID> testfile1.txt
```
jaymzh: We should definitely show that you don't need a file if there's only one

commands/gist.go
```diff
@@ -131,16 +121,18 @@ func createGist(cmd *Command, args *Args) {
 
 func showGist(cmd *Command, args *Args) {
 	args.NoForward()
-
-	host, err := github.CurrentConfig().DefaultHostNoPrompt()
-	utils.Check(err)
-	gh := github.NewClient(host.Host)
-
+	if args.ParamsSize() < 1 {
```
jaymzh: Remove this error message - there may be only one file. :)

commands/gist.go
```diff
@@ -46,19 +52,21 @@ gist show <ID> [<FILENAME>]
 
 hub(1), hub-api(1)
 `,
-		KnownFlags: "\n",
 	}
 
 	cmdShowGist = &Command{
-		Key:        "show",
-		Run:        showGist,
-		KnownFlags: "\n",
```
jaymzh: can you explain the newline passed to known flags as opposed to not listing it? I got that from `pr.go`...

commands/gist.go
```diff
@@ -72,33 +68,27 @@ func init() {
...
 	}
 
-	if len(gist.Files) > 1 && !emitJson && filename == "" {
-		return fmt.Errorf("There are multiple files, you must specify one, or use --json")
+	if len(gist.Files) > 1 && filename == "" {
+		filenames := []string{}
+		for name := range gist.Files {
+			filenames = append(filenames, name)
+		}
+		return fmt.Errorf("the gist contains multiple files, you must specify one:\n%s", strings.Join(filenames, "\n"))
```
jaymzh: should we sort these for consistent output and also to make the tests more exact?

jaymzh:
Yeah I think dropping the json is fine. I like most of your changes, left some comments - you introduced a bug/ broke a feature...

jaymzh:
Hmmm, it seems like your `create` changes broke something as well;

```
[phil@rider (gist) hub]$ cat /tmp/testfile | ./bin/hub create
Existing repository detected
https://github.com/jaymzh/hub
```

jaymzh:
OK! I pushed a commit that fixes the things I commented on. LMK if you have any objections.

commands/gist.go
```diff
@@ -141,18 +146,17 @@ func createGist(cmd *Command, args *Args) {
 
 func showGist(cmd *Command, args *Args) {
 	args.NoForward()
-	if args.ParamsSize() < 1 {
-		utils.Check(cmd.UsageError("you must specify a gist ID"))
```
mislav: How come you removed this? Now the `hub gist show` command (no arguments) will crash instead of gracefully error out with the above message.

features/gist.feature
```diff
@@ -44,10 +44,10 @@ Feature: hub gist
     Then the exit status should be 1
     Then the stderr should contain:
       """
-      the gist contains multiple files, you must specify one:\n
+      This gist contains multiple files, you must specify one:
+        hub_gist1.txt
+        hub_gist2.txt
```
mislav: Nice! Reads much better

commands/gist.go
```diff
@@ -20,7 +21,7 @@ gist show <ID> [<FILENAME>]
 
 ## Commands:
 
-	* _create_:
+    * _create_:
```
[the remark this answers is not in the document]
jaymzh: Yes, to match the rest of the help - the tabs don't show up properly in various terminals, so it's all non-indented.

commands/gist.go
jaymzh: OHHH because I misread - I thought this was checking for a filename. Sorry I'll revert that.

jaymzh:
Added check back in.

commands/gist.go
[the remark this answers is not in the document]
mislav: @jaymzh The identation change broke help formatting for the lines, so I've changed them back to tabs. The formatting seems OK now: `make man-pages && hub help gist`

We use this markdown contents to generate man pages. The indentation will be transformed as markdown is parsed, but for now it needs to stay tabs.

mislav:
Thanks for all your help!

jaymzh:
Woo! Thanks for all of your help as I stumbled through golang!

PR #2272, opened 2019-09-18
fixes #2270 resolving host name with openssh alias

### LINKED ISSUE #2270: With multiple github accouts configured with openssh hub does not recognize repository
I'm using openssh on windows to authenticate my git calls for two different github accounts. I've setup my openssh config according to https://gist.github.com/jexchan/2351996 . As a result my repository origin is git@github.com-sshUser:gitUser/repository.git instead of usual git@github.com:gitUser/repository.git.

With this change hub refuses to recognize my repository with following message : 
Aborted: could not find any git remote pointing to a GitHub repository.

When I switch my ssh config to have only one github user and switch my repository origin not to include sshUser - hub works fine.

VladimirUAZ:
addressing issue https://github.com/github/hub/issues/2270


VladimirUAZ:
not the right way to address the issue

Issue #12, opened 2009-12-13
Honor git global pagination options/behavior


stephencelis:
(I guess it's usually "paging", not "pagination"...)

Running `git` (`git help`, and `git --help`) without any more arguments displays the usage without a pager. Running `hub`, however, doesn't honor this. Writing `hub --no-pager` just sends execution straight to `git` (as do `--paginate` and `-p`). This commit fixes the behavior:

http://github.com/stephencelis/hub/commit/ed49567cd36a2880de9ba3a65892c45557d52594

Not sure if the execution is desirable (i.e., adding the @no_pager variable), but it's the least invasive without parsing through all the global options.

PR #2280, opened 2019-09-28
Implement PR merge command

friederbluemle:
## Background

Initially this was just a small hack in my local `hub` repo. After having used it without any issues almost every day for the past two years, I thought I would open this PR for discussion to see if others are also interested.

Personally, I find it super convenient not having to open a browser/website to **merge a PR** (which I usually create from command line as well using `hub pull-request`).

## Details

The GitHub API has offered a "merge" functionality for a long time:

```
PUT /repos/:owner/:repo/pulls/:pull_number/merge
```

See https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button

So far, this was not supported by `hub`.

I added a new `hub pr merge <PR-NUMBER>` command that takes the PR number and merges the PR (if the user has write access to the repo).

Note: This is the first `PUT` route - I had to add a `PutJSON` function to `github/http.go`.

## Examples

```sh
$ hub pr list
      #1  Create README.md
```

```sh
# (Success case - PR 1 is merged, no output, exit code `0`)
$ hub pr merge 1
```

```sh
# (Error case, for example user has no write access)
$ hub pr merge 1
Error merging pull request: Not Found (HTTP 404)
Not Found
```

I also slightly updated the `hub pr` documentation to better reflect what it can do now.

```sh
$ hub
...
These GitHub commands are provided by hub:

...
   pr             Manage GitHub Pull Requests
...
```

```sh
$ hub merge
Usage: hub pr list [-s <STATE>] [-h <HEAD>] [-b <BASE>] [-o <SORT_KEY> [-^]] [-f <FORMAT>] [-L <LIMIT>]
       hub pr checkout <PR-NUMBER> [<BRANCH>]
       hub pr show [-uc] [-h <HEAD>]
       hub pr show [-uc] <PR-NUMBER>
       hub pr merge <PR-NUMBER>
```

## Future ideas

Some ideas for the future (either as part of this PR or a follow up PR):

- Improve error messages
- Support different merge methods
- Before the merge, see if all checks are configured and have passed - if not, abort merge
- Allow "force" merging (if checks have not passed, see previous point)
- Allow customizing the merge commit message
- Add an option to automatically delete the PR head branch after successful merge

### DIFF
--- a/commands/help.go
+++ b/commands/help.go
@@ -223,7 +223,7 @@ These GitHub commands are provided by hub:
    fork           Make a fork of a remote repository on GitHub and add as remote
    gist           Make a gist
    issue          List or create GitHub issues
-   pr             List or checkout GitHub pull requests
+   pr             Manage GitHub pull requests
    pull-request   Open a pull request on GitHub
    release        List or create GitHub releases
    sync           Fetch git objects from upstream and update branches
--- a/commands/pr.go
+++ b/commands/pr.go
@@ -19,6 +19,7 @@ pr list [-s <STATE>] [-h <HEAD>] [-b <BASE>] [-o <SORT_KEY> [-^]] [-f <FORMAT>] 
 pr checkout <PR-NUMBER> [<BRANCH>]
 pr show [-uc] [-f <FORMAT>] [-h <HEAD>]
 pr show [-uc] [-f <FORMAT>] <PR-NUMBER>
+pr merge <PR-NUMBER>
 `,
 		Long: `Manage GitHub Pull Requests for the current repository.
 
@@ -35,6 +36,9 @@ pr show [-uc] [-f <FORMAT>] <PR-NUMBER>
 		specified, <HEAD> is used to look up open pull requests and defaults to
 		the current branch name. With ''--format'', print information about the
 		pull request instead of opening it.
+
+	* _merge_:
+		Merge a pull request in the current project.
 
 ## Options:
 
@@ -173,12 +177,18 @@ hub-issue(1), hub-pull-request(1), hub(1)
 		--color
 `,
 	}
+
+	cmdMergePr = &Command{
+		Key: "merge",
+		Run: mergePr,
+	}
 )
 
 func init() {
 	cmdPr.Use(cmdListPulls)
 	cmdPr.Use(cmdCheckoutPr)
 	cmdPr.Use(cmdShowPr)
+	cmdPr.Use(cmdMergePr)
 	CmdRunner.Use(cmdPr)
 }
 
@@ -409,6 +419,37 @@ func deducePushTarget(branch *github.Branch, owner string) (*github.Project, err
 		return nil, fmt.Errorf("no remote found for branch %s", branch.ShortName())
 	}
 	return remote.Project()
+}
+
+func mergePr(command *Command, args *Args) {
+	localRepo, err := github.LocalRepo()
+	utils.Check(err)
+
+	project, err := localRepo.MainProject()
+	utils.Check(err)
+
+	gh := github.NewClient(project.Host)
+
+	args.NoForward()
+	if args.Noop {
+		ui.Printf("Would merge pull request for %s\n", project)
+		return
+	}
+
+	words := args.Words()
+
+	if len(words) == 0 {
+		utils.Check(fmt.Errorf("Error: No pull request number given"))
+	}
+
+	prNumberString := words[0]
+	prNumber, err := strconv.Atoi(prNumberString)
+	utils.Check(err)
+
+	err = gh.MergePullRequest(project, prNumber, map[string]interface{}{
+		"merge_method": "merge",
+	})
+	utils.Check(err)
 }
 
 func formatPullRequest(pr github.PullRequest, format string, colorize bool) string {
--- a/github/client.go
+++ b/github/client.go
@@ -145,6 +145,21 @@ func (client *Client) CreatePullRequest(project *Project, params map[string]inte
 	return
 }
 
+func (client *Client) MergePullRequest(project *Project, prNumber int, params map[string]interface{}) (err error) {
+	api, err := client.simpleApi()
+	if err != nil {
+		return
+	}
+
+	res, err := api.PutJSON(fmt.Sprintf("repos/%s/%s/pulls/%d/merge", project.Owner, project.Name, prNumber), params)
+	if err = checkStatus(200, "merging pull request", res, err); err != nil {
+		return
+	}
+
+	res.Body.Close()
+	return
+}
+
 func (client *Client) RequestReview(project *Project, prNumber int, params map[string]interface{}) (err error) {
 	api, err := client.simpleApi()
 	if err != nil {
--- a/github/http.go
+++ b/github/http.go
@@ -445,6 +445,10 @@ func (c *simpleClient) PostJSONPreview(path string, payload interface{}, mimeTyp
 	})
 }
 
+func (c *simpleClient) PutJSON(path string, payload interface{}) (*simpleResponse, error) {
+	return c.jsonRequest("PUT", path, payload, nil)
+}
+
 func (c *simpleClient) PatchJSON(path string, payload interface{}) (*simpleResponse, error) {
 	return c.jsonRequest("PATCH", path, payload, nil)
 }


jaymzh:
Count me as a :+1: 

I'm not too worried about the confusion, `pr merge` is what happens when you click `merge` on a `pr`. But `merge` is just... a local merge. I think it maps well, though the docs should clarify how each works.

iainsmith:
This looks great. 

It would be also worth thinking about rebase and squash actions. 

Hypothetical those could be ‘hub pr rebase/squash N’ or an alternative ‘hub pr accept —merge/squash/rebase’

sbeyer:
I also see no confusing between `hub merge` and `hub pr merge`. I find it much harder  to comprehend that there is `hub pull-request` instead of `hub pr create`...

However, what I am missing in my current workflow, is a quick way to
 1. merge the PR
 2. delete the feature branch on GitHub
 3. delete the feature branch locally (i.e., run `git fetch --prune` and `git branch -d`)

(1) and (2) can currently be done using the browser; (1) can be done with your `pr merge` feature, and (2) is planned for the future. So this is a feature request for your feature PR: please also think of (3) when implementing (2). :-)

jaymzh:
@sbeyer - do you find that when PRs are merged, `git branch -d` works for you?  I find git never knows it's there (perhaps because we do squash merges), and thus `git branch -D` is required... been trying to find a sane way to sanity-check that in a wrapper script but have failed.

sbeyer:
> do you find that when PRs are merged, `git branch -d` works for you?

It does, but I don't use squash merges. Git does not track branch history, so it does not know when a branch is squashed. I btw never got to know why someone wants to use squash merges in general (I can think of a few use cases though... well, that's all off-topic, sorry). However, the example of squash merges (which should probably also be supported by `pr merge`) is a good example where a "delete remote and local branch" option makes even more sense (and a "delete only remote branch" also makes sense because the original commits are kept).

jaymzh:
> I btw never got to know why someone wants to use squash merges in general

I know this is way off topic, but since you asked, and it's a valid question...

On PRs with a ton of back and forth, I don't want the history of "fixed typo" "changed casing" "address code review comments" "address code review comments" "code review" "more fixes"

It's messy, makes it hard to look through history, etc. So you have two options, you can either ask people to clean up their history before you merge and force-push and then merge (which usually a ff-merge), or you can squash-merge, which is really convenient. Assuming you follow best-practices of breaking up large changes into small bite-size PRs, one detailed commit message is sufficient (and, IMO, better).

This is the basic model of Phabricator (from Facebook), btw, except that you always amend, and phabricator keeps track of the delta between each time you update the PR (they call them 'diffs'), so the final history in the git repo is morally the same as a squash merge, but the various revisions of that "diff" are all also discoverable (this would be the same as if you GitHub kept the history of all force-pushes and you could go back through them all).

There's pros-and-cons each way, no way is "the right way", but that's the pros of that model and why people (at least most people I know), use it (especially if they came from google or facebook, both of which use that model).

casz:
The question how will this work with repos that require linear history or have `merge commit` option disabled in repo settings?

From experience with `@dependabot merge` the bot knows this and uses `rebase` option instead.

jaymzh:
I'm not sure how the API works, but if the method of merging isn't specified in the call, will it just use whatever the default is for the repo? That seems better to me.

mislav:
> A question I have, how will this work with repos that require linear history or have `merge commit` option disabled in repo settings?

The API documentation states:
> Merge method to use. Possible values are `merge`, `squash` or `rebase`. Default is `merge`.

It sounds like the API will always default to `merge` if the method is not specified, even if merges are currently prohibited per branch protection rules.

Perhaps this feature should use the API to read branch protection rules and decide on a correct default merge method if the user didn't specify one explicitly. 🤔

sbeyer:
> Perhaps this feature should use the API to read branch protection rules and decide on a correct default merge method if the user didn't specify one explicitly. :thinking:

I (as a potential user) disagree. If you are switching between projects with different defaults, this can lead to unexpected results. Hence I think this feature should always default to `merge`. If this leads to an error, the user is reminded that she should use another merge method, and can then do so *explicitly*.

jaymzh:
> I (as a potential user) disagree. If you are switching between projects with different settings (edit: i.e., branch protection rules), this can lead to unexpected results. Hence I think this feature should always default to merge. If this leads to an error, the user is reminded that she should use another merge method, and can then do so explicitly. (edit: imho the API is perfectly sane regarding this feature.)

Not sure I agree. As it stands when you click the green `merge` button on a repo, you get a different behavior per-repo, right? Why should the CLI by any different? The least-surprising thing, IMO, would be to mirror the website. Just like on the website you can click the drop down and choose a different behavior, but the default depends on the configuration of the repo.

And in fact, if you have commit access to several different repos with several different policies, you *want* the default to correctly handle that.

sbeyer:
> The least-surprising thing, IMO, would be to mirror the website.

I think our disagreement boils down to the question what one considers the least surprising behavior. I am not quite sure if I recall correctly, but I think the website **tells** you **before** clicking the merge button, if it is going to merge, squash or rebase. In an CLI interface (or a REST API) on the other hand, you do not see it and are thus not warned beforehand.

However, I think we agree that the feature needs `--merge`, `--squash`, and `--rebase` options to mimic the options on the website. From a perspective of "user-friendliness", I would agree that it is better to not have an extra option like `--default-behavior`, and that `hub pr merge` without options should simply perform the default behavior.

Since I am torn between the "least surprise" and the "fewer options" viewpoint, and since what one considers "least surprise" seems to be debatable, your solution is probably the best one.

casz:
Perhaps just show an interactive choice with the options available?

jaymzh:
> However, I think we agree that the feature needs --merge, --squash, and --rebase options

@sbeyer - Yes, I absolutely 100% agree. For the cases - or people - who want to explicitly have one kind of merge, those need to be there.

> that hub pr merge without options should simply perform the default behavior.

OK, I think we're both in agreement on all things then. :)

@casz:

> Of course it depends on repo settings whether protected branch requires linear history or based on the merge button options that are enabled inside repo settings.

Yes, that's what we're saying - the default behavior of the webUI depends on the branch protection settings and which merge buttons are enabled, and I'm proposing the CLI should do the same.

casz:
My proposal was instead of defaulting it should present the choice interactively.

casz:
My case in point what is the default of merge is disabled?

When pr have one commit is it rebase or squash?
Multiple commits it is then squash or rebase?

My point is the CLI should not default.

Also when I choose squash it should open the my git code editor and give me the lists of commits and allow me to write the title and description.

jaymzh:
@casz 

> My case in point what is the default if merge is disabled?

I think you mean "default" differently than I do here. On a PR, on the web, if you have "mergecommit" disabled, then the button does "squash", which is what the default for your repo is. If that's the thing that happens when you click the button on the PR, then that's what should happen when you use the CLI. By default, I mean "what happens if you were click the button on the PR" - which could be merge-commit, or rebase, or squash, depending on what you've configured for the repo.

And that's precisely the point - whatever the admins of the repo have configured to happen for PR merging, the CLI should honor that.

casz:
I strongly disagree with assuming squash is the "default" when merge commit is disabled.

I normally leave both rebase and squash enabled in most of the repositories I am a maintainer on.
it depends on the context of the PR and I think it would be nice if the CLI presents the option just as it does on the web. I can click the little fold out option.

I don't want to fight my git code editor when CLI defaults to squash and I have to cancel it because I wanted rebase.

jaymzh:
Wait, hold on, I think we're talking past each other. I'm not being clear, or you're misunderstanding.

Today, it is possible, to setup your repo, so that when you click "merge", if you **do not click the flyout** that $ACTION happens. That can be a merge-commit, or a rebase, or a squash. You can make any of them happen depending on your choice of settings. Whatever happens, on your particular repo, when you click merge **and not the flyout**, is what I'm calling the default. On some of my repos, that will be a rebase, on some of my repos that will be a merge, depending on what the maintainers, as a group, has decided we prefer in the common case.

If, on your repo, when I click "merge" (and **not** the flyout**), I get "rebase", then I'm calling the "default" for your repo rebase. If that's merge-commit, then the default for your repo is merge commit. If that's squash, then the default for your repo is squash.

So, what I'm saying is `hub pr merge` is like clicking merge **without** using the flyout, which means it should just do whatever the "default" for your repo is (where "default" is defined as above) just like it would on the web because you haven't overridden it, but, you can have the equivalent of  "click the flyout" in the cli by doing `hub pr merge --rebase` or `hub pr merge --merge` or `hub pr merge --squash`, at anytime (though, if your branch protections don't allow it, of course that would fail).

I'm not trying to imply that squash or rebase or merge should be a default. I'm simply saying you can already **set** a behavior that is default on a repo if the user doesn't choose to click the little arrow and choose something else... so there already *is* a default on every repo, based on the settings.

* `pr merge` -> is clicking the merge button and not changing the setting you get
* `pr merge --merge` -> is clicking the arrow, choosing `merge commit` and clicking merge
* `pr merge --rebase` -> is clicking the arrow, choosing `rebase` and clicking merge
* `pr merge --squash` -> is clicking the arrow, choosing `squash` and clicking merge

Does that make sense?

casz:
The "default" is what ever you used last on that repo as far I recall.

If you used rebase on one PR it will default to rebase on the next one.

casz:
`pr merge` -> present options if 2 or more are available and demand that this is called from an interactive terminal, the other options below should be used for scripting.
Presenting the options would be an interactive selection and I can use the arrow keys to select and press enter once again to confirm my selection.

`pr merge --merge` -> is clicking the arrow, choosing merge commit and clicking merge
`pr merge --rebase` -> is clicking the arrow, choosing rebase and clicking merge
`pr merge --squash` -> is clicking the arrow, choosing squash and clicking merge

Would be my suggestion

jaymzh:
> The "default" is what ever you used last on that repo as far I recall.

Wait, really? This has not been my experience... now I want to go make a test repo and test.

If that's true, then yes, I agree with you about what `per merge` should do.

jaymzh:
@jetersen - You're right it does seem to be "sticky". So I agree it should force a clear choice with -- options to bypass prompts.

@friederbluemle - that work for yoU?

jetersen:
if anything it should follow the design of https://github.com/cli/cli/pull/873

mislav:
Thank you @friederbluemle for the initial implementation and everyone else who contributed to the discussion!

The feature is now merged:
```sh
gh pr merge 12  # merge PR #12
gh pr merge --squash 12
gh pr merge --rebase 12
gh pr merge --delete-branch 12  # deletes head branch when done
gh pr merge --head-sha=<COMMIT-SHA> 12  # verifies head branch state
gh pr merge 12 -m "override merge commit title"
```

I have decided not to pick a merge method automatically. If people feel strongly about the merge method being `--squash` or `--rebase` on repositories that enforce linear history, you may open an issue or submit a pull request, but for now I expect the caller of `pr merge` to explicitly pass in a merge method as desired.

friederbluemle:
Thank you @mislav :+1:

jaymzh:
Thanks @mislav ! One question, you mention:
> gh pr merge 12  # merge PR #12
but also say:
> I have decided not to pick a merge method automatically.
So what happens when you don't specify one which you show is possible with that command.

friederbluemle:
I would assume it automatically picks the default (i.e. a regular merge commit, as [documented here](https://developer.github.com/v3/pulls/#merge-a-pull-request)):

image

Based on some testing I did a couple of months ago, you'd get an error message if the default method was disabled for that repo via the settings page on GitHub. You'd then need to explicitly declare either `--squash` or `--rebase`.

mislav:
@friederbluemle That's right; thank you! /cc @jaymzh 

I realize that the default behavior is not ideal for repositories where merges are disallowed, but maintainers in those repositories already know their preferred merge method and it shouldn't be hard for them to add the flag.

PR #2289, opened 2019-10-03
Update Datum in LICENSE

crydotsnake:


mislav:
Thank you, but we're fine with leaving the copyright date being the date of the original publication.

We might revisit the line that defines the copyright holder in the future, and then we can easily take care of the year in the same step. But right now it's not necessary.

crydotsnake:
Ok.

PR #2291, opened 2019-10-03
Add completions for git fork

### LINKED ISSUE #2108: fork --remote-name missing in bash completions
Short on time, otherwise I would just PR this. I'll leave it here for now and circle back to it later if nobody else gets to it.

Simply put, all that needs to happen is the flag `--remote-name` needs to be added to `_git_fork`.

https://github.com/github/hub/blob/accfda9aa4009597da4362be31cb179b8392b84d/etc/hub.bash_completion.sh#L222-L236

Thanks in advance!

sinisterstuf:
Based on the manual page for git fork, this pull request adds `--remote-name` and `--org` completions, waiting for user-provided free-text as arguments to both. It also considers the `--remote-name` and `--no-remote` flags to be mutually exclusive. See commit messages for details.

Fixes #2108.

### DIFF
--- a/etc/hub.bash_completion.sh
+++ b/etc/hub.bash_completion.sh
@@ -224,12 +224,18 @@ EOF
     while [ $c -lt $cword ]; do
       i="${words[c]}"
       case "$i" in
-        --remote-name|--org)
+        --org)
           ((c++))
           flags=${flags/$i/}
           ;;
+        --remote-name)
+          ((c++))
+          flags=${flags/$i/}
+          flags=${flags/--no-remote/}
+          ;;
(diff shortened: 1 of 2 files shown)


sinisterstuf:
Shell completion tests don't seem to run in the Travis builds but, but I updated the bash completion feature file to match the code changes in this pull request.

mislav:
Sweet! Thank you 🎉 

> Shell completion tests don't seem to run in the Travis builds

Yeah, unfortunately I had to disable them because Travis made a change to their infrastructure that somehow disabled running an interactive shell inside tmux during builds. 😞

Issue #2288, opened 2019-10-03
Support `pr show --format`


jasonkuhrt:
I would like the ability to easily get the pr number of the current branch. We already have this underlying information since it makes `$ git show pr` possible but rather than opening a link to the PR I just want the PR number sent to `stdout`.

My motivation is to leverage this information in a script, along the lines of:

```sh
yarn publish \
  --tag "pr$(git pr show --number)" \
  --no-git-tag-version \
  --new-version "0.0.0-pr.$(git pr show --number).$(git show -s head --format='%h')"
```

mislav:
Hi thank you for writing in!

You can extract the PR number in bash like so:

```sh
# split fields by "/" and output the last value
hub pr show -u | awk -F/ '{print $(NF)}'
```

For a more sophisticated approach, you can use `hub pr list`:

```sh
# 1. Fetch open PRs where the current branch is the head,
# 2. Output only the PR number.
hub pr list --head="$(git branch --show-current)" --format='%I%n'
```

jasonkuhrt:
Thanks for the workaround @mislav! `-u` gets us really close. I think a `-n` that provides the last step would still be worthwhile, though.

I'm personally not familiar with awk, prefer regexp.

```
git pr show -u | pcregrep -o '\d+$'
```

But yeah the following would be ideal I guess:

```
git pr show -n
```

mislav:
Thank you for the feedback.

Personally, I think supporting `-n`/`--number` would be adding an _additional_ mechanism for selecting specific fields for output that is different than the current mechanism, which is the `--format` flag.

Do you feel that `hub pr show` should then support `--format='%I'?

jasonkuhrt:
> Do you feel that hub pr show should then support --format='%I'?

That would generally align with `git` conventions. I think `--format` would be justifiable if there were at least a few symbols available? What might the use-cases be?

```
%n   PR number
%cn  PR comments number
%a    PR author github handle
...
```

Seems we could come up with quite a few :)

mislav:
> I think `--format` would be justifiable if there were at least a few symbols available?

We already support quite a few in `hub pr list`: https://hub.github.com/hub-pr.1.html#options 😉 

My suggestion was that we could consider also making them available to `hub pr show` with the same syntax so we don't have to reinwent the wheel and add flags such as `-n/--number`.

jasonkuhrt:
@mislav sounds great, I agree

mislav:
[...]

PR #2297, opened 2019-10-08
implement format for `pr show` for #2288

tomlazar:
Adds the same formatting for `pr show` that are used in `pr list`. The format option supercedes the copy `-c, --copy` or url `-u, --url` flags and instead prints out the information from the git api. I choose not to remove `--url` for now to not break any existing scripts that use that flag.


mislav:
Thank you! I've pushed some changes. Let me know how it looks!

tomlazar:
Looks good to me, I tried something around that route but got rid of it while I was testing. turned out it was a problem with how I wrote the test and never put it back.

PR #2299, opened 2019-10-09
[docs] Add link to apt package manager

kerolloz:
I just added a link to the apt package manager from Debian man pages.


kerolloz:
@mislav

PR #2314, opened 2019-10-19
read extraheader from .git

MarcelHillmann:
As developer I like to use our Github Enterprise, which is protected by an keycloak.

We use a tool to set an cookie to the git cli an the hub has use the same cookie too.


MarcelHillmann:
Sorry I have no clue, whay the cucumber test case is broken

github/client.go
```diff
@@ -37,6 +44,14 @@ type Client struct {
 	cachedClient *simpleClient
 }
 
+func SetUserAgent() {
+	if userAgent := os.Getenv("HUB_USERAGENT"); userAgent != "" {
+		UserAgent = userAgent
+	} else if userAgent, err := git.Config("hub.useragent"); err == nil {
+		UserAgent = userAgent
```
mislav: Does your Enterprise instance also require setting a custom user agent string?

github/client.go
```diff
@@ -970,6 +985,29 @@ func (client *Client) apiClient() *simpleClient {
 	return &simpleClient{
 		httpClient: httpClient,
 		rootUrl:    apiRoot,
+		extraHeader: func(r *http.Header) {
+			if headers, err := git.ConfigAll(fmt.Sprintf("http.%s://%s.extraheader", client.Host.Protocol, client.Host.Host)); err == nil {
```
mislav: I'm wondering whether it would be better to design this feature so that `extraHeader` is configured from the "outside" (being passed the list of extra headers), rather than making git requests here. I want to avoid the HTTP client being unnecessarily coupled to git config.

github/client.go
```diff
@@ -970,6 +985,29 @@ func (client *Client) apiClient() *simpleClient {
...
+				url, _ := url.Parse(fmt.Sprintf("%s://%s", client.Host.Protocol, client.Host.Host))
+				cookies := []*http.Cookie{}
+
+				for _, header := range headers {
+					keyValue := strings.Split(header, ":")
+					key, value := strings.TrimSpace(keyValue[0]), strings.TrimSpace(keyValue[1])
+
+					if strings.ToUpper(key) == "COOKIE" && strings.Contains(value, "=") {
```
mislav: Tip: you can use `strings.EqualFold()` to compare strings case-insensitive

github/client.go
```diff
@@ -970,6 +985,29 @@ func (client *Client) apiClient() *simpleClient {
...
+						cookieKV := strings.Split(value, "=")
+						cookieKey, cookieValue := strings.TrimSpace(cookieKV[0]), strings.TrimSpace(cookieKV[1])
+						cookies = append(cookies, &http.Cookie{Name: cookieKey, Value: cookieValue})
+						// r.Add(key,value)
+					}
+				}
+
+				if len(cookies) > 0 && httpClient.Jar == nil {
+					httpClient.Jar, _ = cookiejar.New(&cookiejar.Options{PublicSuffixList: publicsuffix.List})
+					httpClient.Jar.SetCookies(url, cookies)
```
mislav: Is keeping a cookie jar really necessary for API requests to your Enterprise instance?

github/http.go
```diff
@@ -38,6 +38,8 @@ var inspectHeaders = []string{
 	"Location",
 	"Link",
 	"Accept",
+	"Cookie",
+	"User-Agent",
```
mislav: You have changed the format of the HUB_VERBOSE output (you've added additional headers), so this test broke: https://github.com/github/hub/blob/87a31b093bcc30a726b5f09bb4f4b86343659d5d/features/create.feature#L271-L288

mislav (changes requested):
Hi, this is an interesting feature. I'd like to enable you using hub with your Enterprise instance, but I wouldn't add this in this format yet, not until we clean up the code a bit and make some decisions about how `hub.useragent` is configured (I'd like to avoid keeping this in git config). If you're willing, let's iterate on this for now—I've added some thoughts below

github/client.go
MarcelHillmann: This makes sence, I thought that is the default pattern.

github/client.go
MarcelHillmann: We have utility, to handel the process with KeyCloack as IDP.
This tool is writing the git/extraHeader for the git cli workflow.

https://github.com/git/git/blob/042ed3e048af08014487d19196984347e3be7d1c/Documentation/config/http.txt#L52

I could refactor this function, that it is call once and only the "r" manipulation is running

github/client.go
MarcelHillmann: Thanks, never saw it

github/client.go
MarcelHillmann: The dataflow is:
1. Nginx
1. 302: redirect to KeyCloack
   KeyCloack is forwarding to internal server with SessionID and traceIds
1. Client is requesting again the GHE behind Nginx 
1. Nginx is passing to GHE

So there are 2 or 3 round trips with a bunch of cookies

github/http.go
MarcelHillmann: Thanks, that was for development

PR #2317, opened 2019-10-20
Honor the rate limit headers in "hub api"

mdaniel:
Previously, calling `hub api --paginate` with any substantial list of results would exhaust the Ratelimit allocation, since `hub api` did not pause before requesting the next page. With this change, `hub api` will check the rate limit headers and sleep until "Reset" if calling for the next page would exceed the limit.

### DIFF
--- a/commands/api.go
+++ b/commands/api.go
@@ -9,6 +9,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/github/hub/github"
 	"github.com/github/hub/ui"
@@ -86,6 +87,15 @@ var cmdApi = &Command{
 		requests as well. Just make sure to not use '--cache' for any GraphQL
 		mutations.
 
+	--rate-limit
+		If the next '--paginate' request would exceed the rate limit allocation
+		as specified by the server, then print a message to standard error and
+		sleep until the time given in the rate limit reset header.
+
+		Without this flag, '--paginate' will continue to issue requests without
+		regard to the number of remaining rate limit slots, potentially
+		resulting in only partial pagination results.
+
 	<ENDPOINT>
 		The GitHub API endpoint to send the HTTP request to (default: "/").
 		
@@ -136,7 +146,7 @@ func init() {
 	CmdRunner.Use(cmdApi)
 }
 
-func apiCommand(cmd *Command, args *Args) {
+func apiCommand(_ *Command, args *Args) {
 	path := ""
 	if !args.IsParamsEmpty() {
 		path = args.GetParam(0)
@@ -228,6 +238,8 @@ func apiCommand(cmd *Command, args *Args) {
 		body = params
 	}
 
+	rateLimit := args.Flag.Bool("--rate-limit")
+
 	gh := github.NewClient(host)
 
 	out := ui.Stdout
@@ -292,13 +304,38 @@ func apiCommand(cmd *Command, args *Args) {
 		if requestLoop && !parseJSON {
 			fmt.Fprintf(out, "\n")
 		}
+
+		if requestLoop && rateLimit {
+			var rateLimitLeft = -1
+			var rateLimitResetMs = -1
+			var atoiErr error
+			if xRateLimitLeft := response.Header.Get(rateLimitRemainingHeader); len(xRateLimitLeft) != 0 {
+				if rateLimitLeft, atoiErr = strconv.Atoi(xRateLimitLeft); atoiErr != nil {
+					ui.Errorf("Unable to parse header \"%s\": \"%s\" as an int", rateLimitRemainingHeader, xRateLimitLeft)
+					rateLimitLeft = -1
+				}
+			}
+			if xRateLimitReset := response.Header.Get(rateLimitResetHeader); len(xRateLimitReset) != 0 {
+				if rateLimitResetMs, atoiErr = strconv.Atoi(xRateLimitReset); atoiErr != nil {
+					ui.Errorf("Unable to parse header \"%s\": \"%s\" as an int", rateLimitResetHeader, xRateLimitReset)
+					rateLimitResetMs = -1
+				}
+			}
+			if rateLimitResetMs != -1 && rateLimitLeft == 0 {
+				rollover := time.Unix(int64(rateLimitResetMs)+1, 0)
+				ui.Errorf("Pausing until %v for Rate Limit Reset ...\n", rollover)
+				time.Sleep(time.Until(rollover))
+			}
+		}
 	}
 }
 
 const (
-	trueVal  = "true"
-	falseVal = "false"
-	nilVal   = "null"
+	trueVal                  = "true"
+	falseVal                 = "false"
+	nilVal                   = "null"
+	rateLimitRemainingHeader = "X-Ratelimit-Remaining"
+	rateLimitResetHeader     = "X-Ratelimit-Reset"
 )
 
 func magicValue(value string) interface{} {
--- a/commands/api.go
+++ b/commands/api.go
@@ -321,7 +321,7 @@ func pauseUntil(timestamp int) {
 	rollover := time.Unix(int64(timestamp)+1, 0)
 	duration := time.Until(rollover)
 	if duration > 0 {
-		ui.Errorf("API rate limit reached; pausing until %v ...\n", rollover)
+		ui.Errorf("API rate limit exceeded; pausing until %v ...\n", rollover)
 		time.Sleep(duration)
 	}
 }
--- a/commands/api.go
+++ b/commands/api.go
@@ -238,7 +238,7 @@ func apiCommand(_ *Command, args *Args) {
 		body = params
 	}
 
-	rateLimit := args.Flag.Bool("--rate-limit")
+	rateLimitWait := args.Flag.Bool("--rate-limit")
 
 	gh := github.NewClient(host)
 
@@ -253,8 +253,13 @@ func apiCommand(_ *Command, args *Args) {
 	for {
 		response, err := gh.GenericAPIRequest(method, path, body, headers, cacheTTL)
 		utils.Check(err)
-		success := response.StatusCode < 300
 
+		if rateLimitWait && response.StatusCode == 403 && response.RateLimitRemaining() == 0 {
+			pauseUntil(response.RateLimitReset())
+			continue
+		}
+
+		success := response.StatusCode < 300
 		jsonType := true
 		if !success {
 			jsonType, _ = regexp.MatchString(`[/+]json(?:;|$)`, response.Header.Get("Content-Type"))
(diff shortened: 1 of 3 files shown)


commands/api.go
```diff
@@ -250,9 +251,9 @@ func apiCommand(cmd *Command, args *Args) {
 		}
 
 		if includeHeaders {
-			fmt.Fprintf(out, "%s %s\r\n", response.Proto, response.Status)
-			response.Header.Write(out)
-			fmt.Fprintf(out, "\r\n")
+			_, _ = fmt.Fprintf(out, "%s %s\r\n", response.Proto, response.Status)
+			utils.Check(response.Header.Write(out))
+			_, _ = fmt.Fprintf(out, "\r\n")
```
mislav: This PR adds some error handling that is unrelated to the change proposed in this PR:

1. The additions make your PR harder to review;
2. Hub doesn't generally perform error handling around writing to stdout/stderr. Even though error handling is a good rule of thumb in general, I think that the lack of error handling here is fine for now, because it's not presenting problems for anyone.

commands/api.go
```diff
@@ -222,7 +223,7 @@ func apiCommand(cmd *Command, args *Args) {
 			fi, err := os.Open(fn)
 			utils.Check(err)
 			body = fi
-			defer fi.Close()
+			defer utils.Check(fi.Close())
```
mislav: If we couldn't close a file for whatever reason, I think that it would be a much better idea to allow the program to continue (it's not a big deal, after all) rather than to halt it.

commands/api.go
```diff
@@ -290,7 +293,26 @@ func apiCommand(cmd *Command, args *Args) {
 			}
 		}
 		if requestLoop && !parseJSON {
-			fmt.Fprintf(out, "\n")
+			_, _ = fmt.Fprintf(out, "\n")
```
mislav: What does assigning to `_, _` accomplish?

commands/api.go
```diff
@@ -290,7 +293,26 @@ func apiCommand(cmd *Command, args *Args) {
...
+				utils.Check(atoiErr)
+			}
+			if xRateLimitReset := response.Header.Get("X-Ratelimit-Reset"); len(xRateLimitReset) != 0 {
+				rateLimitResetMs, atoiErr = strconv.Atoi(xRateLimitReset)
+				utils.Check(atoiErr)
+			}
+			if rateLimitLeft <= 1 {
+				rollover := time.Unix(int64(rateLimitResetMs)+1, 0)
+				_, _ = fmt.Fprintf(ui.Stderr, "Pausing until %v for Rate Limit Reset ...\n", rollover)
+				time.Sleep(time.Until(rollover))
```
mislav: Nice, elegant implementation of this feature! 👍

commands/api.go
```diff
@@ -290,7 +293,26 @@ func apiCommand(cmd *Command, args *Args) {
...
+			var atoiErr error
+			if xRateLimitLeft := response.Header.Get("X-Ratelimit-Remaining"); len(xRateLimitLeft) != 0 {
+				rateLimitLeft, atoiErr = strconv.Atoi(xRateLimitLeft)
+				utils.Check(atoiErr)
+			}
+			if xRateLimitReset := response.Header.Get("X-Ratelimit-Reset"); len(xRateLimitReset) != 0 {
+				rateLimitResetMs, atoiErr = strconv.Atoi(xRateLimitReset)
+				utils.Check(atoiErr)
+			}
+			if rateLimitLeft <= 1 {
```
mislav: This variable is getting initialized with "0" as default, and here the condition is `<= 1`, which will always match if the server doesn't send the `X-Ratelimit-Remaining` header. This might be what's breaking the tests in CI.

commands/api.go
mdaniel: I did wonder if it would be received as cleaning up things, or noise, which is why they were in separate commits so they could be reverted or removed

I'm hearing they fall more into the "noise" category so I'll remove them

mdaniel:
> Since the rate limit reset window is 1 hour, this functionality could theoretically make the hub api execution last for up to an hour

That wasn't my experience, but I was also using `hub api` with `$GITHUB_TOKEN` set, which I (perhaps erroneously) thought was the most common scenario. However, I tried `hub api` without providing credentials and it wasn't immediately obvious how to get that to work (there's even `client.ensureAccessToken` which implies it must be present)

Anyway, with my token set, the `Reset` window is measured in seconds for me
```
> X-Ratelimit-Limit: 30
> X-Ratelimit-Remaining: 29
> X-Ratelimit-Reset: 1571720200
$ gdate --date='@1571720200'
Mon Oct 21 21:56:40 PDT 2019
$ date
Mon Oct 21 21:56:53 PDT 2019
```

> if the rate limit was exhausted immediately at the beginning of the hour. I'm not sure if it's good for this to be on by default, because I am sure that most users, particularly when running the hub api script manually in their own terminal, would perceive this delay as hub getting "frozen" and terminating the process.

FWIW, the change does include a message to stderr about why it is sleeping and for how long, but I hear you. I stopped short off adding a flag for `--honor-rate-limit` or `--sleep-for-rate-limit` ... because _whew_ naming things, but also because I wasn't sure if it should be opt-in to have `--paginate` honor the rate limit, since maybe people currently just accept that pagination will fail without warning, or have it be opt-out to specifically say "I don't care about the limit, give me as much as you can and die as before". I welcome your input on flag-or-no-flag and for sure welcome what you think would be a good name for it

commands/api.go
mdaniel: I actually did wonder if there was a circumstance under which the headers didn't come back, but you're right, it's a much more sane thing to just initialize the variable. I'll get that straightened out, too

commands/api.go
mdaniel: It was part of the overall "explicit error handling" that made GoLand happy. That's where the rest of those error handling things came from, too: GoLand considers it ... suspicious ... if a call returns `error` but no one checks it

commands/api.go
```diff
@@ -292,6 +293,25 @@ func apiCommand(cmd *Command, args *Args) {
...
+			var atoiErr error
+			if xRateLimitLeft := response.Header.Get("X-Ratelimit-Remaining"); len(xRateLimitLeft) != 0 {
+				rateLimitLeft, atoiErr = strconv.Atoi(xRateLimitLeft)
+				utils.Check(atoiErr)
+			}
+			if xRateLimitReset := response.Header.Get("X-Ratelimit-Reset"); len(xRateLimitReset) != 0 {
+				rateLimitResetMs, atoiErr = strconv.Atoi(xRateLimitReset)
+				utils.Check(atoiErr)
+			}
+			if rateLimitLeft != -1 && rateLimitResetMs != -1 && rateLimitLeft <= 1 {
```
mislav: If `rateLimitLeft == 1`, doesn't that mean that there is 1 more request available during this rate limit window, and that we could take advantage of that and let the request fire rather than waiting until reset? Or am I misunderstanding this condition

mislav:
I'm also wondering whether the appropriate place to implement the wait-until-ratelimit-reset behavior would be in a RoundTripper https://github.com/github/hub/blob/cd81a7bada3d2986f213db424acaa570523a9c9f/github/http.go#L51

This behavior could be off-by-default, but activated on-demand via command-line flag or environment variable. Just thinking out loud

mdaniel:
Just to follow up so you don't think I have forgotten about this, I haven't yet made the time to try out the suggestions (about whether rate limiting is zero or one based, trying to catch the actual rate-limit status code, and trying out the `RoundTripper`), but -- aside from the opt-in flag -- are those merge blockers or nice-to-haves?

mdaniel:
By coincidence, I actually experienced rate-limit exhaustion today, so I can report back some specifics:

Sadly, the API responds with `HTTP/1.1 403 Forbidden` instead of the much more sane `429`, so I'm not comfortable right now saying that I want to trap 403s and go spelunking in the response body for `{"message":"API rate limit exceeded for user ID ..."`, above and beyond the fact that retrying things under `api` makes me super nervous.

I also discovered that it is zero indexed, and it does actually deliver `X-Ratelimit-Remaining: 0` coming back in the `200` response right before the 403, so I'll straighten that out

mislav:
Also, this feature could use a test! See `features/api.feature`

We try to avoid slow tests, so if a scenario is triggering the `Sleep()` call, we should take precaution to not trigger it for longer than 1s.

mdaniel:
> Also, this feature could use a test! See `features/api.feature`

Great, now I need to do ruby :-( That `features` directory could use a `README.md` explaining the relationship between `steps.rb` and the `.feature` files, and/or put that information in `CONTRIBUTING.md` (which does kind of hint at it, but "study the existing ones" is not as friendly for newcomers, IMHO)

The `Vagrantfile` has gone bit-rotten:
```
    linux: Setting up ruby1.9.1-dev (1.9.3.484-2ubuntu1.14) ...
    linux: ERROR:  Error installing bundler:
    linux: 	bundler requires Ruby version >= 2.3.0.
```
and it also erroneously assumes the project is in a `$GOPATH`, which is no longer required thanks to `go.mod` and friends, and similarly its version of go 1.4 is hilariously out of date

After straightening all that out, `make test-all` fails, but I was able to fish out the command that I care about `cucumber -p all features/api.feature` and that's all I tested locally. A miserable "run the tests before pushing" experience, for sure

commands/api.go
```diff
@@ -292,13 +304,34 @@ func apiCommand(cmd *Command, args *Args) {
...
 			fmt.Fprintf(out, "\n")
 		}
+
+		if requestLoop && rateLimit {
+			var rateLimitLeft = -1
+			var rateLimitResetMs = -1
+			var atoiErr error
+			if xRateLimitLeft := response.Header.Get(rateLimitRemainingHeader); len(xRateLimitLeft) != 0 {
+				rateLimitLeft, atoiErr = strconv.Atoi(xRateLimitLeft)
+				utils.Check(atoiErr)
```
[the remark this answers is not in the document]
mdaniel: I am strongly against the very idea of "ignoring errors," as one can tell from my previous attempt to actually trap them before being asked to roll that back. If the project really feels strongly that `hub` should swallow errors, I'll capitulate to get this PR in, but I can assure you that as an end user of software, I have been bitten more times by swallowed errors than I have ever thought to myself "wow, this software talks about errors too much"

mdaniel:
> I think that it won't be too hard to detect the HTTP 403 that happened as a result of rate limiting. You can use ErrorInfo to inspect the message returned from API: 

So, as I asked previously: is that extra work mandatory to land this, or a nice-to-have? Because this started out as scratching an itch of mine, and finding how to trap rate limiting in all of those scenarios without inadvertently introducing weird bugs feels like Real Work&trade;

mislav:
I'm sorry you had a bad experience running tests. The Vagrantfile is misleading; it was for a different purpose than running tests in and is no longer accurate.

Don't worry about the rest of the functionality that I proposed. I can take over from here and get it over the finish line. Thank you for the work you've put into this.

mdaniel:
That's awesome that there is now a docker container; thank you for getting that in

And my thanks for all your continued patience with this. I'm sorry I didn't have the mental bandwidth right now to chase this issue all the way down to its ideal solution

mislav:
@mdaniel Thank you for the work so far. I've added some changes:

- Squashed your commits to simplify history
- Renamed the flag to `--obey-ratelimit` for clarity
- Cleaned up the implementation; e.g. removed the code and cukes around handling parse errors in ratelimit headers
- Have HTTP 403 responses with `--obey-ratelimit` be retried after waiting, even if no `--paginate`

Please review! 🙇

Issue #2281, opened 2019-09-29
hub release -F <FILE>


OmgImAlexis:
Is there an example file to show the format this accepts?

I've tried a few different ways and I keep getting the whole file as the title.

mislav:
When reading the release title and description from a file like this:

```
hub release create -F myfile.txt
```

The formatting rules follow the same logic [as for `--message`](https://hub.github.com/hub-release.1.html):
> The text up to the first blank line in \<MESSAGE> is treated as the release title, and the rest is used as release description in Markdown format.

Therefore, all you need is **one blank line between the title and the body**. Example:

```
This is my release title
This is also part of the title.

Release body starts here.
It can span several lines.

As well as several paragraphs.
This is also release body.
```

sinisterstuf:
The formatting rules are documented under --message and the manual page entry for -F refers to the rules for --message. Is there another piece of documentation that should updated/improved regarding this or is this issue closeable?

mislav:
@sinisterstuf Thanks for checking. I think we can close this since https://github.com/github/hub/commit/4813e859ae10d7561043f47a517fd7c5aeb821b3

PR #2318, opened 2019-10-21
changed help message for --message tag

srafi1:
Fixes issue #2281

### DIFF
--- a/commands/issue.go
+++ b/commands/issue.go
@@ -114,8 +114,10 @@ With no arguments, show a list of open issues.
 		of "always" (default for '--color'), "never", or "auto" (default).
 
 	-m, --message <MESSAGE>
-		The text up to the first blank line in <MESSAGE> is treated as the issue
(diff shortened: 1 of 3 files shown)


commands/issue.go
```diff
@@ -114,8 +114,10 @@ With no arguments, show a list of open issues.
 		of "always" (default for '--color'), "never", or "auto" (default).
 
 	-m, --message <MESSAGE>
-		The text up to the first blank line in <MESSAGE> is treated as the issue
-		title, and the rest is used as issue description in Markdown format.
+		An issue title and an issue description will be read from <MESSAGE>.
+		These should be separated by a blank line (ie: a line with no contents).
```
mislav: No, sorry, I don't believe this improves upon the old text, which reads:

> The text up to the first blank line in \<MESSAGE> is treated as the issue title, and the rest is used as issue description in Markdown format.

It looks to me that you have just expressed the same concept in a slightly different way. I don't see this as an improvement to our documentation.

PR #2325, opened 2019-10-24
Update README.md

anoop666:


mislav:
If you are testing hub or GitHub, it's best to do so in your own repo, rather than in this one! It avoids triggering notifications for hundreds of other people. Thank you!

anoop666:
Ok

PR #2326, opened 2019-10-25
Add support for `hub download --include` flag

### LINKED ISSUE #2251: hub release download - filtering and list
It would be great to add some functionalities to the `hub release download` command:
`--list-only` list the different assets attached to the release
`--include` a glob pattern to filter in assets to download
`--exclude` a glob pattern to filter out assets to download

By default, all assets are downloaded. When there is one asset per architecture, it is useless to download them all.

tomlazar:
fixes #2251

### DIFF
--- a/commands/release.go
+++ b/commands/release.go
@@ -102,7 +102,7 @@ With '--exclude-prereleases', exclude non-stable releases from the listing.
 		does not already exist (default: main branch).
 	
 	-i, --include <PATTERN>
-		Filter the files in the release to those that match the <PATTERN>.
+		Filter the files in the release to those that match the glob <PATTERN>.
 
 	-f, --format <FORMAT>
 		Pretty print releases using <FORMAT> (default: "%T%n"). See the "PRETTY
@@ -393,6 +393,7 @@ func downloadRelease(cmd *Command, args *Args) {
 	utils.Check(err)
 
 	pattern := args.Flag.Value("--include")
+	found := false
 	for _, asset := range release.Assets {
 		// if the include glob is not empty and it doesn't match the pattern
 		if pattern != "" {
@@ -403,9 +404,19 @@ func downloadRelease(cmd *Command, args *Args) {
 			}
 		}
 
+		found = true
 		ui.Printf("Downloading %s ...\n", asset.Name)
 		err := downloadReleaseAsset(asset, gh)
 		utils.Check(err)
+	}
+
+	if !found && pattern != "" {
+		errorMsg := fmt.Sprintf("glob pattern '%v' filters all files, possible assets are...\n", pattern)
+		for _, asset := range release.Assets {
+			errorMsg += fmt.Sprintf("%v\n", asset.Name)
+		}
+
+		utils.Check(fmt.Errorf(errorMsg))
 	}
 
 	args.NoForward()
--- a/commands/release.go
+++ b/commands/release.go
@@ -5,7 +5,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"time"
 
@@ -103,8 +102,7 @@ With '--exclude-prereleases', exclude non-stable releases from the listing.
 		does not already exist (default: main branch).
 	
 	-i, --include <PATTERN>
-		Display the files that match the prodived glob when looking at the
-		release files.	
+		Filter the files in the release to those that match the <PATTERN>.
 
 	-f, --format <FORMAT>
 		Pretty print releases using <FORMAT> (default: "%T%n"). See the "PRETTY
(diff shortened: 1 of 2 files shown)


commands/release.go
```diff
@@ -100,6 +101,10 @@ With '--exclude-prereleases', exclude non-stable releases from the listing.
 	-t, --commitish <TARGET>
 		A commit SHA or branch name to attach the release to, only used if <TAG>
 		does not already exist (default: main branch).
+	
+	-i, --include <PATTERN>
+		Display the files that match the prodived glob when looking at the
```
mislav: Does this flag affect the "displaying" or the _downloading_ of releases?

commands/release.go
```diff
@@ -386,7 +394,16 @@ func downloadRelease(cmd *Command, args *Args) {
 	release, err := gh.FetchRelease(project, tagName)
 	utils.Check(err)
 
+	include := args.Flag.Value("--include")
+	includeRe, err := regexp.Compile(include)
```
mislav: If the pattern is indeed a "glob", I had expected that it follows glob rules, not golang regexp rules. Glob patterns are defined roughly (example taken from `man bash`):

> `*`      Matches any string, including the null string.
> `?`      Matches any single character.
> `[...]`  Matches any one of the enclosed characters.

features/release.feature
```diff
@@ -636,7 +636,68 @@ MARKDOWN
...
+                },
+              ],
+            },
+          ]
+        }
+        get('/repos/mislav/will_paginate/assets/9876') {
+          halt 401 unless request.env['HTTP_AUTHORIZATION'] == 'token OTOKEN'
+          halt 415 unless request.accept?('application/octet-stream')
+          status 302
+          headers['Location'] = 'https://github-cloud.s3.amazonaws.com/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz'
```
mislav: It's not necessary that these tests perform a 302 redirect. They could just respond with a 200 and return the tarball "contents" here. After all, we are testing the `--include` flag in this scenario, not that redirects are followed (that is covered in other tests)

features/release.feature
```diff
@@ -636,7 +636,68 @@ MARKDOWN
...
+          ""
+        }
+        get('/releases/12204602/22ea221a-cf2f-11e2-222a-b3a3c3b3aa3a.gz', :host_name => 'github-cloud.s3.amazonaws.com') {
+          halt 400 unless request.env['HTTP_AUTHORIZATION'].nil?
+          halt 415 unless request.accept?('application/octet-stream')
+          headers['Content-Type'] = 'application/octet-stream'
+          "ASSET_TARBALL"
+        }
+        """
+        When I successfully run `hub release download v1.2.0 --include amd`
```
mislav: Could we test some pattern-matching features of the globa pattern here?

* I would expect that `amd` only matches the full string "amd" and nothing else
* If we wanted to match any download that has "amd" in the name, I would imagine that we'd use this glob pattern: `*-amd*`

commands/release.go
tomlazar: I started to look into that but couldn't find a way to do it in go without downloading a [new library](https://godoc.org/github.com/gobwas/glob) or re-implementing that myself.

tomlazar: nevermind filepath.Match does exactly what this needs

mislav: @tomlazar You can use stdlib for this https://golang.org/pkg/path/filepath/#Match

tomlazar: @mislav yep thanks, I found that and used it in the new commit

commands/release.go
```diff
@@ -394,14 +392,15 @@ func downloadRelease(cmd *Command, args *Args) {
...
+	pattern := args.Flag.Value("--include")
 	for _, asset := range release.Assets {
 		// if the include glob is not empty and it doesn't match the pattern
-		if include != "" && !includeRe.MatchString(asset.Name) {
-			continue
+		if pattern != "" {
+			isMatch, err := filepath.Match(pattern, asset.Name)
+			utils.Check(err)
+			if !isMatch {
+				continue
```
mislav: Should the `hub release download` process exit with an error if no assets matched the pattern? The stderr output could potentially print the available files

commands/release.go
```diff
@@ -103,8 +102,7 @@ With '--exclude-prereleases', exclude non-stable releases from the listing.
 		does not already exist (default: main branch).
 	
 	-i, --include <PATTERN>
-		Display the files that match the prodived glob when looking at the
-		release files.	
+		Filter the files in the release to those that match the <PATTERN>.
```
mislav: Could we clarify that this is a glob pattern? E.g.

> Download only the assets whose name matches the glob \<PATTERN>.

tomlazar: yep, added that in.

commands/release.go
tomlazar: sounds good, I was going to add a message if there were no assets at all but I don't think that's possible since a release will contain the source code at the time.

PR #2341, opened 2019-11-04
docs: remove `--classic` from Ubuntu install

mightyiam:
I fee it's better to let snap explain to the user than to bypass that message:

```
$ snap install hub
error: This revision of snap "hub" was published using classic confinement and thus may perform
       arbitrary system changes outside of the security sandbox that snaps are usually confined to,
       which may put your system at risk.

       If you understand and want to proceed repeat the command including --classic.
```


mislav:
Thanks for the suggestion, but I would rather keep installation instructions that work rather than ones that don't.

The message shown by `snap install hub` is certainly useful, but we assume that users are already familiar with how Snap works and what `--classic` means, and if they are not, we assume that they are going to get educated elsewhere rather than through our documentation.

mightyiam:
Your call but I'd like to disagree. `--classic` bypasses security.

mislav:
`--classic` is no less secure than installing Debian packages. https://github.com/felicianotech/hub-snap/issues/8#issuecomment-462224520

PR #2348, opened 2019-11-08
Move regex compilation to package variables

yaserazfar:
The regexp.MustCompile function panics if the regex fails to compile and should only be used in globals or init functions, otherwise if the regex is invalid it may panic at run time.

I have moved all the occurrences of the function to be global variables so that they are checked during compile time.

### DIFF
--- a/commands/alias.go
+++ b/commands/alias.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strings"
 
 	"github.com/github/hub/ui"
@@ -115,8 +114,7 @@ end`
 			eval = `eval "$(hub alias -s)"`
 		}
 
-		indent := regexp.MustCompile(`(?m)^\t+`)
-		eval = indent.ReplaceAllStringFunc(eval, func(match string) string {
+		eval = indentRe.ReplaceAllStringFunc(eval, func(match string) string {
 			return strings.Repeat(" ", len(match)*4)
 		})
 
--- a/commands/apply.go
+++ b/commands/apply.go
@@ -55,6 +55,12 @@ hub-apply(1), hub-cherry-pick(1), hub(1), git-am(1)
 `,
 }
 
+var (
+	gistRegexp   = regexp.MustCompile("^https?://gist\\.github\\.com/([\\w.-]+/)?([a-f0-9]+)")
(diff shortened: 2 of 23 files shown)


mislav:
Hi Yaser, thank you for your contribution.

Your pull request title and description are very well put together and it was easy for me to understand what change you are proposing and for what reason. 👍 

However, when I reviewed the diff itself, what struck me was:

- **The diff is relatively large** (105 changed lines) for just moving variable definitions around. We project maintainers generally avoid merging large diffs unless it's for a really good reason. I rarely merge contributions that are mostly cosmetic changes to my projects because they might cause new bugs or merge conflicts for other contributions down the line, but the original change itself dosn't contribute almost nothing to better functionality of the software, and (in the end) the risk vs. payoff tradeoff isn't really worth it for me.

- **The fix isn't addressing any real problem** that the hub project is facing right now. You are right that `MustCompile` can panic at runtime if supplied with an invalid regex, but in practice _every current regex in the codebase is valid_, so none of them will ever panic. If we tried to introduce an invalid regex to the codebase, the associated test suite will fail in CI, and such changes would not get merged until the regex was fixed. In the years that I was maintaining hub (this project has 1,500 total issues), nobody has ever reported that they have a problem with MostCompile panicking at runtime. Therefore, this PR attempts to solve a problem that, for our project, only exists in theory but not in practice. Attempting to solve such problems is often not a good allocation of software development resources.

- **Ultimately, this fix doesn't change any behavior.** I agree that it would be better if MustCompile defined at package level would error at build time instead of panicking at run time, but moving them in a way that you did still results in them erroring out in runtime.

Consider this program:

```go
package main

import "fmt"
import "regexp"

var re = regexp.MustCompile(`*`)

func main() {
	fmt.Println(re)
}
```

By your PR description, this approach is supposed to make MustCompile error out at build time, but with Go 1.13, I'm not seeing this behavior. This program compiles, and it will panic at runtime.

With all things considered, I will not be merging this PR, nor will I be providing code suggestions on how to make the change better. The hub codebase has many problems and bugs—as evident by open issues in this issue tracker—which I would welcome solutions too, but invalid regular expressions is not one of them.

Thank you again! 🙇

yaserazfar:
Thank you for your kind feedback Mislav!
You made it very easy to understand why my PR could not be merged. I also just tested out the MustCompile as a global variable and what you said was true, which is weird because from my understanding of golang, I thought global variables and init functions are compiled before run-time.

PR #2355, opened 2019-11-11
allow editing release tag name

### LINKED ISSUE #2347: [Feature request] Editing tag names via `hub release edit`
GithubAPI supports [editing tag names](https://developer.github.com/v3/repos/releases/#edit-a-release), but  It seems that hub doesn't support it.

0xch4z:
Closes #2347

### DIFF
--- a/commands/release.go
+++ b/commands/release.go
@@ -107,6 +107,9 @@ With no arguments, shows a list of existing releases.
 	-i, --include <PATTERN>
 		Filter the files in the release to those that match the glob <PATTERN>.
 
+	--newtag <TAG-NAME>
+		Update the tag associated with the release.
+
 	-f, --format <FORMAT>
 		Pretty print releases using <FORMAT> (default: "%T%n"). See the "PRETTY
 		FORMATS" section of git-log(1) for some additional details on how
@@ -209,6 +212,7 @@ hub(1), git-tag(1)
 		-m, --message MSG
 		-F, --file FILE
 		-t, --commitish C
+		--newtag TAG
 `,
 	}
 
@@ -568,6 +572,9 @@ func editRelease(cmd *Command, args *Args) {
 	params := map[string]interface{}{}
 	if args.Flag.HasReceived("--commitish") {
 		params["target_commitish"] = args.Flag.Value("--commitish")
+	}
+	if args.Flag.HasReceived("--newtag") {
+		params["tag_name"] = args.Flag.Value("--newtag")
 	}
 	if args.Flag.HasReceived("--draft") {
 		params["draft"] = args.Flag.Bool("--draft")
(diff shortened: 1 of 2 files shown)


commands/release.go
```diff
@@ -104,6 +104,9 @@ With '--exclude-prereleases', exclude non-stable releases from the listing.
 	-i, --include <PATTERN>
 		Filter the files in the release to those that match the glob <PATTERN>.
 
+	-T, --tagname <NEW-TAG-NAME>
```
mislav: Since I don't imagine this flag being used too often, we probably don't need the shorthand `-T`. Also, I'm thinking that the flag might benefit from a more descriptive name, since it might not be clear what `--tagname FOO` will do when editing a release.

What do you think about:
- `--newtag <TAG-NAME>`
- `--edit-tag <NEW-TAG-NAME>`

Charliekenney23: For sure, `-T` does feel unnecessary. I was just trying to be most consistent- I'll remove it.

I like `--newtag`. I think that makes much more sense!

shadoath:
Bump! Would love to see this feature merged in.

commands/release.go
shadoath, line 107:
```suggestion
	--newtag <NEW-TAG-NAME>
```

commands/release.go
```diff
@@ -206,6 +209,7 @@ hub(1), git-tag(1)
 		-m, --message MSG
 		-F, --file FILE
 		-t, --commitish C
+		-T, --tagname
```
shadoath, line 212:
```suggestion
		-newtag
```

commands/release.go
```diff
@@ -566,6 +570,9 @@ func editRelease(cmd *Command, args *Args) {
 	if args.Flag.HasReceived("--commitish") {
 		params["target_commitish"] = args.Flag.Value("--commitish")
 	}
+	if args.Flag.HasReceived("--tagname") {
+		params["tag_name"] = args.Flag.Value("--tagname")
+	}
```
shadoath, line 575:
```suggestion
	if args.Flag.HasReceived("--newtag") {
		params["tag_name"] = args.Flag.Value("--newtag")
	}
```

shadoath (commented):
Added suggestions for changes.

lion-starsirio (approved):

shadoath:
@mislav & @Charliekenney23 Can we get another review and update of this PR?
Would be really nice feature to add to what hub can do.
Thank you

Charliekenney23:
Woah, completely forgot about this PR. Just addressed the feedback and wrote a test. Should be good now.

shadoath:
@Charliekenney23 Looks like those tests didn't pass.

Charliekenney23:
> @Charliekenney23 Looks like those tests didn't pass.

@shadoath I can see that there is a failure...

If you actually read the details of the above workflow runs you'll see that the same `features/fork.feature:162` test is the culprit in the upstream master branch this commit is based on. Please put more effort into investigating before pinging

PR #2357, opened 2019-11-13
`script/version`: make sure to look at the correct `.git/` directory

dscho:
I noticed while trying to build `hub` from an unpacked archive inside an unrelated repository (https://github.com/git-for-windows/MINGW-packages) that `script/version` was happy to pick up incorrect tags.

This PR fixes that.

### DIFF
--- a/script/version
+++ b/script/version
@@ -2,6 +2,7 @@
 # Displays hub's release version
 set -e
 
+export GIT_CEILING_DIRECTORIES=${PWD%/*}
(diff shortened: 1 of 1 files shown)


script/version
```diff
@@ -2,6 +2,7 @@
 # Displays hub's release version
 set -e
 
+export GIT_CEILING_DIRECTORIES=${PWD%/*}
```
mislav: Good catch!! Thank you 🌟

Issue #2252, opened 2019-09-02
Issue transferring


mlewand:
Not that long ago GH added a feature to transfer a single issue.

It would be nice to add this feature in `hub` binary too.

I guess the basic syntax might looks something like:

```
hub issue transfer NUMBER TARGET_REPO

# e.g.
hub issue transfer 123 myUser/newRepo
```

My use case is that I'm centralizing issue tracker for multirepo architecture. So I want to use `hub` binary to move a ton of issues to another repository.

mislav:
This is a good feature idea. I thought that it wasn't possible to transfer issues using the GitHub API, but as it turns out this is possible using the GraphQL API, so I have written a `hub api` wrapper script to demonstrate how you can programatically do this today: https://github.com/mislav/hub-api-utils/blob/master/bin/hub-issue-transfer

Note that if you're transferring issues in bulk, you can probably edit that script to also look up issue/repository node IDs (needed for the `transferIssue` mutation) in bulk, which would potentially speed up the overall script.

Since I believe all this should be simpler and potentially baked into hub, I will keep this issue open as a feature request. In the meantime please use `hub api` to programatically achieve what you need and let me know if you encounter any issues.

mlewand:
Amazing, thank you for the input @mislav.

srafi1:
Is anyone working on this? If not, I'd like to give it a shot.

mislav:
@srafi1 Go ahead! 👍

mislav:
This was implemented 🙇 🏆 https://github.com/github/hub/commit/4c2e44146988dfb385a26f649298f274a5017756

PR #2363, opened 2019-11-17
Transfer Issues to Other Repositories

srafi1:
Implements transferring issues to other repositories as requested in issue #2252. This feature differs from the feature request in one way: the user only specifies TARGET_REPO rather than USER/TARGET_REPO. This is because users cannot move issues between different users or organizations according to [this guide](https://help.github.com/en/github/managing-your-work-on-github/transferring-an-issue-to-another-repository).

Format:
```
hub issue transfer <ISSUE_NUMBER> <TARGET_REPO>
```

Example:
```
hub issue transfer 2252 hub-fork
```


commands/issue.go
```diff
@@ -710,3 +720,105 @@ func milestoneValueToNumber(value string, client *github.Client, project *github
...
+                }
+            }
+        }`, issueNumber, owner, currRepo)
+	body := make(map[string]interface{})
+	body["query"] = query
+
+	response, err := gh.GenericAPIRequest("POST", "graphql", body, nil, 0)
+	utils.Check(err)
+
+	responseData := make(map[string]interface{})
```
mislav: You can have this be:

```go
responseData := struct {
  Data struct {
    Repository struct {
      Issue struct {
        ID string `json:"id"`
      }
    }
  }
  Errors []struct {
    Message string
  }
}{}
```

And then you can umarshal into that and read `responseData.Data.Repository.Issue.ID` directly without using type assertions.

commands/issue.go
```diff
@@ -710,3 +720,105 @@ func milestoneValueToNumber(value string, client *github.Client, project *github
...
+	response, err = gh.GenericAPIRequest("POST", "graphql", body, nil, 0)
+	utils.Check(err)
+
+	responseData = make(map[string]interface{})
+	err = response.Unmarshal(&responseData)
+	utils.Check(err)
+
+	if _, keyExists := responseData["errors"]; keyExists {
+		ui.Errorf("Error transferring the issue\n")
+		os.Exit(1)
```
mislav: Tip: `utils.Check(err)` does the printing and `os.Exit(1)` combo for you 😉

commands/issue.go
```diff
@@ -21,6 +21,7 @@ issue [-a <ASSIGNEE>] [-c <CREATOR>] [-@ <USER>] [-s <STATE>] [-f <FORMAT>] [-M
 issue show [-f <FORMAT>] <NUMBER>
 issue create [-oc] [-m <MESSAGE>|-F <FILE>] [--edit] [-a <USERS>] [-M <MILESTONE>] [-l <LABELS>]
 issue labels [--color]
+issue transfer <ISSUE_NUMBER> <TARGET_REPO>
```
mislav: Nit: we prefer using hyphens here 
```suggestion
issue transfer <ISSUE-NUMBER> <TARGET-REPO>
```

mislav:
This also needs tests in order to have CI passing again. You can find existing tests in `features/issue.feature` and run them with `bin/cucumber features/issue.feature`, but we don't have any examples of tests that handle GraphQL API calls since we didn't use GraphQL before this. If this proves to hard for you to write, just holler and I will take over. 🙇

srafi1:
Thanks for the feedback! I implemented the method for unmarshalling that you suggested and the code looks much cleaner now without type assertions. I haven't tried writing any tests yet, so I will try that (hopefully during the weekend).

commands/issue.go
```diff
@@ -710,3 +721,106 @@ func milestoneValueToNumber(value string, client *github.Client, project *github
...
+	project, err := localRepo.MainProject()
+	utils.Check(err)
+
+	owner := project.Owner
+	currRepo := project.Name
+	host := project.Host
+	gh := github.NewClient(host)
+
+	query := fmt.Sprintf(`
+        query($issue: Int = %s, $owner: String = "%s", $repo: String = "%s") {
```
mislav: Even though this works, the query should ideally not contain any values:
```suggestion
        query($issue: Int!, $owner: String!, $repo: String!) {
```
instead, pass values as `variables` parameter in JSON:
```go
variables := map[string]interface{}{
  "issue": issueNumber,
  "owner": owner,
  "repo": currRepo,
}
body["variables"] = variables
```

srafi1:
I put variables in their own parameter, but I haven't written any tests. I looked at `steps.rb` and `issue.feature`, but I'm not familiar with Ruby or Cucumber, so I don't know how to write them. Can you help me out with this?

mislav:
@srafi1 Great work; thank you! I've also added some cleanups and tests.

One thing that only ocurred to me after I took over your branch was that the first 2 GraphQL queries can be performed in a single request (rather than over 2 requests), so I've implemented that.

I've also added a specialized `Client.GraphQL` function to move us away from `GenericAPIRequest` since the latter was only intended for the `hub api` command.

PR #2369, opened 2019-11-20
Add file location when using snap

SuperSandro2000:


etc/README.md
```diff
@@ -19,6 +19,8 @@ if [ -f /path/to/hub.bash_completion ]; then
 fi
 ```
 
+If you are using snap you can point to ``/snap/hub/current/go/src/github.com/github/hub/etc/hub.bash_completion.sh``
```
mislav: Thank you for the tip, but I'm not sure that this is the way to solve this. This file location seems "internal", as in it might change if the structure of the snap change or we change the Go build process. I don't want to suggest our users to reference any files within the snap, as the whole point of containerisation is that there are boundaries between filesystems.

If people want shell completion, they can download the files from this repo and set them up in their shell.

SuperSandro2000: > If people want shell completion, they can download the files from this repo and set them up in their shell.

There is a fundamental problem with that and snaps: The snap gets auto updated but the shell completion does not. I tried to solve it this way but if you know a better one which maybe "mounts" the shell completion to right location, eg ``/etc/bash_completion.d/`` this would be awesome. I think anything in 2019 with more than 2 options should have a shell completion which ideally is setup automatically.

mislav: Agreed with all your points.

> I think anything in 2019 with more than 2 options should have a shell completion which ideally is setup automatically.

That would be great. If there was a simple API I could implement when releasing binaries for hub that everyone's system (macOS, Windows, Linux, other *nix) would then pick up and automatically wire up shell completions, I would be very happy. I am not aware that such an API exists, as everyone seems to set up their shell in a different way.

Are `/etc/bash_completion.d/*` sourced by default for all bash users on Linux?

SuperSandro2000: The default bash_completion configuration on Ubuntu sources all files in ``/etc/bash_completion.d/`` for all users almost at the end of ``/usr/share/bash-completion/bash_completion``.

My directory contains the following files already:
```
ls /etc/bash_completion.d/
apport_completion  docker-compose  fail2ban  git-prompt
```
Note: I put the docker-compose file there.

For cmd I don't think that there is any completion by default without anything like cmder or clink. I am unsure about macOS cause I never used one.

felicianotech:
@SuperSandro2000 If I understood this PR correctly, you were trying to get Bash completion for hub to work when using the hub snap.

If so, there's an open issue in the repo for the snap where we're tracking this request: https://github.com/felicianotech/hub-snap/issues/26

mislav is correct that you don't want to do it the way you were trying and that way will likely break at some point.

PR #2373, opened 2019-11-27
Update and correct README

CyberFlameGO:
Added a new heading called hub because that usually makes a documentation more welcoming, as well a few other edits seen in the commit page


mislav:
Thank you, but I don't think any of these changes substantially improve our documentation. I feel that the GitHub.com-vs-hub paragraph is now even harder to read.

PR #2374, opened 2019-11-27
Update and correct README

CyberFlameGO:
Added a new heading called hub because that usually makes a documentation more welcoming, as well a few other edits seen in the commit page


mislav:
https://github.com/github/hub/pull/2373#issuecomment-559134758

PR #2378, opened 2019-12-02
commands: consult global color.ui setting when --color=auto

mariusae:
This makes the behavior consistent with other git commands.
Also fix up a bug in the documentation of `--color`.


mislav:
Thank you for the contribution! I've pushed some improvements, and also reverted the change to docs; I don't think it's necessary to explicitly point out that `--color` respects `git config color.ui`. Since hub tries to behave like git, respecting git config should go without saying.

mariusae:
Thanks!

tedmiston:
my watch scripts thank you for this pr @mariusae :)

Issue #2379, opened 2019-12-04
Crash report - interface conversion: interface {} is nil, not string


macrat:
Error (*runtime.TypeAssertionError): `interface conversion: interface {} is nil, not string`

Stack:

```
goroutine 1 [running]:
panic(0x138a5c0, 0xc00013c150)
	/usr/local/Cellar/go/1.13.4/libexec/src/runtime/panic.go:679 +0x1b2
github.com/github/hub/github.(*yamlConfigDecoder).Decode(0x16fc9f0, 0x1478dc0, 0xc000148000, 0xc00013a060, 0xc000148000, 0x0)
	github/config_decoder.go:52 +0x66d
github.com/github/hub/github.(*configService).Load(0xc0000ada30, 0xc000144060, 0x1b, 0xc00013a060, 0x0, 0x0)
	github/config_service.go:42 +0xf5
github.com/github/hub/github.CurrentConfig(0xc00001828c)
	github/config.go:326 +0xe1
github.com/github/hub/github.newProject(0xc000018285, 0x6, 0xc00001828c, 0x6, 0xc00001827a, 0xa, 0xc000018270, 0x3, 0x4)
	github/project.go:164 +0x347
github.com/github/hub/github.NewProjectFromURL(0xc0000a4300, 0x1, 0x0, 0x0)
	github/project.go:129 +0x134
github.com/github/hub/github.(*Remote).Project(0xc0000adbc8, 0x0, 0x0, 0x9)
	github/remote.go:27 +0x33
github.com/github/hub/github.(*GitHubRepo).MainProject(0xc00000e840, 0x0, 0x0, 0xc0000adbe8)
	github/localrepo.go:252 +0x99
github.com/github/hub/commands.ciStatus(0x16d8de0, 0xc0000fe240)
	commands/ci_status.go:90 +0xa2
github.com/github/hub/commands.(*Command).Call(0x16d8de0, 0xc0000fe240, 0xc000099a70, 0x1)
	commands/commands.go:45 +0x7f
github.com/github/hub/commands.callRunnableCommand(0x16d8de0, 0xc0000fe240, 0x7ffeefbff964, 0x9)
	commands/runner.go:87 +0x39
github.com/github/hub/commands.(*Runner).Execute(0xc000010058, 0xc00000e060, 0x2, 0x2, 0xc000000180, 0xc0000adf50)
	commands/runner.go:70 +0x436
main.main()
	main.go:17 +0x7b

```

Runtime:

```
GOOS: darwin
[...]

PR #2380, opened 2019-12-04
Add type assertion test to YAML config decoder

macrat:
In my first time to use `hub`, my configuration file was incorrect.
And `hub` was crushed because of failed to type conversion. Please see #2379.

I added type check into the YAML configuration decoder.
Now, `hub` won't crush even if given an incorrect configuration file. It will just ignored it.

I think probably it more friendly if reporting about incorrect configuration.
But I didn't edit except a configuration decoder at the moment.

### DIFF
--- a/fixtures/test_configs.go
+++ b/fixtures/test_configs.go
@@ -71,3 +71,44 @@ github.com:
 
 	return &TestConfigs{file.Name()}
 }
+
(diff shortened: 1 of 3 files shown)


mislav:
This is great! Thank you 🎉 🎉

PR #2388, opened 2019-12-11
Update Debian package name in README

alexfornuto:
Title says it all.

FWIW I needed to install as `git-hub` on "Debian" inside of ChromOS, so there _may be_ a discrepancy there. If so, then this could be a new line instead.


alexfornuto:
Never mind - apparently that's a different package and Hub isn't available in Deb9, which is what Crostini ships by default.

PR #2390, opened 2019-12-14
use `command grep` vs `grep`

dsifford:
This change accounts for uses who might have an alias set to grep which is not compatible with `grep -E` (e.g. `alias grep=rg`)


mislav:
Thank you!

PR #2428, opened 2019-12-22
improve Project SameAs logic

johanavril:
use strings.EqualFold instead of strings.ToLower with equal
operation when comparing string ignore case

BenchmarkSameAsToLower-8     	29738065	        39.1 ns/op	       0 B/op	       0 allocs/op
BenchmarkSameAsEqualFold-8   	42597748	        25.5 ns/op	       0 B/op	       0 allocs/op

### DIFF
--- a/github/project.go
+++ b/github/project.go
@@ -23,9 +23,9 @@ func (p Project) String() string {
 }
 
 func (p *Project) SameAs(other *Project) bool {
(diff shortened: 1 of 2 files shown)


johanavril:
thanks for the feedback, I updated the test case data to exercise the case folding.

Issue #1551, opened 2017-08-27
HUB_VERBOSE environment variable is undocumeneted


e-beach:
The HUB_VERBOSE environment variable, which causes network requests and git commands executed by hub to be logged to stdout, is useful for both contributors and end-users. For contributors, it is obviously useful for debugging, but it is also useful for users who want to understand the commands that hub executes, especially when hub behaves unexpectedly.

I was only able to discover this variable by looking through commands.go:

https://github.com/github/hub/blob/a45f4573a32e346d7046b47657263cd17cb60c9e/cmd/cmd.go#L111-L119

Would it be appropriate to document it in the contributing guide?

mislav:
Yes it would absolutely be useful to document it both in the contributing guide and the man page. Thank you for suggesting

coleGillespie:
it is still a bit confusing in the documentation. i was not sure what to set it to. it looks like i can set it to anything other than empty string. i spent more time that i should have trying to figure that out.

PR #2430, opened 2019-12-30
Update hub.1.md

coleGillespie:
Give more context as to the expected value of the environment variable for environment variable `HUB_VERBOSE`
see comment https://github.com/github/hub/issues/1551#issuecomment-569767297


mislav:
Thank you for the suggestion. To avoid phrasing `set this to anything other than ""`, which sounds awkward because it describes the behavior using negation, I added the wording "if this environment variable is set", making it closer to how git documents boolean environment variables.

Issue #1686, opened 2018-02-25
Edit existing issues


alphaCTzo7G:
Hi all,

Is it possible to edit existing issues in a repo?

I checked the manpage and `hub issue --help`, but there is no information about editing a existing issue. 

I know there are other tools https://github.com/node-gh/gh, https://github.com/stephencelis/ghi, which can do this.

Perhaps these could be merged into `hub`, instead of replicating their work..

mislav:
See https://github.com/github/hub/issues/893#issuecomment-368439928

Issue #1523, opened 2017-07-28
Updating a pull request


devedse:
I would like to update the description of a pull request.

Is this already implemented or can I make a feature request of this?

See: https://developer.github.com/v3/pulls/#update-a-pull-request

e-beach:
I'm fairly certain that this feature does not currently exist. The maintainer seems to not have much time to devote to this project, so good luck with getting your PR merged.

mislav:
Updating a PR isn't implemented yet, but since we have `hub release edit` command already, editing of issues/PRs might be implemented in the future.

devedse:
For now I've used the octokit.net library which worked even better for me since hub in combination with the command line seems to have a max pull request body length.

For me this issue has low priority as of now :smile:.

strokirk:
Being able to update the PR would be great, even if just certain parts initially.

I'm combining `hub` with [`ghi`](https://github.com/stephencelis/ghi) for my current workflow of updating the labels of PRs from the command line, but it would be great if this was possible using just `hub`.

Here is a short script using them together: https://gist.github.com/strokirk/afd1b292d51f57453bfb2d609e75792d

mislav:
> since hub in combination with the command line seems to have a max pull request body length.

There is no code in hub that limits pull request body length. If you think you have discovered a bug, please open another issue and tell us how you passted the body text to `hub pull-request`, thank you!

mislav:
In hub master there is now the `hub issue update <NUMBER>` command #2432 which can update the following fields of an issue or pull request:
- title
- body
- labels
- assignees
- milestone
- state

dcsobral:
I've tried updating "draft" via the example in #2312, and it doesn't work. Updating other things such as "title" worked fine.

vlerenc:
Same experience here @dcsobral. I can change fields like the title with `hub`, the v3 ReST API, and the v4 GraphQL API, but I have no clue how to `convert-to-draft` or `ready-for-review` a PR. I see the GitHub web UI calling backend functions named this way, but neither the API docs show me how to change the draft status nor would the `hub` allow to do that (the above call seems to explore the generic API call feature anyhow, so this doesn't come as a surprise, but I tried it nevertheless/just in case).

Do you in the meantime or does anybody know how to tweak the draft status of a GitHub PR programmatically?

mislav:
[...]

PR #2432, opened 2020-01-04
Add the ability to edit issues

jfahrer:
This PR adds the ability to edit existing issues using the same interface as `hub issue create`. 

There is a little duplication between creating and editing issues that could be removed. I'm not convinced that removing the duplication would increase clarity and readability.

### DIFF
--- a/commands/issue.go
+++ b/commands/issue.go
@@ -647,9 +647,6 @@ func updateIssue(cmd *Command, args *Args) {
 
 	gh := github.NewClient(project.Host)
 
-	issue, err := gh.FetchIssue(project, strconv.Itoa(issueNumber))
-	utils.Check(err)
-
 	params := map[string]interface{}{}
 	setLabelsFromArgs(params, args)
 	setAssigneesFromArgs(params, args)
@@ -659,21 +656,25 @@ func updateIssue(cmd *Command, args *Args) {
 		messageBuilder := &github.MessageBuilder{
 			Filename: "ISSUE_EDITMSG",
 			Title:    "issue",
-			Edit:     args.Flag.Bool("--edit"),
-			Message:  strings.Replace(fmt.Sprintf("%s\n\n%s", issue.Title, issue.Body), "\r\n", "\n", -1),
 		}
 
 		messageBuilder.AddCommentedSection(fmt.Sprintf(`Editing issue #%d for %s
 
 Update the message for this issue. The first block of
-text is the title and the rest is the description.`, issue.Number, project))
+text is the title and the rest is the description.`, issueNumber, project))
 
+		messageBuilder.Edit = args.Flag.Bool("--edit")
 		flagIssueMessage := args.Flag.AllValues("--message")
 		if len(flagIssueMessage) > 0 {
 			messageBuilder.Message = strings.Join(flagIssueMessage, "\n\n")
 		} else if args.Flag.HasReceived("--file") {
 			messageBuilder.Message, err = msgFromFile(args.Flag.Value("--file"))
 			utils.Check(err)
+		} else {
+			issue, err := gh.FetchIssue(project, strconv.Itoa(issueNumber))
+			utils.Check(err)
+			existingMessage := fmt.Sprintf("%s\n\n%s", issue.Title, issue.Body)
+			messageBuilder.Message = strings.Replace(existingMessage, "\r\n", "\n", -1)
 		}
 
 		title, body, err := messageBuilder.Extract()
--- a/commands/issue.go
+++ b/commands/issue.go
@@ -20,8 +20,8 @@ var (
 issue [-a <ASSIGNEE>] [-c <CREATOR>] [-@ <USER>] [-s <STATE>] [-f <FORMAT>] [-M <MILESTONE>] [-l <LABELS>] [-d <DATE>] [-o <SORT_KEY> [-^]] [-L <LIMIT>]
 issue show [-f <FORMAT>] <NUMBER>
 issue create [-oc] [-m <MESSAGE>|-F <FILE>] [--edit] [-a <USERS>] [-M <MILESTONE>] [-l <LABELS>]
-issue labels [--color]
 issue update <NUMBER> [-m <MESSAGE>|-F <FILE>] [--edit] [-a <USERS>] [-M <MILESTONE>] [-l <LABELS>]
+issue labels [--color]
 issue transfer <NUMBER> <REPO>
 `,
 		Long: `Manage GitHub Issues for the current repository.
@@ -35,6 +35,10 @@ With no arguments, show a list of open issues.
 
 	* _create_:
 		Open an issue in the current repository.
+
+	* _update_:
+		Update fields of an existing issue specified by <NUMBER>. Use ''--edit''
+		to edit the title and message interactively in the text editor.
 
 	* _labels_:
 		List the labels available in this repository.
@@ -631,15 +635,7 @@ func updateIssue(cmd *Command, args *Args) {
 	if issueNumber == 0 {
 		utils.Check(cmd.UsageError(""))
 	}
-
-	flagsUpdatingFields := []string{"--message", "--edit", "--file", "--labels", "--milestone", "--assign"}
-	updatesFields := false
-	for _, name := range flagsUpdatingFields {
-		if args.Flag.HasReceived(name) {
-			updatesFields = true
-		}
-	}
-	if !updatesFields {
+	if !hasField(args, "--message", "--file", "--labels", "--milestone", "--assign", "--edit") {
 		utils.Check(cmd.UsageError("please specify fields to update"))
 	}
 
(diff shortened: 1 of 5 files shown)


commands/issue.go
```diff
@@ -618,6 +635,97 @@ text is the title and the rest is the description.`, project))
...
+	flagIssueAssignees := commaSeparated(args.Flag.AllValues("--assign"))
+	if len(flagIssueAssignees) > 0 {
+		params["assignees"] = flagIssueAssignees
+	}
+
+	milestoneNumber, err := milestoneValueToNumber(args.Flag.Value("--milestone"), gh, project)
+	utils.Check(err)
+	if milestoneNumber > 0 {
+		params["milestone"] = milestoneNumber
+	}
```
jfahrer: These flags could be moved into shared functions that could be called from edit and create. That might add additional indirection and I would personally keep it the way it is. Let me know what you think.

mislav:
For posterity: ref. #1686 #1523

jfahrer:
I addressed the issues you pointed out. Let me know if there is anything else.

> ```
> hub issue edit 1337 --edit
> ```
> 
> The fact that one has to write "edit" twice feels awkward to me, but I don't know what's a better solution.
Agreed. Maybe a `--no-edit` flag would make more sense here? Then we could open the editor by default and one can still script updating an issue without changing the title/body

> I'm not sure which would be the better way in the long run. Since this is a new feature, perhaps we could consider another API that's even better than both `hub issue edit --edit` and `hub release edit -m ""`? However, I won't require code changes from you around this until we reach some consensus.
If you have ideas for the API let me know, I'm happy to tackle them.

jfahrer:
@mislav, any thoughts on the PR and API?

mislav:
@jfahrer Sorry, I had to think this over.

Off the top of my head, these are parts of an issue that a person might want to edit:
- **title** (potentially in editor)
- **body** (potentially in editor)
- **assignees** (set of multiple)
- **labels** (set of multiple)
- **milestone**

We want to enable the user to be able to programatically edit one or any combination of these fields. We also want to enable the user to _interactively_ edit title & body by opening their text editor, similarly to their experience with `create`.

Lets look at some hypothetical `edit` usage and try to guess what it does:

- `hub issue edit 123` - I feel that this should open the text editor for editing title & body; i.e. I don't think that the user should pass `--edit` to opt into that.

- `hub issue edit 123 --labels foo,bar` - Here it feels like the user wants to replace the old labels on an issue with labels "foo", "bar". Automatically opening an interactive text editor for editing title & body here might be disruptive to the user's original goal, so maybe we shouldn't do that?
  - `hub issue edit 123 --edit --labels foo` - Perhaps this is how we can opt-in to editor even if we're changing other fields too?
  - `hub issue edit --labels ""` - Could we enable clearing existing labels by setting them to an empty value?
  - `hub issue edit --labels "+foo,-bar"` - Should we consider supporting selectively adding/removing labels using a special syntax like this?

- `hub issue edit 123 -m "hello"` - Here it looks like the user wants to edit the issue title non-interactively. What shuld happen with existing issue body; should it remain the same or be blanked out?

- `hub issue edit 123 -m "new title" -m "new body"` - Set both title and body to new values non-interactively. I think this will work with your current implemenation. Should we support editing only body while keeping title the same? How would we go about that.

Sorry for so many questions. I'm just brainstorming how I envision editing on the command-line to go. I think the reason we didn't support editing so far is because is so much more complex than creating. Anyway, any thoughts that you have are welcome.

jfahrer:
> @jfahrer Sorry, I had to think this over.

Thanks for taking the time!


> * `hub issue edit 123` - I feel that this should open the text editor for editing title & body; i.e. I don't think that the user should pass `--edit` to opt into that.

An alternative could be using a different keyword than `edit`. Maybe `hub issue update 123`? In this case, the `hub issue update 123 --edit` is less weird. If no options are passed, we could just open the issue on github.com

> * `hub issue edit 123 --labels foo,bar` - Here it feels like the user wants to replace the old labels on an issue with labels "foo", "bar". Automatically opening an interactive text editor for editing title & body here might be disruptive to the user's original goal, so maybe we shouldn't do that?

Since the editor would contain the current title and body I'm not sure it would be that disruptive. There should however be a way to prevent the editor from opening. I saw `--no-edit` somewhere else in the API. Overall, I guess this is an argument for keeping the `--edit` flag tho. 

>   * `hub issue edit 123 --edit --labels foo` - Perhaps this is how we can opt-in to editor even if we're changing other fields too?

It feels a little confusing that the behavior changes based on whether certain flags are passed or not. I feel like my prior comment about changing the keyword is better suited here.

>   * `hub issue edit --labels ""` - Could we enable clearing existing labels by setting them to an empty value?
>   * `hub issue edit --labels "+foo,-bar"` - Should we consider supporting selectively adding/removing labels using a special syntax like this?
>   * `hub issue edit --add-labels foo --remove-labels bar` - …or like this?

I think the `--add-labels` and `--remove-labels` would be easier from a users perspective. We can support those alongside `--labels`. Clearing labels with `--labels ""` makes sense to me. I think this might work already but I have to test.

> * `hub issue edit 123 -m "hello"` - Here it looks like the user wants to edit the issue title non-interactively. What should happen with existing issue body; should it remain the same or be blanked out?

You are right. I wonder if we should support `--title` and `--body` as flags instead? And ditch supporting `-m` for this sub-command.

> * `hub issue edit 123 -m "new title" -m "new body"` - Set both title and body to new values non-interactively. I think this will work with your current implementation. 👍

This will currently work

>   Should we support editing only body while keeping the title intact, and if so, how would we go about that?

I assume you mean interactively? I can't really judge whether this something needed or not. It doesn't seem like a crucial feature tho.

mislav:
> An alternative could be using a different keyword than `edit`. Maybe `hub issue update 123`? In this case, the `hub issue update 123 --edit` is less weird.

That's a great idea. Let's go with `hub issue update <number> [<flags>]`. If no flags are given, the command should fail and suggest that you either need to specify `--edit` or pass some specific fields to be updated.

> If no options are passed, we could just open the issue on github.com

I think this could be surprising, so let's not define a default just yet. We can always choose a good default in the future when it becomes clear how people are using `issue update`.

> > Should we support editing only body while keeping the title intact, and if so, how would we go about that?
> 
> I assume you mean interactively? I can't really judge whether this something needed or not.

You're right, it's pretty edge-casey so we don't have to worry about that for now.

> I think the `--add-labels` and `--remove-labels` would be easier from a users perspective. We can support those alongside `--labels`.

I also think that this would be the most user-friendly, but you can skip implementing add/remove labels for now, and just support `--labels` replacing the list of labels with new ones. We can see if people ask for granular label add/remove operations, and avoid expanding the scope of your PR for now.

> I wonder if we should support `--title` and `--body` as flags instead? And ditch supporting `-m` for this sub-command.

Yeah, `-m` potentially having title and body combined was not a great design choice for hub. (I've inherited that from `git commit`.) I'm up for having separate `--title` and `--body`, but I think we should still support `update -m "..."` for symmetry with `create`.

Thank you for your thoughtful suggestions!

jfahrer:
> That's a great idea. Let's go with `hub issue update <number> [<flags>]`. If no flags are given, the command should fail and suggest that you either need to specify `--edit` or pass some specific fields to be updated.

Excellent. I updated the PR. I couldn't find anything that returns the number of flags or a list of flags so that the `hub issue update` command can fail if no flags are provided. Did I miss it? Is there a better way to determine if a flag was provided?
If not, I would add a function to `ArgsParser` that returns the number of provided flags.

> Yeah, `-m` potentially having title and body combined was not a great design choice for hub. (I've inherited that from `git commit`.) I'm up for having separate `--title` and `--body`, but I think we should still support `update -m "..."` for symmetry with `create`.

Sounds good. Let me add `--title` and `--body` in a separate PR. I assume it would make sense to add those flags to other commands like `issue create` and `pull-request` as well?

> I also think that this would be the most user-friendly, but you can skip implementing add/remove labels for now, and just support `--labels` replacing the list of labels with new ones. We can see if people ask for granular label add/remove operations, and avoid expanding the scope of your PR for now.
> 

I'll skip this for now then.

mislav:
> If not, I would add a function to `ArgsParser` that returns the number of provided flags.

That might be one approach, but instead I would opt for the more explicit approach of checking whether specific flags were passed (in case there will be other `update` flags in the future that aren't related to any specific field):

```go
if !args.HasReceived("--edit") && !args.HasReceived("--message") && !args.HasReceived("--labels") && ... {
  utils.Check(cmd.UsageError("please specify fields to update"))
}
```

jfahrer:
> That might be one approach, but instead I would opt for the more explicit approach of checking whether specific flags were passed (in case there will be other `update` flags in the future that aren't related to any specific field):

You got it

utils/args_parser.go
```diff
@@ -183,6 +183,15 @@ func (p *ArgsParser) HasReceived(name string) bool {
 	return found && len(f.values) > 0
 }
 
+func (p *ArgsParser) HasReceivedOneOf(names []string) bool {
```
jfahrer: Might be too much to open the ArgsParser API up like this. It felt more natural to put this logic here instead of the Issue. Let me know if you want me to move it.

mislav: I think this is too specific for ArgsParser (which I intentionally kept small), so I'd rather that this is inlined in issue.go or encapsulated as a function that takes an ArgsParser instance and a list of flags.

Btw you can have a Go function accept a list of arguments instead of a slice:

```go
func HasReceivedOneOf(names ...string) bool {
}

// usage:
HasReceivedOneOf("foo", "bar")
```

jfahrer: > I think this is too specific for ArgsParser (which I intentionally kept small), so I'd rather that this is inlined in issue.go or encapsulated as a function that takes an ArgsParser instance and a list of flags.

Sounds good, I inlined the check in issue.go.
 
> Btw you can have a Go function accept a list of arguments instead of a slice:
 
Cool, I didn't think of that. Thanks for the tip!

mislav:
@jfahrer Pushed some tweaks to your branch, such as supporting updating issue state. Thank you for the amazing work here! ✨

jfahrer:
Thank you!

PR #2454, opened 2020-01-26
Use full option names in examples

silvenon:
This way it's easier to skim instead of always having to read commments
for explanations.

### DIFF
--- a/index.html
+++ b/index.html
@@ -55,7 +55,7 @@ to date</em>, and <em>share logs or code snippets via Gist</em>.
 <span class=p></span><strong>hub sync</strong>
 
 <span class="comment"># list latest open issues in the current repository</span>
-<span class=p></span><strong>hub issue</strong> -L 10
+<span class=p></span><strong>hub issue</strong> --limit 10
 
 <span class="comment"># open the current project's issues page</span>
 <span class=p></span><strong>hub browse</strong> -- issues
@@ -66,7 +66,7 @@ to date</em>, and <em>share logs or code snippets via Gist</em>.
 <span class="result">→ open https://github.com/rbenv/ruby-build/wiki</span>
 
 <span class="comment"># share log output via Gist</span>
-<span class=p></span><strong>hub gist create</strong> -c build.log
+<span class=p></span><strong>hub gist create</strong> --copy build.log
 <span class="result">→ (the URL of the new private gist copied to clipboard)</span>
 </pre>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -102,15 +102,15 @@ <h2>Lowering the barrier to contributing to open-source</h2>
 <span class=p></span>git commit -am "done with feature"
 
 <span class="comment"># It's time to fork the repo!</span>
-<span class=p></span><strong>hub fork</strong> --remote-name=origin
+<span class=p></span><strong>hub fork</strong> --remote-name origin
```
silvenon: I changed this for the sake of consistency.

index.html
```diff
@@ -147,22 +147,24 @@ <h3>Drop down to the API level</h3>
...
 <span class=p></span>REPO="github/hub"
 <span class=p></span>SHA="b0db79db"
-<span class=p></span><strong>hub api graphql</strong> -t -f q="repo:<span class=value>$REPO</span> type:pr <span class=value>$SHA</span>" -f query=<span class=value>'
-  query($q: String!) {
-    search(query: $q, type: ISSUE, first: 3) {
-      nodes {
-        ... on PullRequest {
-          url
+<span class=p></span><strong>hub api graphql</strong> --flat \
+  --raw-field query="repo:<span class=value>$REPO</span> type:pr <span class=value>$SHA</span>" \
```
mislav: Uhhh I don't like seeing `--field` and `--raw-field` in api examples; they are already verbose enough without that. I think I would like to keep `-F` and `-f` here and let the reader infer that they are fields.

index.html
```diff
@@ -127,13 +127,13 @@ <h2>Automating tasks for fun and profit</h2>
...
-<span class=p></span><strong>hub issue</strong> -a YOUR_USER -l urgent
+<span class=p></span><strong>hub issue</strong> --assignee YOUR_USER --labels urgent
 
 <span class="comment"># List the URLs of at most 20 pull requests based on the "develop" branch:</span>
-<span class=p></span><strong>hub pr list</strong> -L 20 -b develop --format='%t [%H] | %U%n'
+<span class=p></span><strong>hub pr list</strong> --limit 20 --base develop --format='%t [%H] | %U%n'
 
 <span class="comment"># Create a GitHub Release from master using release notes from a file</span>
-<span class=p></span><strong>hub release create</strong> -c -F release-notes.txt v2.3.0
+<span class=p></span><strong>hub release create</strong> --copy --file release-notes.txt v2.3.0
```
mislav: Since a filename immediately follows `-F`, I wonder whether that's enough for the reader to understand that this is the flag for passing a file and that we don't need to expand it. I'm reading through these changed examples now and thinking that we could maybe get away with shortening `--file` back into `-F`

index.html
silvenon: I agree, fixed.

index.html
silvenon: I agree, fixed.

Issue #1644, opened 2018-01-10
Fatal: Could not read Username for "https://github.com", No such device or address


simonepri:
Hello! :octocat:

Today I was doing the setup of a **git** automation inside a Docker container using **hub** and I ran into something that seems a **bug**.

### Authentication method
Since the problem is related to an authentication problem let me give you some context before to talk about the error I have encountered.
In order to perform all the operations, I was using a [personal token](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/) as mentioned in [hub's man page](https://hub.github.com/hub.1.html#GitHub-OAuth-authentication)

The part of the script that was configuring hub was looking like this:
```bash
hub config --global hub.protocol https
hub config --global user.email $email
hub config --global user.name $username
export GITHUB_TOKEN=$token
```

### The problem
My automation was trying to perform the following operations on a public repository:
- clone the repository ✅
- fork the repository ✅
- add some changes to the fork ✅
- push the changes to the fork ❌
- open a pull request on the original repository ❌

The problem was that the push and the pull-request commands both gave me this (*meaningless*) error:
![fatal: could not read Username for "https://github.com", No such device or address](https://user-images.githubusercontent.com/3505087/34791899-a763e228-f646-11e7-97a9-bbfd64282b4b.png)

After a long struggling, I've then found a solution that seems a bit *wired* to me.

In order to be able to push on my fork I was needed to change this:
```
hub remote set-url upstream https://github.com/$owner/$repo.git
hub remote set-url origin https://github.com/$username/$repo.git
```
to this:
```
hub remote set-url upstream https://$token:x-oauth-basic@github.com/$owner/$repo.git
hub remote set-url origin https://$token:x-oauth-basic@github.com/$username/$repo.git
```

Why the **$GITHUB_TOKEN** environmental variable isn't enough in this situation?
The `$GITHUB_TOKEN `env var enables me to create a new repo (forking) on my account but does not give me the possibility to push on the same? Am I missing something?

### Info to reproduce the error
Below you can find the "complete" script that lets you reproduce the error:

#### First (non-working) version of my script (*click on Details*):
<details>

```bash
# You must set these variables: $email, $username, $token, $repo, $owner 

hub config --global hub.protocol https
hub config --global user.email $email
hub config --global user.name $username
export GITHUB_TOKEN=$token

# Clones the repo
rm -rf ./$repo/
hub clone $owner/$repo
cd ./$repo/

# Fork the repo
hub fork

# Update remotes
[...]

PR #2461, opened 2020-02-06
Add information about configuring actions with hub

JLLeitschuh:
There's a lot of scattered information about how to properly configure hub to be used on GitHub actions.
This attempts to condense this to a simplified bit of information captured in the README.

See:
 - #1644
 - #225

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,8 +98,8 @@ EOF
 # Set the default basic-auth credentials for gits http access
 cat >>~/.netrc <<EOF
 machine github.com
-login $GITHUB_ACTOR
-password $GITHUB_TOKEN
+  login $GITHUB_ACTOR
+  password $GITHUB_TOKEN
 EOF
 ```
 
--- a/README.md
+++ b/README.md
@@ -70,15 +70,38 @@ steps:
     GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 ```
 
+##### Interacting with another Repository
+
 Note that the default GITHUB_TOKEN will only work for API operations within _the
 same repo that runs this workflow_. If you need to access or write to other
 repositories, [generate a Personal Access Token][pat] with `repo` scope and add
 it to your [repository secrets][].
 
-
 [github actions]: https://help.github.com/en/actions/automating-your-workflow-with-github-actions
 [pat]: https://github.com/settings/tokens
 [repository secrets]: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets
+
+Additionally, in order to use git with this external repository you will need to make additional configuration
+changes to your actions execution enviroment.
+
+```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,16 +70,39 @@ steps:
...
+mkdir ~/.config
+cat >>~/.config/hub <<EOF
+github.com:
+- user: $GITHUB_ACTOR
+EOF
+
+# Set the default basic-auth credentials for gits http access
+cat >>~/.netrc <<EOF
+machine github.com
+login $GITHUB_ACTOR
```
JLLeitschuh:
```suggestion
  login $GITHUB_ACTOR
```

README.md
```diff
@@ -70,16 +70,39 @@ steps:
...
+password $GITHUB_TOKEN
```
JLLeitschuh:
```suggestion
  password $GITHUB_TOKEN
```

README.md
```diff
@@ -70,16 +70,39 @@ steps:
...
 [github actions]: https://help.github.com/en/actions/automating-your-workflow-with-github-actions
 [pat]: https://github.com/settings/tokens
 [repository secrets]: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets
 
+Additionally, in order to use git with this external repository you will need to make additional configuration
+changes to your actions execution enviroment.
+
+```bash
+# Configure hub to use https instead of the git protocol
+git config --global hub.protocol https
```
mislav: For which hub command is this necessary? I'm guessing `hub clone`/`hub remote add`?

README.md
```diff
@@ -70,16 +70,39 @@ steps:
...
+
+# Configure the hub user
+mkdir ~/.config
+cat >>~/.config/hub <<EOF
+github.com:
+- user: $GITHUB_ACTOR
```
mislav: For which hub command is this necessary? If `GITHUB_TOKEN` is set, ideally `~/.config/hub` should not be necessary in this environment

README.md
```diff
@@ -70,16 +70,39 @@ steps:
...
+EOF
+
+# Set the default basic-auth credentials for gits http access
+cat >>~/.netrc <<EOF
```
mislav: Is this necessary even if you use `actions/checkout@v2`? I think it might properly set up push access to repositories so you don't have to do this manually

JLLeitschuh: That wasn't my experience.

README.md
JLLeitschuh: That wasn't my experience.

mislav:
Thank you, but I'm going to close this since I think our current instructions for GitHub Actions already cover most usages. GITHUB_TOKEN is enough to configure hub authentication (no need to write `~/.config/hub`), and cloning and pushing to other repositories within Action is git-specific, not hub-specific, and in my view should be left as an exercise to the user.

PR #2463, opened 2020-02-07
remove bmizerany/assert

lufia:
## motivation
github.com/bmizerany/assert is no longer maintained.

## done
I added **internal/assert** and replaced bmizerany/assert with it. By this changes the failure messages will change as below.

### assert.Equal

```console
// old
% go test
--- FAIL: TestNew (0.00s)
    assert.go:15: /Users/lufia/src/github.com/github/hub/cmd/cmd_test.go:12
    assert.go:24: ! "vim -x-noplugin" != "vim --noplugin"
FAIL
exit status 1

// new
% go test                        
--- FAIL: TestNew (0.00s)
    cmd_test.go:11: 
          string(
        - 	"vim -x-noplugin",
        + 	"vim --noplugin",
          )
    cmd_test.go:12: 
          int(
        - 	1,
        + 	0,
          )
FAIL
exit status 1
```

### assert.NotEqual
```console
% go test
--- FAIL: TestCommandUseErrorWhenMissingSubcommand (0.00s)
    assert.go:15: /Users/lufia/src/github.com/github/hub/commands/commands_test.go:53
    assert.go:62: !  Unexpected: <<nil>>
% go test
--- FAIL: TestCommandUseErrorWhenMissingSubcommand (0.00s)
    commands_test.go:52: 
        Unexpected: <<nil>>
```


lufia:
I think there are two options how to replace bmizerany/assert package: use testing of stdlib, or use other third-party package similar bmizerany/assert.

Which do you like?

cmd/cmd_test.go
```diff
@@ -2,19 +2,25 @@ package cmd
 
 import (
 	"testing"
-
-	"github.com/bmizerany/assert"
 )
 
 func TestNew(t *testing.T) {
 	execCmd := New("vim --noplugin")
-	assert.Equal(t, "vim --noplugin", execCmd.Name)
```
mislav: @lufia Thanks for suggesting this!

To avoid having to update many tests (which in general is always risky), could we find another library that offers the same interface, or roll out our own mini-package that simply implements the assertion methods we need? That way we could simply switch the import statements and not have to update any code.

Also, I don't feel the pressing need to migrate away from bmizerany/assert right now, unless the repository is getting deleted. It works well and we don't need it to gain any new features. 

Let me know what you think!

lufia: @mislav Thank you for your reply.

My motivation was I want to upgrade all dependencies.

However the command `go get -u all` don't upgrade github.com/kr/pty@v1.1.1 to v1.1.8.

- github.com/bmizerany/assert  is depended on kr/pretty
- github.com/kr/pretty is depended on kr/text@v0.1.0
- github.com/kr/text@v0.1.0 is depended on kr/pty@v1.1.1

I read that bmizerany/assert is no longer maintained. Moreover github.com/kr/pty is archived. Thus I thinked replacing bmizerany/assert is better than keeping it.

mislav: Thanks for that info!

> or roll out our own mini-package that simply implements the assertion methods we need? That way we could simply switch the import statements and not have to update any code.

What do you think about this proposal? I think that it would be my preferred solution and wouldn't require changing any tests. Let me know if you would be able to do it in this PR.

You could put the package under `internal/assert/`.

lufia: @mislav I'll try it!

lufia:
I replaced this and updated this issue.

commands/push_test.go
```diff
@@ -1,21 +1,10 @@
 package commands
 
 import (
-	"github.com/bmizerany/assert"
+	"github.com/github/hub/internal/assert"
 	"testing"
 )
 
-func testPush(t *testing.T) {
```
mislav, line 8: Any reason why this test was deleted?

lufia: This func was not starting with capital T so I thought it is a test helper. And it was not used from anywhere.

PR #2472, opened 2020-02-13
Drop the wrapper part, in the docs' title too

dwijnand:


index.html
```diff
@@ -2,7 +2,7 @@
 ---
 <!doctype html>
 <head>
-<title>hub · the command-line wrapper for git</title>
+<title>hub · a helpful extension to command-line git</title>
```
mislav: Hi, interesting suggestion! Any specific reason why you think that "wrapper" is not appropriate here? Technically, hub is a wrapper because it proxies all `hub <foo>` invocations to `git <foo>` (unless it was a custom hub command). So to say “wrapper” it's not really incorrect.

dwijnand: You thought you'd decided to move away from "hub is a wrapper" a few years ago, dropping the suggestion/recommendation to alias hub as git.

Basically I was synchronising [this change](https://github.com/github/hub/commit/fc25ac35d2680b863d9c26f5682ffb54322e9ecd#diff-eacf331f0ffc35d4b482f1d15a887d3bR225) you made back in Dec 2018 to the title too.

mislav: Gotcha! That makes sense.

mislav:
Thank you! I've removed "helpful" because I think all software believes that it's helpful, so that goes without saying.

PR #2485, opened 2020-02-28
Add support for determining default branch when calling hub pull-request form worktree

### LINKED ISSUE #2484: Hub doesn't use my default base branch when executing pull-request cmd from worktrees
**Command attempted:**

```HUB_VERBOSE=1 hub pull-request -p -m "Hub creating a PR to my default branch upstream/core_master"```

**More info:**

```
$ hub --version
git version 2.21.1
hub version 2.14.1
```

I've used ```git remote set-head upstream core_master``` and ```git remote set-head origin core_master``` to set my default branches.

I noticed this [issue](https://github.com/github/hub/issues/154) originally added support for hub creating pull-requests to the defualt base branch when one wasn't specified using the `-b` flag.

**What happened:**

When I do ```hub pull-request``` from my clone, hub correctly chooses _upstream/core_master_ as my default base branch; however, if I execute the same command in a worktree of this clone, hub defaults to using _upstream/master_.

From the worktree, I see the following:
```json
{"base":"master","head":"pHWChip:feature/colt_testing","maintainer_can_modify":true,"title":"Testing"}
```

From the original clone, I see the following:

```json
{"base":"core_master","head":"pHWChip:feature/colt-testing-2","maintainer_can_modify":true,"title":"Testing"}
```

**Question:**
Is this intended behavior? Is this a bug? Or am I requesting a new feature here?

csaska:
Fixes #2484 . 

When ```hub pull-request```  is called from a worktree, hub does not correctly find the default branch for the repo. 

~This code is a bit of a hack; it attempts to check for `worktree` in the path pointed to by ```$GIT_DIR/.git``` of the worktree.~

~This should work for the default cause; however, I am not sure if it will work in all scenarios, specifically when users decide to set ```git config extensions.worktreeConfig true```~

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -84,7 +84,9 @@ func CommondirName() (string, error) {
 	dirCmd.Stderr = nil
 	output, err := dirCmd.Output()
 	dir := firstLine(output)
-	if dir == "--git-common-dir" {
+	if err != nil {
+		return "", fmt.Errorf("Not a git repository (or any of the parent directories): .git")
+	} else if dir == "--git-common-dir" {
 		return "", fmt.Errorf("unable to determine git commondir directory")
 	}
 	return dir, err
--- a/git/git.go
+++ b/git/git.go
@@ -27,12 +27,16 @@ func Dir() (string, error) {
 	if cachedDir != "" {
 		return cachedDir, nil
 	}
-
-	dirCmd := gitCmd("rev-parse", "-q", "--git-dir")
-	dirCmd.Stderr = nil
-	output, err := dirCmd.Output()
+	gitDir, err := CommondirName()
 	if err != nil {
-		return "", fmt.Errorf("Not a git repository (or any of the parent directories): .git")
+		// Fall back for older version of git prior to v2.7.0 (added worktrees)
+		dirCmd := gitCmd("rev-parse", "--git-dir")
+		dirCmd.Stderr = nil
+		output, err := dirCmd.Output()
+		if err != nil {
+			return "", fmt.Errorf("Not a git repository (or any of the parent directories): .git")
+		}
+		gitDir = firstLine(output)
 	}
 
 	var chdir string
@@ -46,8 +50,6 @@ func Dir() (string, error) {
 			}
 		}
 	}
-
-	gitDir := firstLine(output)
 
 	if !filepath.IsAbs(gitDir) {
 		if chdir != "" {
@@ -73,6 +75,17 @@ func WorkdirName() (string, error) {
 	dir := firstLine(output)
 	if dir == "" {
 		return "", fmt.Errorf("unable to determine git working directory")
+	}
+	return dir, err
+}
+
+func CommondirName() (string, error) {
+	dirCmd := gitCmd("rev-parse", "--git-common-dir")
+	dirCmd.Stderr = nil
+	output, err := dirCmd.Output()
+	dir := firstLine(output)
+	if dir == "--git-common-dir" {
+		return "", fmt.Errorf("unable to determine git commondir directory")
 	}
 	return dir, err
 }


csaska:
On second thought, perhaps these changes should be added to Dir() function 
https://github.com/github/hub/blob/b0db79dbf3988dd25c0053abc39a77fc92c46bfc/git/git.go#L26

csaska:
Perhaps the fix is as simply as changing `--git`-dir to `--git-common-dir`

https://github.com/github/hub/blob/b0db79dbf3988dd25c0053abc39a77fc92c46bfc/git/git.go#L31

git/git.go
```diff
@@ -28,7 +28,7 @@ func Dir() (string, error) {
 		return cachedDir, nil
 	}
 
-	dirCmd := gitCmd("rev-parse", "-q", "--git-dir")
+	dirCmd := gitCmd("rev-parse", "-q", "--git-common-dir")
```
mislav: What is the earliest git version where `--git-common-dir` was introduced?

csaska: Looks like it was first introduced in https://git-scm.com/docs/git-rev-parse/2.5.1.

csaska: Based on these two pages, I'm inclined to say that it will always work, even if the path is relative.

https://git-scm.com/docs/git

`GIT_COMMON_DIR`
> If this variable is set to a path, non-worktree files that are normally in $GIT_DIR will be taken from this path instead. Worktree-specific files such as HEAD or index are taken from $GIT_DIR

`commondir`
>If this file exists, $GIT_COMMON_DIR (see git[1]) will be set to the path specified in this file if it is not explicitly set. If the specified path is relative, it is relative to $GIT_DIR. The repository with commondir is incomplete without the repository pointed by "commondir"

csaska: If you're setting `GIT_COMMON_DIR`, then based on the docs, it seems like you should be expecting this dir to be used for all non-worktree files. I guess your question though is concerned with what other code is using `Dir()`? So, what other hub features rely on `GIT_DIR`, and do they ever expect to use worktree-specific files?

mislav: Requiring git 2.5 all of the sudden is a dealbreaker, since thus far we've supported git 1.7+

csaska: Are you against something like...
```
        dirCmd := gitCmd("rev-parse", "-q", "--git-dir")
	parseCmd.Stderr = nil
	gitDir, err := parseCmd.Output()

	parseCmd := gitCmd("rev-parse", "--is-inside-worktree")
	parseCmd.Stderr = nil
	isWorktree, err := parseCmd.Output()

        if (isWorkTree) {
            // assume worktree and use GIT_DIR/../../
            gitDir = gitDir + "/../../"

            // check if path exists, otherwise fallback to regular behavior
        }
```

csaska: Okay, so git worktrees aren't support in git@v.1.7.0. Any hub command attempted from a worktree in git@1.7.0 will fail.

So, changing this code to use ```git rev-parse --git-common-dir``` instead of ```git rev-parse --git-dir``` will not change the behavior in any git release that does not include worktree support. 

image

image

csaska: Okay so now looking at the first release of git with worktrees.

image

Based on these results, I would say there is no issue in changing ```git rev-parse --git-dir``` to ```git rev-parse --git-common-dir```. If you are in a worktree prior to git@v2.6.0, you ```rev-parse``` command will fail anyway.

mislav: > Based on these results, I would say there is no issue in changing `git rev-parse --git-dir` to `git rev-parse --git-common-dir`. If you are in a worktree prior to git@v2.6.0, your `rev-parse` command will fail anyway.

That's a good point; thank you for looking into this in such depth!

I'm wondering what about when you're _not_ in a worktree, but in a regular git working directory, and you issue `git rev-parse --git-common-dir` with an old git version. Wouldn't it fail because `--git-common-dir` flag is not recognized there?

csaska: lol yes you're right...I got so hung up on worktrees that I forgot about the nomal scenario.

git/git.go
```diff
@@ -27,12 +27,16 @@ func Dir() (string, error) {
...
 	if err != nil {
-		return "", fmt.Errorf("Not a git repository (or any of the parent directories): .git")
+		// Fall back for older version of git prior to v2.7.0 (added worktrees)
+		dirCmd := gitCmd("rev-parse", "--git-dir")
+		dirCmd.Stderr = nil
+		output, err := dirCmd.Output()
+		if err != nil {
+			return "", fmt.Errorf("Not a git repository (or any of the parent directories): .git")
+		}
+		gitDir = firstLine(output)
```
csaska: @mislav Thanks for the responses!  I added this function. Thoughts being this perhaps isn't the last time a ```hub``` command has different behavior in a worktree.

https://github.com/github/hub/blob/479ed29d9253c88f4da78caf8c9061a6da760528/git/git.go#L82

In ```Dir()```, if the function returns an error, we fall back to the original behavior of the code

csaska:
>This PR just feels like we're making an already complicated approach even more complicated.

I completely agree.

PR #2495, opened 2020-03-07
Fix case upper to lower

peaceiris:
Convert case upper to lower in the get script for GitHub Actions.

```diff
- Location
+ location
```


mislav:
Good catch!

PR #2504, opened 2020-03-20
Linting Cleanup

kfcampbell:
Current master branch (HEAD `2cff4c2a`) currently has 1045 linting errors. This PR represents a first pass at cleaning some of those up.

Many of them are missing `godoc` annotations which will require a little bit more time from me to do accurately, so I've elected to leave them out of this PR for now. I've also decided to leave out error message changes as I'm not sure what (if any) automation or monitoring tools depend on the error message strings.


kfcampbell:
Yeah I have a couple more things I'd like to do if that's okay! I have a hacky way to measure the changes (`golint ./... > output.txt` and then I count the lines in the file) and I want to "publish" the before/after numbers on this PR, and I also have a few more changes to make.

I'm sorry, I'm trying to balance a few things here so that's why it's taking a while. I'll get there!

mislav:
No rush! Thank you 🙇

mrsakryckman:
Thank you

mislav:
> Many of them are missing `godoc` annotations which will require a little bit more time from me to do accurately, so I've elected to leave them out of this PR for now.

Since this project isn't a library to be imported in other Go projects, I find it acceptable that many methods do not have godoc annotations. With this in mind, what do you think about using `golangci-lint run` as a linter for this project instead of golint? I find this meta-linter being pretty solid quality out-of-the-box and, most importantly, it surpresses “exported method * should have comment or be unexported” errors from golint.

kfcampbell:
Master branch linting errors: 1045
Current HEAD of this branch linting errors: 279
Total errors removed by this PR: 766

I believe this is ready for review now!

kfcampbell:
Regarding the linting change, I'm not super in love with `golint` for reasons listed in the package:

> We will not be adding pragmas or other knobs to suppress specific warnings, so do not expect or require code to be completely “lint-free”. In short, this tool is not, and will never be, trustworthy enough for its suggestions to be enforced automatically, for example as part of a build process.

(see [here for the source](https://pkg.go.dev/golang.org/x/lint/golint?tab=overview)).

So I think that's a reasonable change!

kfcampbell:
Note that all of the remaining linting errors are missing comments for exported members or error string capitalization linting, which as shown above is outside of scope for this PR.

@mislav this merge conflict situation is a little weird...I've resolved it locally and pushed my changes, but would you mind going over the code to make sure the changes from the last commit are appropriate? Thank you!

PR #2544, opened 2020-05-17
Fix private data leak

### LINKED ISSUE #2222: hub should not leak local user account to github servers
When using hub for the firs time it will automatically create an OAuth token. This gets logged into the security history viewable on the accounts profile under the line "oauth_authorization.create".  This log line will be in the format of:

> oauth_authorization.create – Personal access token (hub for user@hostname) 

There is no reason to send the local unix account to Github servers, as local account username might be different from Github username, and as such private information (e.g. that identity could be used elsewhere by the user or might be one-off). This probably also violates computer privacy laws (e.g. GDPR) as only user's github account name/password/2fac shall be sufficient to access the service via the CLI. Finally this is not mentioned in the documentation at all, which means the user has no way of making an informed choice whether to use the CLI or not.

Tool should ask for a OAuth key "vanity name" to be submitted on first connection if the user wants to audit access logs instead of user@hostname.

quantimnot:
fix #2222
Don't read the user or hostname.
Use the value of env var HUB_MACHINE or a default value "`<unidentified machine>`".

Here is the result in my security log:
Created authorization for Personal access token (hub for `<unidentified machine>`) with gist, repo scope(s)

### DIFF
--- a/features/authentication.feature
+++ b/features/authentication.feature
@@ -7,7 +7,7 @@ Feature: OAuth authentication
       """
       require 'socket'
       require 'etc'
-      machine_id = "#{Etc.getlogin}@#{Socket.gethostname}"
+      machine_id = "<unidentified machine>"
 
       post('/authorizations') {
         assert_basic_auth 'mislav', 'kitty'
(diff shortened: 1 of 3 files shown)


chdsbd:
Hey, sorry about closing this issue. I was testing a bug with GitHub.

Issue #2560, opened 2020-06-07
Add Package for WSL Windows Subsystem for Linux.  


Donaldpherman:
I'm trying to install hub on WSL using snap install but I get the following error.

sudo snap install hub --classic
Interacting with snapd is not yet supported on Windows Subsystem for Linux.
This command has been left available for documentation purposes only.

piroor:
I also tried to install `hub` on my Ubuntu 18.04LTS on WSL1 on Windows 10 and I saw same error.

# What I did

1. Start Windows Terminal and open new tab with Ubuntu on WSL1. Or, simply start Ubuntu on WSL. Then a bash prompt appears.
2. Run `scoop install hub` as the installation command for Windows, and I saw a "command not found" error.
3. Run `choco install hub` as another installation command for Windows, and I saw a "command not found" error again.
4. Run `sudo snap install hub --classic` as the installation command for Ubuntu, and I saw a message "Interacting with snapd is not yet supported on Windows Subsystem for Linux. This command has been left available for documentation purposes only.".
5. With the instruction under the "Standalone" section, I go to the latest release: https://github.com/github/hub/releases/latest
6. Because my PC has AM64 CPU, I downloaded https://github.com/github/hub/releases/download/v2.14.2/hub-linux-amd64-2.14.2.tgz
   
   ```
   $ cd /tmp
   $ wget https://github.com/github/hub/releases/download/v2.14.2/hub-linux-amd64-2.14.2.tgz
   ```
7. Extract files from the downloaded file.
   
   ```
   $ unar hub*
   ```
8. Move `bin/hub` from extracted files to an executable directory.
   
   ```
   $ mv hub*/bin/hub ~/local/bin/
   ```
9. Run `hub`, then I saw a usage message of the command successfully.

# Expected

There is an instruction to install `hub` for Ubuntu on WSL.

# Actual

There is no special instruction for Ubuntu on WSL. There are instructions for Windows and Ubuntu but both don't work for Ubuntu on WSL.

Donaldpherman:
That worked great.  Thanks for the help.

mislav:
I have just removed generic Ubuntu installation instructions for Snap from our readme because hub snap did not work well even on Ubuntu proper.

It's a bummer for now, but WSL users will have to resort to "Standalone" installation flow until there is a proper Ubuntu package available from a repository.

anonymouse64:
[...]

PR #2562, opened 2020-06-13
Add instruction for installation on Ubuntu on WSL1

piroor:
This should fixe #2560 . How about this change?


mislav:
Thank you, but I think this makes installation instructions needlessly complex. WSL users should be able to follow the current version of "Standalone" instructions just fine.

PR #2566, opened 2020-06-20
Update README to include hub on conda forge for all platforms

ctb:
see https://github.com/conda-forge/hub-feedstock for conda-forge hub package for all platforms.


mislav:
Thank you!

PR #2585, opened 2020-07-27
Enhance `hub api` to recognise an empty array field

simon-engledew:
Hey, thanks for hub!

I made this tiny tweak to allow me to bypass `required_contexts` when doing staging deployments. The endpoint requires an empty array:

  image

But hub currently sends the JSON string `"[]"`. This change will send the empty array instead.

Changing the behaviour might not be desirable in the main repository! 😄 If nothing else, hopefully this has a bunch of useful keywords for anyone else with the same problem.

Let me know if you want me to do any additional work on it or feel free to close it if it's too niche.

Cheers,
~ Si

### DIFF
--- a/commands/api.go
+++ b/commands/api.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -40,8 +41,6 @@ var cmdAPI = &Command{
 		If <VALUE> is "true", "false", "null", or looks like a number, an
 		appropriate JSON type is used instead of a string.
 
-		If <VALUE> is "[]", an empty JSON array is used instead of a string.
-
 		It is not possible to serialize <VALUE> as a nested JSON array or hash.
 		Instead, construct the request payload externally and pass it via
 		''--input''.
@@ -55,6 +54,10 @@ var cmdAPI = &Command{
 		Same as ''--field'', except that it allows values starting with "@", literal
 		strings "true", "false", and "null", as well as strings that look like
 		numbers.
+
+	-j, --json-field <KEY>=<VALUE>
+		Same as ''--field'', except that it expects a valid JSON value that will be
+		passed directly to <ENDPOINT>.
 
 	--input <FILE>
 		The filename to read the raw request body from. Use "-" to read from standard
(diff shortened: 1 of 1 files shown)


simon-engledew:
Sure - that sounds like a more useful addition! 😄 

I was a little bit worried about overloading the `--field` argument with behaviour that could break existing (valid) usage.

Rather than add a special case that checks for a valid array, I've added a new argument: `--json-field`. 

This will validate that the JSON is correct, but otherwise pass it unaltered directly to ENDPOINT.

I originally used `json.Valid`, but you can't get the error message returned from the scanner. 😭 Unmarshalling the JSON into an empty interface gives you a helpful error message at the expense of some additional overhead:

image

simon-engledew:
Using --field

image

Using --json-field

image

mislav:
> Rather than add a special case that checks for a valid array, I've added a new argument: `--json-field`.

Thanks for doing that! But I would like us to backtrack a little bit.

The primary point of the `hub api` command was to handle JSON encoding & decoding for environments that don't have access to JSON facilities (e.g. bash scripts). With the new `--json-field` flag, it looks like people will be required to construct valid JSON by hand again, which defeats the point of `hub api`.

Instead, what do you think about the syntax proposed in the GitHub CLI repo? https://github.com/cli/cli/issues/1484

simon-engledew:
> > Rather than add a special case that checks for a valid array, I've added a new argument: `--json-field`.
> 
> Thanks for doing that! But I would like us to backtrack a little bit.
> 
> The primary point of the `hub api` command was to handle JSON encoding & decoding for environments that don't have access to JSON facilities (e.g. bash scripts). With the new `--json-field` flag, it looks like people will be required to construct valid JSON by hand again, which defeats the point of `hub api`.
> 
> Instead, what do you think about the syntax proposed in the GitHub CLI repo? [cli/cli#1484](https://github.com/cli/cli/issues/1484)

Oh, that looks cool – great idea! 👍I think that is a way better solution. 😸

It looks like the syntax for Rails array parameters, so there is prior art - and that seems to be holding up pretty well after over a decade of use. 😄

I also really like the nested object syntax, although I can see that being more divisive. 🤔 I guess the alternative is adding more subcommands for APIs that require nested inputs and keeping `api` lean? Tough call!

Thanks for taking the time to think about this issue! I'll close this PR and keep an eye on that issue instead. 👁️

PR #2589, opened 2020-08-08
Add --unassigned flag to issue command. Fixes #2487

### LINKED ISSUE #2487: Get all unclaimed issues
**The problem I'm trying to solve:**
A command that can check all unclaimed issues

cvhariharan:
This PR adds an --unassigned flag to issue command to allow fetching unassigned issues. 
### Example usage
```
hub issue --unassigned
```

### DIFF
--- a/commands/issue.go
+++ b/commands/issue.go
@@ -17,7 +17,7 @@ var (
 	cmdIssue = &Command{
 		Run: listIssues,
 		Usage: `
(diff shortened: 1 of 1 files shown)


cvhariharan:
`--no-assignee` looks like a good option. `--assignee=false` or `--assignee=null` also feels intuitive but are we sure that no one has those usernames? I will make the changes to use `--no-assignee`.

PR #2598, opened 2020-08-24
fix: Update checkout and cache to v2

ryuichi1208:
Update checkout and cache used in GitHub Actions workflow to v2

* [actions/checkout](https://github.com/actions/checkout)
* [actions/cache](https://github.com/actions/cache)


mislav:
Thanks!

PR #2630, opened 2020-10-08
Summary in README

AlexandroGonSan:


mislav:
Thank you, but I feel our README is already short enough and does not benefit a lot from adding a Table of Contents. If we've added it, we'd have to maintain it as well after README edits, which can be tedious.

PR #2638, opened 2020-10-18
Use stable release names

pckilgore:
Release using a stable name on an os-arch basis, so that scripts can easily target latest release without API tokens / requests / jq-jit-su.

This seemed more clear to discuss as a PR than as an issue.  Feel free to implement in a different way!

Thanks for hub.

### DIFF
--- a/script/package
+++ b/script/package
@@ -8,7 +8,7 @@ os="${1?}"
 arch="${2?}"
 version="${3?}"
 
-if [ $CI ]; then
+if [ ! -z "$CI" ]; then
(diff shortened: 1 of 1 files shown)


mislav:
This is a great idea. I can see how this naming scheme is useful for scripts.

Unfortunately, this would break backwards compatibility with existing scripts that download hub assets on every release. Changing the naming convention now would cause more trouble than the value that would be added. Because of that, I feel compelled to close this.

Thank you for your contribution!

PR #2656, opened 2020-12-01
github: fix unbuffered channel passed to signal.Notify

cuonglm:
The docs at os/signal.Notify warn about this signal delivery loss bug at
https://golang.org/pkg/os/signal/#Notify, which says:

    Package signal will not block sending to c: the caller must ensure
    that c has sufficient buffer space to keep up with the expected signal
    rate. For a channel used for notification of just one signal value,
    a buffer of size 1 is sufficient.

Caught by a static analysis tool from Orijtech, Inc. called "sigchanyzer"

### DIFF
--- a/github/config.go
+++ b/github/config.go
@@ -198,7 +198,7 @@ func getPassword() (string, error) {
 		return "", err
 	}
 
-	c := make(chan os.Signal)
+	c := make(chan os.Signal, 1)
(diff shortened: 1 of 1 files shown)


cuonglm:
Related Go issue: https://github.com/golang/go/issues/9399

mislav (approved):
Thank you!

PR #2697, opened 2021-01-18
Modify installation instructions for Ubuntu

antony:
Changes to use apt, not snap. Fixes https://github.com/github/hub/issues/2695

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Fedora Linux | [DNF](https://fedoraproject.org/wiki/DNF) | `sudo dnf install hub
 Arch Linux | [pacman](https://wiki.archlinux.org/index.php/pacman) | `sudo pacman -S hub`
 FreeBSD | [pkg(8)](http://man.freebsd.org/pkg/8) | `pkg install hub`
 Debian | [apt(8)](https://manpages.debian.org/buster/apt/apt.8.en.html) | `sudo apt install hub`
-Ubuntu | [Snap](https://snapcraft.io) | [We do not recommend installing the snap anymore.](https://github.com/github/hub/issues?q=is%3Aissue+snap)
+Ubuntu | [[apt(8)](http://manpages.ubuntu.com/manpages/focal/man8/apt.8.html)<sup>1</sup> | `sudo apt install hub`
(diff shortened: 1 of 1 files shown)


mislav (commented):
Thank you! Looks great ✨ 

1. Which version of hub does this install on Ubuntu?
2. If Ubuntu 18 isn't supported by this, should we add a notice?

lukaszstolarczuk:
Ubuntu has currently hub in 2.7.0 (https://packages.ubuntu.com/source/groovy/hub)

side note: did you guys thinking of bumping it (or asking maintainers to do so) soon, to some more recent version?

antony:
> Thank you! Looks great 
> 
> 1. Which version of hub does this install on Ubuntu?

2.7.0 is the latest I've found. It's not likely to be the latest version unless it is extremely actively maintained or automated. Most apt dependencies will install a slightly older version which has been tested by the Canonical team or contributors.

> 2. If Ubuntu 18 isn't supported by this, should we add a notice?

People should be able to reasonably expect installation instructions on the latest LTS and newer, after all - that's why it's called an LTS. Whether you want to support distributions that are now no longer supported (older than the 2 year long term support lifecycle) is up to you, but to keep things simple and clean, I would suggest not doing so.

mloskot:
> Which version of hub does this install on Ubuntu?
> > 2.7.0 is the latest I've found. 

This may be too ancient version that perhaps does not work well with latest Git credential helpers what may cause authentication problems, see https://github.com/github/hub/issues/2701

finefoot:
Not only WSL is affected: I just had the same issue with the `ubuntu:latest` Docker container (which is 20.04 focal) where `apt install hub` installed version 2.7.0 which didn't work due to `Unauthorized (HTTP 401) Bad credentials`. ([I posted on Stack Overflow](https://stackoverflow.com/questions/65894993/github-actions-with-hub-results-in-unauthorized-http-401-bad-credentials/65924815) at first because I didn't expect the outdated version to be the issue, but something with my GitHub Actions workflow.) The solution was to install the latest `hub` binary, manually, via `wget`. So, unfortunately, it seems like installing via `apt` can't be recommended at the moment.

PR #2719, opened 2021-03-03
Upgrade aruba to stable release ~v1.0.0

olivierlacan:
The alpha release was firing warnings under Ruby 3.0.0 because
it was doing something deprecated with Pathname.


mislav:
Thanks for letting us know! It looks like a lot of tests are now broken, though

mislav:
I managed to fix breakage after Aruba upgrade. Thank you for suggesting this!

Issue #2838, opened 2021-09-30
hub does not allow to update labels with space 


gagandeepsingh-dev:
**Command attempted:**

hub issue update 2 -l bug,documentation,enhancement,good goissue
**What happened:**

The label which got added was good and goissue was missed.
hub cli is not allowing to set labels with space for example labels such as good first issue cannot be set using the hub cli
**More info:**

git version 2.30.1 (Apple Git-130)
hub version refs/heads/master

gagandeepsingh-dev:
https://github.com/github/hub/pull/2839

gagandeepsingh-dev:
Workaround : hub issue update 2 -l "bug,documentation,enhancement,good goissue"

If you encode the above in double quotes it works

gagandeepsingh-dev:
@defunkt @evanphx @mojodna @schacon

evanphx:
@gagandeepsingh-dev it's incredibly rude to tag random people.

gagandeepsingh-dev:
> @gagandeepsingh-dev it's incredibly rude to tag random people.

I apologize for the distress I may have caused you.

PR #2839, opened 2021-09-30
Allow multiple labels to be passed in the update comand

gagandeepsingh-dev:
https://github.com/github/hub/issues/2838


mislav:
Sorry, but your workaround as discovered here is the proper fix for the issue: https://github.com/github/hub/issues/2838#issuecomment-934290814

We don't want to change the flag parsing code to accommodate spaces. All values with spaces need to be enclosed in quotes (or escaped by other means), always.

PR #2900, opened 2022-01-13
Update README.md

mslinn:
People might not notice that the Ubuntu instructions are the same as the Debian instructions.


mislav (commented):
Thank you, but the word "Ubuntu" is clearly visible in the line above.

mslinn:
@mi

DJSlydeezy:
I see it

PR #3001, opened 2022-08-25
silly little chnage

ashan-tharaka:


mislav (commented):
Thanks for the suggestion! But, this introduces a grammar error and I think the description reads better without it.

Issue #2992, opened 2022-08-01
unknown option in fish completion


jiegec:
**Command attempted:**

`hub status`

**What happened:**

It prints the following message:

```shell
~> git scomplete: --file: unknown option

~/.nix-profile/share/fish/vendor_completions.d/hub.fish (line 52):
complete -f -c hub -n ' __fish_hub_using_command pull-request' -s F --file -d "Read the pull request title and description from <FILE>"
^
from sourcing file ~/.nix-profile/share/fish/vendor_completions.d/hub.fish

(Type 'help complete' for related documentation)
t                            stc^C
```

**More info:**

Hub version:

```
git version 2.37.1
hub version 2.14.2
```

Fish version: 3.5.1

Relevant code: https://github.com/github/hub/blob/c8e68d548a39ec0fab6f674a669c21b54d4eec61/etc/hub.fish_completion#L52

Prokjetko001:
Vhhb

Prokjetko001:
Hsjwjs

Sithell:
Can confirm, reproduced it on my machine.
Quick fix would be to remove `--file` option from
`complete -f -c hub -n ' __fish_hub_using_command pull-request' -s F --file -d "Read the pull request title and description from <FILE>"` in `hub.fish`

jiegec:
Project is dead, sad.

Issue #2470, opened 2020-02-12
Long-term status of the hub project?


dsifford:
I apologize if this is not the correct vector to inquire about this, but I don't think stack overflow would suffice here.

With the recent announcement of the new [GitHub CLI](https://cli.github.com/), what's the roadmap look like for this project?

It seems as of right now, hub is more feature-packed, but it looks as though GitHub is throwing all its weight behind the new cli.

Would love to hear your thoughts @mislav

rgieseke:
https://mislav.net/2020/01/github-cli/

mislav:
Hi, I'm back from vacation. 👋 

Yes, hub is more feature-packed, but those features plus hub's compatibility with git (acting as a git wrapper) are hard to maintain. That's why when we thought about making GitHub CLI, we started from scratch.

I can't really tell you what the long-term status of the hub project is going to be 🙂Hub never had a long-term roadmap, it just happend because I worked on it in my free time and people contributed to it.

I'm going to continue working on it in my free time (although, not adding any substantially new features) and if you all continue to contribute as before, let's see what happens. 😉 

Thank you for asking!

derrgill:
Since GitHub CLI is now the official tool for GitHub _specifically_, might hub become more platform-agnostic / vendor-agnostic to also work with, e.g., GitLab.com and/or Bitbucket.org, and/or might the project merge with any existing similar project of that broadened scope?

mislav:
> might hub become more platform-agnostic / vendor-agnostic to also work with, e.g., GitLab.com and/or Bitbucket.org

@derrgil If I was making hub all over again, I would definitely make it with such adapters in mind! I think that the wider git community deservers platform-agnostic tools. 🚀

However, in practice, I don't think this is going to happen with hub. I'm barely finding time to make relatively small additions and fixes to hub, so it's very unlikely that I will ever attempt a large-scale refactor to support platforms that I personally have no experience with.

Furthermore, a truly vendor-agnostic tool should not be owned, subsidized by, or influenced by one of those vendors. Since I work for GitHub, you cannot and should not trust me to create a truly vendor-agnostic version control tool. Even when I am acting in good faith, I have an inherent bias towards GitHub.

A tool that you are describing would be one that is made by someone unaffiliated by either GitHub, BitBucket, or GitLab. I'm sure something like that already exists out there!

VineGaming:
> I apologize if this is not the correct vector to inquire about this, but I don't think stack overflow would suffice here.
> 
[...]

PR #3036, opened 2023-01-26
Fix `--file` fish completion

9999years:
I assume the fish completions were created by copy/pasting `--help` output, and `--file` was mistakenly left in instead of being translated to `-l file` (that is, `--file` was passed to `complete` instead of indicating that there is an option _named_ `--file`.) This leads to an error when using the completions:

```
complete: --file: unknown option

/run/current-system/sw/share/fish/vendor_completions.d/hub.fish (line 52): complete -f -c hub -n ' __fish_hub_using_command pull-request' -s F --file -d "Read the pull request title and description from <FILE>" ^
from sourcing file /run/current-system/sw/share/fish/vendor_completions.d/hub.fish

(Type 'help complete' for related documentation)
```

This patch fixes this error.


andys8:
See https://github.com/github/hub/issues/2992

ollema:
thanks!

I wonder if this will be released anytime soon - considering that the latest release was in 2020

9999years:
Yes, this will likely not be merged/released (see https://github.com/github/hub/issues/2470), but having the PR here means packagers can cherry-pick the patch for distribution, as I've done here for nixpkgs: https://github.com/NixOS/nixpkgs/pull/217636

PR #2948, opened 2022-03-28
Fix bash completion with git 2.30.0 and higher

cblecker:
This implements the fix proposed by @ches here: https://github.com/github/hub/issues/2684#issuecomment-754664270

I've been running this in my local env for over a year without issue, but when installing `hub` on a new machine, it messed up my bash completions for git, so I thought I'd actually open this PR.

fixes #2684

PR #3342, opened 2023-10-24
bash completion: add support for git 2.30 and on

### LINKED ISSUE #2684: bash completion incompatibility with git 2.30.0
**Command attempted:**
Bash completion doesn't trigger for `git` and `hub` commands (e.g. `git p<tab><tab>` doesn't suggest git subcommands but rather attempts a more general bash file/path completion)

**What happened:**
With `git` got upgraded yesterday from `2.29.0` to `2.30.0` via `brew` bash completion within new shell invocation doesn't work anymore.

**More info:**
`hub` version `2.14.2`, OS: `macOS Catalina 10.15.6`
Deinstalling `hub` fixes the issue. There were no issues with `git` of `2.29.0` version.
I'm trying to figure out whether it's only me or someone else is experiencing such issue too as turning off bash completion scripts sourcing and sourcing manually `/usr/local/etc/bash_completion.d/git-completion.bash` and then `/usr/local/etc/bash_completion.d/hub.bash_completion.sh` makes bash completion work as expected 😕

eli-schwartz:
The _git backwards compat wrapper was dropped upstream: https://github.com/git/git/commit/441ecdab37fefdacf32575a60aa523b2367c46f7

Instead, we rely on __git_complete to detect loaded bash completions, since we will use it later on to set up completions -- it is now public API.

There is a gap between git 2.30 and 2.31 where _git does not exist, but there is no public API to create completions. Starting 2.31, we are formally permitted to copy/imitate the upstream completions with:

```
__git_complete mycmd git_cmd
```

For 2.30 specifically, we have to pass the internal completion function used instead of "git_cmd", but it's difficult to detect this necessity in a forwards-compatible way. Try it that way first, in the assumption that if the internal completion function still exists it does the same thing.

Closes #2684 
Closes #2948

### DIFF
--- a/etc/hub.bash_completion.sh
+++ b/etc/hub.bash_completion.sh
@@ -2,12 +2,12 @@
 # This script complements the completion script that ships with git.
 
 # If there is no git tab completion, but we have the _completion loader try to load it
-if ! declare -F _git > /dev/null && declare -F _completion_loader > /dev/null; then
+if ! declare -F __git_complete > /dev/null && declare -F _completion_loader > /dev/null; then
   _completion_loader git
 fi
 
 # Check that git tab completion is available and we haven't already set up completion
-if declare -F _git > /dev/null && ! declare -F __git_list_all_commands_without_hub > /dev/null; then
+if declare -F __git_complete > /dev/null && ! declare -F __git_list_all_commands_without_hub > /dev/null; then
   # Duplicate and rename the 'list_all_commands' function
   eval "$(declare -f __git_list_all_commands | \
         sed 's/__git_list_all_commands/__git_list_all_commands_without_hub/')"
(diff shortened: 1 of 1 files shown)


etc/hub.bash_completion.sh
```diff
@@ -382,6 +382,15 @@ EOF
   }
 
   # Enable completion for hub even when not using the alias
-  complete -o bashdefault -o default -o nospace -F _git hub 2>/dev/null \
-    || complete -o default -o nospace -F _git hub
+  if declare -F _git >/dev/null; then
+    # git < 2.30
+    complete -o bashdefault -o default -o nospace -F _git hub 2>/dev/null \
+      || complete -o default -o nospace -F _git hub
+  elif declare -F __git_main; then
```
ches:
```suggestion
  elif declare -F __git_main >/dev/null; then
```

Consistent with the preceding conditional branch, this fixes `__git_main` being echoed during shell session initialization.

ches (commented):
Three years since https://github.com/mislav/hub/issues/2684#issuecomment-754664270 I finally wanted to look into what the proper backward-compatible and API-stable patch should be, and @eli-schwartz has already done the homework.

Thanks and 👍🏼, verified this patch works for me.

etc/hub.bash_completion.sh
eli-schwartz: Oops, yes.

eli-schwartz: In fact, in my local git clone it turns out I had already made this change!! and used it locally. The problem is, that I forgot to `git add` it...

